PR #4, opened 2019-05-09
Rebase vm-setup to latest dev-scripts version

hardys:
This aligns with the cleanups in dev-scripts, so that
we can soon switch to consuming a single version of the ansible
from this repo.

Note this rebases on the dev-scripts version in https://github.com/openshift-metal3/dev-scripts/pull/520 so this should be WIP until that merges, then we can update dev-scripts to pull from this location.  There's still some duplication in the scripts, but then mostly we'll be able to use a single implementation in both places.


russellb:
merged the other PR, will give this a try too

russellb:
I rebased this on master, ran it, and successfully provisioned a CentOS host using the `./create_machine.sh` helper.  I'm happy to merge this if you're ready to drop the WIP.

hardys:
Thanks @russellb, I rebased and also tested locally and this should be good to go now I think so removed the WIP

PR #9, opened 2019-05-15
vm-setup: remove additional DNS records

hardys:
These are needed in dev-scripts for OpenShift, but aren't needed
here so add an optional variable instead which we'll set only
when calling from openshift-metal3/dev-scripts


hardys:
This will need to be merged together with https://github.com/openshift-metal3/dev-scripts/pull/536

PR #10, opened 2019-05-15
explicitly set extradisks to boolean for flavors

dustinblack:
It seems that when picking up the `extradisks` var for `flavors.master.extradisks` and `flavors.worker.extradisks` it is resulting in Ansible treating the `false` as a string instead of a boolean. This  results in the `when` check for the _Create additional blockdevice for objectstorage nodes_ in `vm-setup/roles/libvirt/tasks/vm_setup_tasks.yml` passing when it should fail, causing the extradisks to be created and attached to the VMs when they shouldn't be.


hardys:
lgtm, thanks!

PR #16, opened 2019-06-03
Work around python*-zmq conflict

hardys:
As reported in https://github.com/openshift-metal3/dev-scripts/issues/581

We should probably look at moving this repo setup into the vm-setup roles
but for now a quick-fix is to exclude the conflicting packages.

### DIFF
--- a/01_install_requirements.sh
+++ b/01_install_requirements.sh
@@ -25,6 +25,9 @@ fi
 if ! grep -q zeromq /etc/yum.repos.d/epel.repo; then
   sudo sed -i '/enabled=1/a exclude=zeromq*' /etc/yum.repos.d/epel.repo
 fi
+if ! grep -q zmq /etc/yum.repos.d/epel.repo; then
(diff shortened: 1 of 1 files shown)


01_install_requirements.sh
```diff
@@ -25,6 +25,9 @@ fi
 if ! grep -q zeromq /etc/yum.repos.d/epel.repo; then
   sudo sed -i '/enabled=1/a exclude=zeromq*' /etc/yum.repos.d/epel.repo
 fi
+if ! grep -q zmq /etc/yum.repos.d/epel.repo; then
+  sudo sed -i '/^exclude=zeromq/ s/$/,*zmq/' /etc/yum.repos.d/epel.repo
+fi
```
russellb: This works around the problem for me, but I had to run it manually on my system.  `sudo yum -y update` from line 13 failed, so this will need to get run before that on an existing system.

hardys:
Dropping in favor of https://github.com/metal3-io/metal3-dev-env/pull/18

guoruijing:
It works for me

PR #19, opened 2019-06-07
Add RHEL8 compatibility to vm-setup.

russellb:
I've been trying to use the vm-setup playbooks on RHEL8.  This PR includes the changes I made to get it to work in my environment.

I still don't have the end-to-end workflow working, so there may be more changes later, but this is a start.

### DIFF
--- a/vm-setup/roles/libvirt/defaults/main.yml
+++ b/vm-setup/roles/libvirt/defaults/main.yml
@@ -23,6 +23,7 @@ extradisks_list:
 extradisks_size: 8G
 
 # The packages required to set up our desired libvirt environment.
+# (Tested on Centos 7)
 libvirt_packages:
   - qemu-kvm
   - libvirt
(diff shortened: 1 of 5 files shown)


vm-setup/roles/libvirt/defaults/main.yml
```diff
@@ -23,13 +23,27 @@ extradisks_list:
...
+# We expect virtualbmc to already be installed on rhel8 as a pre-req to running this,
+# as there's no rhel package available yet.
+libvirt_packages_rhel8:
+  - qemu-kvm
+  - libvirt
+  - python3-libvirt
+  - libguestfs-tools
+  - python3-lxml
+  - polkit-pkla-compat
+  - python3-netaddr
```
hardys: Where do we install virtualbmc in the el8 case?

russellb: not done here, it's assuming it got installed from source before any of this runs.  In dev-scripts, it installs it in the first prereqs script.  We could do the same in the 01 script for metal3-dev env too, but I forgot to do it.

hardys: Ack Ok not blocking on it, perhaps we should add it here though so this environment will also work on el8 - we can make that a follow-up though if you prefer.

hardys: probably we should move all the install stuff from the scripts into ansible, but that's definitely a follow-up, I'll try to take a look at it.

PR #20, opened 2019-06-25
Ensure redhat-lsb-core is installed

markmc:
Without lsb_release installed, I was getting:

  error while evaluating conditional (ansible_os_family == \"RedHat\"
  and ansible_lsb.major_release|int == 7): 'dict object' has no
  attribute 'major_release'

Introduced by RHEL8 support in commit b8508737


russellb:
thanks!

PR #23, opened 2019-06-28
Use latest kustomize & avoid unneeded '&&'s

e-minguez:
As `set -e` is configured, if the first chained command fails it will fail as well.
Also replaced 'install' to 'sudo mv' as install doesn't move the file.


russellb:
Lgtm, ideally I would have split the two changes into two separate commits. Just feedback for the future.

PR #25, opened 2019-06-28
Avoid using --vm-driver flag every time

e-minguez:
Also, having a function will make room for more customization options in the future (if any)


russellb:
thanks!

PR #29, opened 2019-07-10
Update Readme

jan-est:
Operator-sdk installation instructions added


russellb:
Thanks for the patch!

Would you mind updating the commit message to describe a bit more about the change?  It could be something like:

`README.md: Note that operator-sdk is required by baremetal-operator."

jan-est:
Sure!

jan-est:
Hi @russellb I update the commit message.

russellb:
Thanks again

Issue #21, opened 2019-06-27
Development environment on Ubuntu 


namnx228:
I'm thinking about a dev-env which can run on Ubuntu, so we will have more OS choices. Besides, this dev-env should be in Ansible playbooks only, not playbooks call bash-script and bash-scripts call playbook, so it will be easier to maintain and add more feature to it.
__What do you think about this ?__

hardys:
Sounds reasonable to me, AFAIK this should work fairly easily on any environment that supports libvirt, and I agree refactoring some of the bash scripts into the ansible roles would be helpful (although perhaps that should be tracked as a separate task) - I made a similar comment on https://github.com/metal3-io/metal3-dev-env/pull/19 and plan to spend some time on that unless anyone does it before I get to it :)

namnx228:
> Sounds reasonable to me, AFAIK this should work fairly easily on any environment that supports libvirt, and I agree refactoring some of the bash scripts into the ansible roles would be helpful (although perhaps that should be tracked as a separate task) - I made a similar comment on #19 and plan to spend some time on that unless anyone does it before I get to it :)

You're right, the refactoring task may be separated from the task that makes the dev-env running on Ubuntu. If there is nobody taking care of this task (Ubuntu task) right now, I will do it.

russellb:
Ubuntu support sounds great.

I'm more neutral on the ansible part, but I'm OK with it.

namnx228:
Hi,
I have made some changes and the dev-env now can run on my Ubuntu local machine. But there are still some problems with `bridges`. If we want to configure two `bridges` and make them permanently, the way to do it is different between Ubuntu 18.04 and lower version. Ubuntu 18.04 uses `Netplan` to do the all networking stuffs, including adding `bridges`, but lower versions of Ubuntu use `/etc/network/interfaces` to do that. However, it is the same if we just add `bridges` temporarily using `brctl`. 
However, the temporary bridges will disappear after we restart our machines. 

So, what do you think ? Should we add bridges temporarily on Ubuntu, or permanently ?

hardys:
I think using brctl is fine - particularly if there are some manual/script steps which can be re-run to revive an environment after an unexpected host reboot?

namnx228:
> I think using brctl is fine - particularly if there are some manual/script steps which can be re-run to revive an environment after an unexpected host reboot?

[...]

PR #34, opened 2019-07-29
Make dev-env support Ubuntu. 

namnx228:
# The dev-env  can run on both Ubuntu and CentOS. It can detect automatically the OS it is running on.

# List of commits made
Disable Apparmor driver of Libvirt so VMs can start
Update Makefile with OS detectable feature.
Use Ansible 2.8.2 in place of 2.5.2
Install Go 1.12 using apt 
Update previous commit
Install Pyenv and use Pip to install python packages
Fix error in pyenv installation
Fix error: current user doesn't have permission to access libvirt
change permission each time libvirtd is restarted
Add user to group 'libvirt' without relogin
Extract OS distribution tighter
Make dnsmasq configuration only on  Centos
Solve the Libvirt permission problem
Changes made to use of ip commands in ubuntu_bridge_network_configuration.sh and network_teardown_tasks.yml
Update ubuntu_bridge_network_configuration.sh
Added to ip commands

Co-authored-by: Maël Kimmerlin <mael.kimmerlin@est.tech>
Co-authored-by: Kanwar Saad <saadbravo@gmail.com>
Co-authored-by: Jan Tilles <jan.tilles@est.tech>
Co-authored-by: Kashif Khan <kashif.khan@est.tech>


russellb:
Thanks for the patch!  I'm certainly fine with merging Ubuntu support.  I will test this again in my CentOS environment before merging.

The only thing I noticed is that the ansible playbooks also work on RHEL, so some of the new checks you added for CentOS will need to check for RHEL as well.  Instead of checking the distribution, you can check `ansible_os_family == "RedHat"` and that will work for both CentOS and RHEL.

Don't worry about the bash scripts and RHEL.  They would need a few more changes unrelated to your changes.

dtantsur:
I understand it's not exactly fair to ask you that, but have you considered rewriting this all in ansible instead of creating more bash scripts?

namnx228:
Hi Russellb,
Thanks for your quick check.
I have pushed another commit to update the checks as your suggestion. 
Please don't hesitate to tell me if you need any kinds of support. I'm happy to maintain the feature that I have done and this project as well.

vm-setup/roles/virtbmc/tasks/teardown_tasks.yml
```diff
@@ -10,9 +10,17 @@
     - "/root/.vbmc/"
   become: true
 
-- name: Stop/disable the Virtual BMCs (virtualbmc >= 1.4.0+)
+- name: Stop/disable the Virtual BMCs (virtualbmc >= 1.4.0+) on CentOS
+  when:
+    - ansible_distribution == 'CentOS' or ansible_distribution == 'Fedora'
```
maelk: This also needs to be changed to support RHEL and not Fedora

vm-setup/roles/virtbmc/tasks/setup_tasks.yml
```diff
@@ -52,13 +52,22 @@
     vbmc_libvirt_uri: "qemu+ssh://{{ non_root_user }}@{{ vbmc_address }}/session?socket=/run/user/{{ non_root_user_uid }}/libvirt/libvirt-sock&keyfile=/root/.ssh/id_rsa_virt_power&no_verify=1&no_tty=1"
   when: vbmc_libvirt_uri is not defined
 
-- name: Start the Virtual BMCs (virtualbmc >= 1.4.0+)
+- name: Start the Virtual BMCs (virtualbmc >= 1.4.0+) on CentOS
```
maelk: maybe change CentOS ?

vm-setup/roles/libvirt/tasks/network_setup_tasks.yml
```diff
@@ -100,11 +100,23 @@
...
+  become: true
+
+- name: Whitelist bridges for unprivileged access on Ubuntu or Fedora
+  lineinfile:
+    dest: '/etc/qemu/bridge.conf' 
+    line: "allow {{ item.bridge }}"
+    create: yes
+  with_items: "{{ networks }}"
+  when: 
+    - (ansible_facts['distribution'] == "Ubuntu") or (ansible_facts['distribution'] == "Fedora")
```
maelk: Fedora is not in the list of supported OS in the checks in lib/common.sh

namnx228:
> I understand it's not exactly fair to ask you that, but have you considered rewriting this all in ansible instead of creating more bash scripts?

As we have discussed in issue #21, Ubuntu support and rewriting this all in Ansible should be considered as two separated tasks.  We don't want to put too much change into one pull request, so this pull request is only for Ubuntu support.

At the moment, no one is working at rewriting in Ansible (as far as I know), so anyone can take that task.

russellb:
This still works fine for me on CentOS 7.  Thanks!

PR #36, opened 2019-07-31
Add an option to use Docker instead of Podman

maelk:
Creates the environment variable CONTAINER_RUNTIME . The variable can be set to "docker" or "podman" depending on the runtime you want to use. It defaults to Podman. The requested runtime will be installed automatically.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -61,7 +61,7 @@ else
       fi
       sudo ifdown provisioning || true
       sudo ifup provisioning
-  
+
       # Need to pass the provision interface for bare metal
       if [ "$PRO_IF" ]; then
           echo -e "DEVICE=$PRO_IF\nTYPE=Ethernet\nONBOOT=yes\nNM_CONTROLLED=no\nBRIDGE=provisioning" | sudo dd of=/etc/sysconfig/network-scripts/ifcfg-$PRO_IF
@@ -69,7 +69,7 @@ else
           sudo ifup $PRO_IF
       fi
   fi
-  
+
   if [ "$MANAGE_INT_BRIDGE" == "y" ]; then
       # Create the baremetal bridge
       if [ ! -e /etc/sysconfig/network-scripts/ifcfg-baremetal ] ; then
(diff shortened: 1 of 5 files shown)


host_cleanup.sh
```diff
@@ -6,13 +6,13 @@ source lib/common.sh
...
-    sudo podman ps --all | grep -w "$name$" && sudo podman rm $name -f
+    sudo ${CONTAINER_RUNTIME} ps | grep -w "$name$" && sudo ${CONTAINER_RUNTIME} kill $name
+    sudo ${CONTAINER_RUNTIME} ps --all | grep -w "$name$" && sudo ${CONTAINER_RUNTIME} rm $name -f
 done
 
 # Remove existing pod
-if  sudo podman pod exists ironic-pod ; then
-    sudo podman pod rm ironic-pod -f
+if  sudo ${CONTAINER_RUNTIME} pod exists ironic-pod ; then
+    sudo ${CONTAINER_RUNTIME} pod rm ironic-pod -f
```
russellb: Should this be run with podman only like pod creation?

02_configure_host.sh
```diff
@@ -61,23 +61,23 @@ else
       fi
       sudo ifdown provisioning || true
       sudo ifup provisioning
-  
+
```
russellb: Nit: ideally unrelated formatting changes would be split out into another commit. You can leave it though

host_cleanup.sh
evrardjp: Agreed with @russellb .

maelk: fixed, thank you

02_configure_host.sh
maelk: ok, sorry

russellb:
tested, still works fine in my env

sb1975:
So where do I specify CONTAINER_RUNTIME ?
By default will it install docker now ?
Can we update the main document to clarify this please ?

russellb:
> So where do I specify CONTAINER_RUNTIME ?

config_$USER.sh

> By default will it install docker now ?

No, still podman by default

> Can we update the main document to clarify this please ?

Patches welcome the README :-)

PR #37, opened 2019-08-01
Move all download code into 01_install_requirement.sh

namnx228:
Hi, 
In this pull request, I move all download code to 01_installation_requirement.sh in order to make CI easier. We can use this script as a part of cloud-init , so we will have all needed binary after booting VM from a test image without running the whole dev-env. 

List commits made:
Move all download code into 01-install-requirement.sh

Add a OS checks in host_cleanup.sh 

Rebase onto upstream/metal3-dev-env and remove the all_os_install-requirements.sh

### DIFF
--- a/vm-setup/roles/packages_installation/tasks/main.yml
+++ b/vm-setup/roles/packages_installation/tasks/main.yml
@@ -1,0 +1,36 @@
+- name: Install packages on Ubuntu
+  block:
+    - name: Install packages using standard package manager
+      package:
+        name: "{{ packages.ubuntu.packages +  packages.ubuntu.libvirt }}"
+        state: present
+      become: true
+    - name: Install packages using pip
+      pip:
+        name: "{{ packages.ubuntu.pip }}"
+        state: present
+      become: true
+  when: ansible_facts['distribution'] == "Ubuntu"
+
+
+- name: Install packages on CentOS
+  block:
+    - name: Install packages using standard package manager
+      block: 
+        - name: Install packages on rhel7
+          when: ansible_lsb.major_release|int == 7
+          package:
+             name: "{{ packages.centos.rhel7.packages }}"
+             state: present
+        - name: Install packages on rhel8
+          when: ansible_lsb.major_release|int == 8
+          package:
+             name: "{{ packages.centos.rhel8 }}"
+             state: present
+      become: true
+    - name: Install packages using pip
+      pip:
+        name: "{{ packages.centos.pip }}"
+        state: present
+      become: true
+  when: ansible_os_family == "RedHat"
--- a/vm-setup/roles/packages_installation/tasks/main.yml
+++ b/vm-setup/roles/packages_installation/tasks/main.yml
@@ -2,7 +2,7 @@
   block:
     - name: Install packages using standard package manager
       package:
-        name: "{{ packages.ubuntu.packages +  packages.ubuntu.libvirt }}"
+        name: "{{ packages.ubuntu.packages}}"
         state: present
       become: true
     - name: Install packages using pip
@@ -25,7 +25,7 @@
         - name: Install packages on rhel8
           when: ansible_lsb.major_release|int == 8
           package:
-             name: "{{ packages.centos.rhel8 }}"
+             name: "{{ packages.centos.rhel8.packages }}"
              state: present
       become: true
     - name: Install packages using pip
--- a/vm-setup/vars/packages_installation.yml
+++ b/vm-setup/vars/packages_installation.yml
@@ -1,0 +1,111 @@
+packages_installation:
+  ubuntu:
+    packages:
+     - crudini 
+     - curl 
+     - dnsmasq 
+     - figlet 
+     - golang 
+     - nmap 
+     - patch 
+     - psmisc 
+     - python-pip 
+     - wget
+     - libvirt-bin 
+     - libvirt-clients 
+     - libvirt-dev 
+     - jq 
+     - nodejs 
+     - golang-go 
+     - unzip 
+     - yarn 
+     - genisoimage 
+     - qemu-kvm
+     - libguestfs-tools
+     - gir1.2-polkit-1.0
+     - libpolkit-agent-1-0
+     - libpolkit-backend-1-0
+     - libpolkit-gobject-1-0
+    pip:
+     - ansible==2.8.2
+     - lolcat 
+     - yq 
+     - virtualbmc 
+     - python-ironicclient 
+     - python-ironic-inspector-client 
+     - lxml 
+     - netaddr 
+     - requests 
+     - setuptools 
+     - libvirt-python 
+  centos:
+    rhel7:
+        packages:
+         - crudini 
+         - curl 
+         - dnsmasq 
+         - figlet 
+         - golang 
+         - NetworkManager 
+         - nmap 
+         - patch 
+         - psmisc 
+         - python-pip 
+         - python-requests 
+         - python-setuptools 
+         - vim-enhanced 
+         - wget
+         - ansible 
+         - bind-utils 
+         - jq 
+         - libguestfs-tools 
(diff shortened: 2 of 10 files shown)


russellb:
IIRC, your goal is that you want one script to run that will download everything that would be downloaded as part of metal3-dev-env?

My main conern with this particular patch is moving things out of ansible and into bash.  If anything, I'd rather go the other direction.

namnx228:
I have updated my pull request. I make another playbook to handle the package installation part. Please take a look at it. Thanks

russellb:
Thanks for the update.  I'm doing some testing on this.  I also need to make sure I have this working with one other dev env repo that makes use of the VM setup ansible playbook (https://github.com/openshift/dev-scripts.git).  If it's anything related to the other repo, I'll post fixes here.

vm-setup/vars/packages_installation.yml
```diff
@@ -0,0 +1,95 @@
...
+     - yq 
+     - virtualbmc 
+     - python-ironicclient 
+     - python-ironic-inspector-client 
+     - lxml 
+     - netaddr 
+     - requests 
+     - setuptools 
+     - libvirt-python 
+  centos:
```
russellb: Some of the common packages listed under `centos` don't exist for RHEL8.  The ones I hit were:

```
        "No package crudini available.",
        "No package figlet available.",
        "No package python-pip available.",
        "No package python-requests available.",
        "No package python-setuptools available.",
        "No package python-ironicclient available.",
        "No package python-ironic-inspector-client available.",
        "No package python-openstackclient available."
```

None of those are hard requirements, though.

How about instead of `centos` and then separate lists under `centos.libvirt` for rhel7 vs rhel8, let's just have 2 top level lists, `rhel7` and `rhel8`.  We can drop these from the rhel8 list.

russellb: @namnx228 would you like to work on this?  or would you like me to do this part?

namnx228: Hi @russellb, sorry for me late reply,
Let me do this part. 
However, Are you sure that if we drop these from rhel8 list, the dev-env still works ?

russellb: yeah, don't worry about it.  i just need the playbook to run successfully on rhel 8.  if these are needed on rhel 8, they'll have to be installed some other way, and i'll deal with that separately

namnx228: Ok, I will remove them in the next commit

nordixinfra:

russellb:
Thanks for the updates.  I tested this successfully on centos 7.  Now to check RHEL 8...

russellb:
on RHEL8:

```
    "changed": false,                                                                                                                                                                                                                         
    "invocation": {                                                                                                                                                                                                                           
        "module_args": {                                                                                                                                                                                                                      
            "allow_downgrade": false,                                                                                                                                                                                                         
            "autoremove": false,                                                                                                                                                                                                              
            "bugfix": false,                                                                                                                                                                                                                  
            "disable_gpg_check": false,                                                                                                                                                                                                       
            "disable_plugin": [],                                                                                                                                                                                                             
            "disablerepo": [],                                                                                                                                                                                                                
            "download_only": false,                                                                                                                                                                                                           
            "enable_plugin": [],                                                                                                                                                                                                              
            "enablerepo": [],                                                                                                                                                                                                                 
            "exclude": [],                                                                                                                                                                                                                    
            "install_repoquery": true,                                                                                                                                                                                                        
            "install_weak_deps": true,
            "installroot": "/",
            "lock_timeout": 30,
            "name": {
                "package": [
                    "curl",
                    "dnsmasq",
                    "golang",
                    "NetworkManager",
                    "nmap",
                    "patch",
                    "psmisc",
                    "vim-enhanced",
                    "wget",
                    "ansible",
                    "bind-utils",
                    "jq",
                    "libguestfs-tools",
                    "nodejs",
                    "redhat-lsb-core",
                    "unzip",
                    "genisoimage",
                    "qemu-kvm",
                    "libvirt",
                    "libvirt-devel",
                    "libvirt-daemon-kvm",
                    "python3-libvirt",
                    "libguestfs-tools",
                    "python3-lxml",
                    "polkit-pkla-compat",
                    "python3-netaddr",
                    "virt-install"
                ]
            },
            "security": false,
            "skip_broken": false,
            "state": "present",
            "update_cache": false,
            "update_only": "no",
            "validate_certs": true
        }
    },
    "msg": "argument name is of type <class 'dict'> and we were unable to convert to list: <class 'dict'> cannot be converted to a list"
```

maelk:
This also fails for Ubuntu 
`The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'libvirt'\n\nThe error appears to be in '/home/****/metal3/vm-setup/roles/packages_installation/tasks/main.yml': line 3, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  block:\n    - name: Install packages using standard package manager\n      ^ here\n`

vm-setup/roles/packages_installation/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
+- name: Install packages on Ubuntu
+  block:
+    - name: Install packages using standard package manager
+      package:
+        name: "{{ packages.ubuntu.packages +  packages.ubuntu.libvirt }}"
```
maelk: There is no libvirt key anymore.

maelk:
passed the tests : https://jenkins.nordix.org/view/Airship/job/airship_metal3io_metal3_dev_env_integration_test/4/console

hardys:
Overall this looks good, thanks! 

One question though, does it make sense to nest the rhel7/rhel8 package lists under "centos"?

Sometimes the repo configuration or even package names are subtly different between centos and RHEL, so I wonder if it'd be cleaner to just have packages.ubuntu.packages, packages.centos7.packages, packages.rhel8.packages etc, and that also reflects more closely the platforms that are currently tested?

Not blocking, just wanted to ask that question, thanks for working on this :)

russellb:
@hardys thanks for taking a look!  I just came here to add you to ask you to take a look.  :-)

I agree with your feedback.  That change would make sense to me.  This seems to be working on centos7 and rhel8 right now.  Maybe we can do that as a follow-up?

russellb:
just got this failure on centos 7 --

Sep 09 19:37:43 |  [0;31m    "msg": "The conditional check 'ansible_lsb.major_release|int == 7' failed. The error was: error while evaluating conditional (ansible_lsb.major_release|int == 7): 'dict object' has no attribute 'major_release'\n\nThe error appears to be in '/opt/dev-scripts/metal3-dev-env/vm-setup/roles/packages_installation/tasks/main.yml': line 20, column 11, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n      block:\n        - name: Install packages on rhel7\n          ^ here\n"[0m

russellb:
I think I know why ... I removed installation of `redhat-lsb-core` from the script that runs the playbook since the playbook installs it, but that is required for this ansible module to work.  We still need that installed by the script before ansible runs.

russellb:
> I think I know why ... I removed installation of `redhat-lsb-core` from the script that runs the playbook since the playbook installs it, but that is required for this ansible module to work. We still need that installed by the script before ansible runs.

I see this is fine in metal3-dev-env, I just need to fix it in the other repo that uses these playbooks.

namnx228:
Thanks for your feedbacks.
What do you think if I change the playbook to have `packages.ubuntu.packages`, `packages.redhat_os.centos7.packages`, packages.redhat_os.rhel8.packages`, so we can also have `packages.redhat_os.pip` ?

russellb:
That sounds good, we can do it in a follow-up though.  Thanks for your patience on getting this merged!

PR #38, opened 2019-08-01
Parameterize the repos url and ref

maelk:
When cloning the Baremetal-Operator and CAPI-provider-baremetal
repositories, parameterize the git URL and branch ref to allow
specifying fork repositories and branches for testing in CI. Add
a FORCE_REPO_UPDATE flag to delete the repositories to clean
previous states. Defaults to false.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -10,22 +10,36 @@ M3PATH="${GOPATH}/src/github.com/metal3-io"
 BMOPATH="${M3PATH}/baremetal-operator"
 CAPBMPATH="${M3PATH}/cluster-api-provider-baremetal"
 
+BMOREPO="${BMOREPO:-https://github.com/metal3-io/baremetal-operator.git}"
+BMOBRANCH="${BMOBRANCH:-master}"
+CAPBMREPO="${CAPBMREPO:-https://github.com/metal3-io/cluster-api-provider-baremetal.git}"
+CAPBMBRANCH="${CAPBMBRANCH:-master}"
+FORCE_REPO_UPDATE="${FORCE_REPO_UPDATE:-false}"
+
 function clone_repos() {
     mkdir -p ${M3PATH}
+    if [[ -d ${BMOPATH} && "${FORCE_REPO_UPDATE}"=="true" ]]; then
+      rm -rf ${BMOPATH}
+    fi
     if [ ! -d ${BMOPATH} ] ; then
         pushd ${M3PATH}
-        git clone https://github.com/metal3-io/baremetal-operator.git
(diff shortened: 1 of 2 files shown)


03_launch_mgmt_cluster.sh
```diff
@@ -10,22 +10,36 @@ M3PATH="${GOPATH}/src/github.com/metal3-io"
 BMOPATH="${M3PATH}/baremetal-operator"
 CAPBMPATH="${M3PATH}/cluster-api-provider-baremetal"
 
+BMOREPO="${BMOREPO:-https://github.com/metal3-io/baremetal-operator.git}"
+BMOBRANCH="${BMOBRANCH:-master}"
+CAPBMREPO="${CAPBMREPO:-https://github.com/metal3-io/cluster-api-provider-baremetal.git}"
+CAPBMBRANCH="${CAPBMBRANCH:-master}"
+FORCE_REPO_UPDATE="${FORCE_REPO_UPDATE:-false}"
```
russellb: Can you document these in `config_example.sh` ?

maelk: ok, done

maelk:
@russellb Any further comments ? thanks

russellb:
Sorry, been on vacation all this week.

config_example.sh
```diff
@@ -15,3 +15,49 @@
...
+#
+# Select the Container Runtime, can be "podman" or "docker"
+# Defaults to "podman"
+#
+#export CONTAINER_RUNTIME="podman"
+
+#
+# Set the metal3 go path
+#
+#export M3PATH="${GOPATH}/src/github.com/metal3-io"
```
russellb: Do you expect this to ever be changed?

russellb: Looks like the script doesn’t let it be changed

config_example.sh
```diff
@@ -15,3 +15,49 @@
...
+
+#
+# Set the Baremetal operator go path
+#
+#export BMOPATH="${M3PATH}/baremetal-operator"
+
+#
+# Set the Baremetal cluster api provider go path
+#
+#export CAPBMPATH="${M3PATH}/cluster-api-provider-baremetal"
```
russellb: Same with these two

config_example.sh
maelk: I don't think they would be changed. Even if we work with forks of the repositories of Metal3, the location should remain the same IMO. Only which repository we actually clone would change. I can anyways parameterize it if you prefer.

maelk: sorry, those should not be there, they should not be changing indeed.

maelk:
fixed and squashed. Thank you

PR #42, opened 2019-08-12
Fix shellcheck errors

maelk:
Some small fixes to remove shellcheck errors, mostly quoting variables

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -1,7 +1,9 @@
 #!/bin/bash
 set -xe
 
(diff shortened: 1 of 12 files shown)


03_launch_mgmt_cluster.sh
```diff
@@ -1,7 +1,9 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
russellb: Why did this one need to change but not some of the others?

maelk: It did not, that's a mess up, sorry. fixed and squashed

PR #43, opened 2019-08-13
add option to run local operators

maelk:
Instead of using 'make deploy' for Baremetal Operator and CAPI
provider baremetal, add an option to use 'make run' and log the
output.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -9,6 +9,20 @@ source lib/common.sh
 eval "$(go env)"
 export GOPATH
 
+# Environment variables
+# M3PATH : Path to clone the metal3 dev env repo
+# BMOPATH : Path to clone the baremetal operator repo
+# CAPBMPATH: Path to clone the CAPI operator repo
+#
+# BMOREPO : Baremetal operator repository URL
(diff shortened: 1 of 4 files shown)


maelk:
Add a check skip, if the operators are running locally, we should not verify the number of replicas running in the kubernetes cluster (they should be 0)

03_launch_mgmt_cluster.sh
```diff
@@ -18,6 +19,9 @@ CAPBMREPO="${CAPBMREPO:-https://github.com/metal3-io/cluster-api-provider-bareme
 CAPBMBRANCH="${CAPBMBRANCH:-master}"
 FORCE_REPO_UPDATE="${FORCE_REPO_UPDATE:-false}"
 
+BMO_RUN_LOCAL="${BMO_RUN_LOCAL:-false}"
+CAPBM_RUN_LOCAL="${CAPBM_RUN_LOCAL:-false}"
```
russellb: Can you add docs for these new variables to `config_user.sh` ?

maelk: done. I also added a test to check that they are running locally

russellb:
This is working for me.  I needed operator-sdk installed, so I added that to our requirements install script too.

PR #45, opened 2019-08-16
Add retry loop for deployment verification and test script

maelk:
- add a retry loop, waiting for the setup to complete (for example for the baremetal hosts to get ready etc.)
- add a test script that provisions two machines 
- some shell linting fixes in lib/common.sh

Issue #39

### DIFF
--- a/04_verify.sh
+++ b/04_verify.sh
@@ -2,10 +2,6 @@
 
 set -u
 
-# Redirect to stdout for logging
-# Workaround to avoid returning logs in functions
-exec 3>&1
-
 # shellcheck disable=SC1091
 source lib/logging.sh
 # shellcheck disable=SC1091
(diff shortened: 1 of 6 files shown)


russellb:
I tried this and it seemed to work.  Could we change it so it doesn't like "FAIL" while waiting for something that we know will take a while?  Can it just indicating that it's still waiting, and not say FAIL unless it's giving up and considering it a failure?

maelk:
@russellb Thank you for the comment. I have now refactored the code. It waits before either succeeding or failing, but does not say FAIL in between anymore

russellb:
Thanks.  I tried this again today.  The output looks much better.  I did get two failures:

```
FAIL - centos-1 baremetal host in correct state : provisioned
...
FAIL - centos-1 baremetal host reachable by ssh
```

and that machine was provisioned successfully, so maybe the timeout just needs to be increased a bit more.

maelk:
@russellb Thank you. I have increased the number of retries, the timeout will happen after 20 minutes. This can be configured with environment variables, so I added those in config_example.sh

russellb:
OK - tested again and this is working for me. My only minor suggestion is that it would be nice to know what task is being waited on.  Right now you don't find out until it's done.

Thanks!

PR #46, opened 2019-08-16
Kinder replace minikube/namnx228

namnx228:
In this pull request, I add `kinder ` as another option to create K8S. `kinder` is necessary for CI as it is more lightweight and runs faser than  `minikube` . 

By default, `minikube` is still used to deploy K8S in the dev-env.

Note: This PR is developed on top of PR: `improvement/namnx228-download-1-place`. Therefore, please review that branch before this one.


russellb:
Thanks for the PR!  I don't mean to ignore this, I'm just waiting until we finish #37 since this depends on it

nordixinfra:

russellb:
It looks like this needs a rebase, and then i'll try this out, thanks!

maelk:
@russellb this is now rebased, thanks

russellb:
sorry, i introduced a conflict in the Makefile.  I tried to fix it but can't push to the branch.  (github gives you the choice to allow maintainers to push to your branch for a PR)

01_install_requirements.sh
```diff
@@ -59,3 +49,51 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
...
+      chmod +x docker-machine-driver-kvm2
+      sudo mv docker-machine-driver-kvm2 /usr/local/bin/.
+  fi
+else
+# Install kinder to replace minikube
+  if [ ! -d ${GOPATH}/src/kubeadm ]; then
+    pushd ${GOPATH}/src/
+    git clone https://github.com/kubernetes/kubeadm.git
+    pushd kubeadm
+    popd
```
russellb: looks like these last 2 lines don't do anything

01_install_requirements.sh
```diff
@@ -8,18 +9,7 @@ else
...
-    sudo mv minikube /usr/local/bin/.
-fi
-
-if ! which docker-machine-driver-kvm2 2>/dev/null ; then
-    curl -LO https://storage.googleapis.com/minikube/releases/latest/docker-machine-driver-kvm2
-    chmod +x docker-machine-driver-kvm2
-    sudo mv docker-machine-driver-kvm2 /usr/local/bin/.
-fi
-
+
```
russellb: extra blank line

01_install_requirements.sh
```diff
@@ -59,3 +49,51 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
...
+    pushd ${GOPATH}/src/
+    git clone https://github.com/kubernetes/kubeadm.git
+    pushd kubeadm
+    popd
+    popd
+  fi
+  
+  # if [ ! -d ${GOPATH}/src/sigs.k8s.io/kind ]; then
+  #   GO111MODULE="on" go get -u sigs.k8s.io/kind@v0.4.0
+  # fi
```
russellb: want to remove this?  or add a comment about why it's here?

01_install_requirements.sh
```diff
@@ -59,3 +49,51 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
...
+  
+  if [ ! -f ${GOPATH}/bin/kinder ]; then
+    pushd $HOME/go/src/kubeadm/kinder
+    GO111MODULE=on go install
+    popd
+  fi
+  
+  sudo docker pull kindest/node:v1.15.0
+  
+  # sudo usermod -aG docker $USER
```
russellb: want to remove this?  or add a comment about why it's here?

02_configure_host.sh
```diff
@@ -63,14 +63,14 @@ else
       if [ ! -e /etc/sysconfig/network-scripts/ifcfg-provisioning ] ; then
           echo -e "DEVICE=provisioning\nTYPE=Bridge\nONBOOT=yes\nNM_CONTROLLED=no\nBOOTPROTO=static\nIPADDR=172.22.0.1\nNETMASK=255.255.255.0" | sudo dd of=/etc/sysconfig/network-scripts/ifcfg-provisioning
       fi
-      sudo ifdown provisioning || true
-      sudo ifup provisioning
+      sudo ip link set provisioning down || true
+      sudo ip link set provisioning up || true
```
russellb: Why is this changed?

russellb: same question about the other places where `ifdown`/`ifup` were changed to `ip link set up` / `down`

03_launch_mgmt_cluster.sh
```diff
@@ -79,6 +79,20 @@ function launch_minikube() {
     minikube start
 }
 
+function launch_kinder() {
+  kinder="${GOPATH}/bin/kinder"
```
russellb: shouldn't this already be in `$PATH` ?

03_launch_mgmt_cluster.sh
```diff
@@ -79,6 +79,20 @@ function launch_minikube() {
...
+  if [[ $(sudo $kinder get clusters) == "" ]]; then
+    sudo $kinder create cluster
+    sudo $kinder do kubeadm-init
+    kinder_kubeconfig_path="$(sudo $kinder get kubeconfig-path)"
+    sudo rm -f ${HOME}/.kube/config
+    mkdir -p ${HOME}/.kube/
+    sudo mv $kinder_kubeconfig_path ${HOME}/.kube/config
+    sudo chown ${USER}:${USER} ${HOME}/.kube/config
+    #TODO: Attach `provisioning` interface to node which runs Ironic service.
```
russellb: not sure I understand the TODO here -- Ironic is running on the host where the `provisioning` bridge is located -- can you expand on what's left to do here?

namnx228:
Hi Russelb, I have enabled that permission for you.

config_example.sh
```diff
@@ -57,6 +57,9 @@
 #
 #export CAPBM_RUN_LOCAL=true
 
+#Set Kubenetes cluster
```
russellb: Can you add a comment about what the choices are?

default: `minikube`, also supported: `kinder` ?

lib/common.sh
```diff
@@ -2,6 +2,7 @@
 
 eval "$(go env)"
 
+
```
russellb: extra blank line added here

vm-setup/roles/libvirt/defaults/main.yml
```diff
@@ -23,7 +23,5 @@ extradisks_list:
 # size of the disks to create when using extradisks
 extradisks_size: 8G
 
-
-
```
russellb: unrelated changes

maelk:
Please ignore last ci failure. it is unrelated to the patchset

namnx228:
@russellb , thanks for your review, I will update the fix soon

01_install_requirements.sh
namnx228: Ok, I will remove them

01_install_requirements.sh
namnx228: Solved!

01_install_requirements.sh
namnx228: removed it!

01_install_requirements.sh
namnx228: removed!!

03_launch_mgmt_cluster.sh
namnx228: Yes, but not in $PATH of root user. Therefore, I use the full path to run Kinder with sudo right.

03_launch_mgmt_cluster.sh
namnx228: Sorry, my mistake. 
Previously, I thought that if we move Ironic to run inside a container, we will need `provisioning` bridge to make it visible to the network outside the cluster. But it seems to be my misunderstanding.

config_example.sh
namnx228: Done

02_configure_host.sh
namnx228: ifdown/ifup doesn't work without net-tools package in Ubuntu, so and just change use ip. It work fine on all linux distributions, and we don't need to install additional packages for it.

russellb: I don't think those commands are equivalent, though?

namnx228: Yeah, but in context of dev-env, I don't see any problems when we use `ip link set`. 
Do you think are there any problems with it ? But If you really think that this change is not problematic, I will drop it in this pull request.

namnx228:
This pull request is still in testing process

01_prepare_host.sh
```diff
@@ -13,19 +13,7 @@ else
   # shellcheck disable=SC1091
   source centos_install_requirements.sh
 fi
-
-if ! which minikube 2>/dev/null ; then
```
maelk: The access rights on this file have changed!

maelk: That's why the CI failed

namnx228: What is the original access rights on this file ?

namnx228: Fixed. Please trigger the test

02_configure_host.sh
russellb: I'd at least rather see it in a separate PR since it doesn't seem related to adding kinder support

03_launch_mgmt_cluster.sh
russellb: ah, OK, makes sense.

config_example.sh
russellb: looks like an addition is still needed to clarify which one is the default

PR #54, opened 2019-09-10
Cleanups and fixes for RHEL8

hardys:
This makes use of the recently added packages_installation role to install the needed packages on RHEL8, I also removed some unused packages for all platforms (lolcat/figlet) and moved the packages map definition to a more typical roles/default/main.yml location.

WIP as this has only been tested on RHEL8 so far, and not yet tested e2e with dev-scripts.

### DIFF
--- a/centos_install_requirements.sh
+++ b/centos_install_requirements.sh
@@ -16,14 +16,17 @@ fi
 sudo yum -y update
 
 # Install additional repos as needed for each OS version
+source /etc/os-release
+# VERSION_ID can be "7" or "8.x" so strip the minor version
+DISTRO="${ID}${VERSION_ID%.*}"
 if [ ! -f /etc/yum.repos.d/epel.repo ] ; then
-    if grep -q "Red Hat Enterprise Linux release 7" /etc/redhat-release ; then
+    if [[ $DISTRO == "rhel7" ]]; then
         sudo yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
-    elif grep -q "Red Hat Enterprise Linux release 8" /etc/redhat-release ; then
+    elif [[ $DISTRO == "rhel8" ]]; then
         sudo subscription-manager repos --enable=ansible-2-for-rhel-8-x86_64-rpms
         sudo alternatives --set python /usr/bin/python3
     else
-        sudo yum -y install epel-release --enablerepo=extras
+        sudo yum -y install epel-release dnf --enablerepo=extras
     fi
 fi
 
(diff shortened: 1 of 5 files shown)


hardys:
cc @namnx228 - this moves the vars file added in #37 to to a role default, I wasn't sure if there was a reason you defined that at the top-level though?

hardys:
@russellb - this is working for me on RHEL8 now, I'm thinking we can probably switch the 01 script in dev-scripts to source centos_install_requirements.sh and remove some duplicated things?

Also it seems there are delorean packages for python3-virtualbmc now, so we can remove the from-source install of that too.

russellb:
> @russellb - this is working for me on RHEL8 now, I'm thinking we can probably switch the 01 script in dev-scripts to source centos_install_requirements.sh and remove some duplicated things?
> 
> Also it seems there are delorean packages for python3-virtualbmc now, so we can remove the from-source install of that too.

sounds good to me.  want to put up a dev-scripts PR that uses this to check centos7?

hardys:
> sounds good to me. want to put up a dev-scripts PR that uses this to check centos7?

https://github.com/openshift-metal3/dev-scripts/pull/804 pushed, testing still in-progress, in particular I've not yet proved it works on a fresh centos install but lets see if it passes CI.

I think we can probably also remove the dev-scripts requirement for yq now, so we can potentially remove the need to pip install anything.

elfosardo:
can we remove ansible and wget from the list since they're already installed by the requirements script?
also wondering if we should rename centos_install_requirements.sh

hardys:
> can we remove ansible and wget from the list since they're already installed by the requirements script?

They aren't installed in the metal3-dev-env 01_install_requirements.sh, and the idea is to remove that from the dev-scripts one too, ref openshift-metal3/dev-scripts#804

We could rename the requirements script, but that seems cosmetic - what would we call it centos_rhel_requirements.sh?

hardys:
Oh sorry elfosardo I see what you mean, yes we could remove them from the ansible packages data, not sure that gains us much, but I guess if someone wanted to use an inventory that contained nodes other than the virthost it's one less requirement that gets installed.

namnx228:
> cc @namnx228 - this moves the vars file added in #37 to to a role default, I wasn't sure if there was a reason you defined that at the top-level though?

Yes, You can move like this. 
At the beginning, I defined `package_installation` as the top-level for increasing the scalability for the playbook.  It is easier to add and manage variables if we keep a structure like that. But if it is not necessary to do much update to these variables, you can move it as you have done.

centos_install_requirements.sh
```diff
@@ -27,7 +30,7 @@ fi
 # Install required packages
 # python-{requests,setuptools} required for tripleo-repos install
 sudo yum -y install \
-  python-pip \
```
namnx228: Is python-pip installed in RHEL7 and RHEL8 by default?

hardys: No it's not, but it will be installed via packages_installation before we attempt any pip installs, so I removed it here and added python3-pip for RHEL8 (packages_installation already defined python-pip for centos/RHEL7 but that won't work on RHEL8)

vm-setup/roles/packages_installation/defaults/main.yml
```diff
@@ -86,6 +83,9 @@ packages_installation:
          - nmap 
          - patch 
          - psmisc 
+         - python3-pip
```
hardys: @namnx228  FYI, and python-pip is already defined for el7 above - I checked in the role tasks and we'll do the RPM install before the pip install task so this should be OK I think?

hardys:
Ok https://github.com/openshift-metal3/dev-scripts/pull/804 passed and I've tested this locally on CentOS7 and RHEL8 so removing the WIP, ready for review.

vm-setup/roles/packages_installation/defaults/main.yml
namnx228: Yes, I think it is okay. 
In Ubuntu, when I installed Ansible by apt, it didn't get the latest version, but a version which leaded to some errors. That is the reason why I used pip to install Ansible on both Ubuntu and Centos to get the latest version or a specific version that we know it works properly. 
But I believe that softwares version installed by RPM  are tested extremely carefully, so it is okay if we install Ansible using yum.

PR #55, opened 2019-09-16
Add more verbose messages for test failures

maelk:
Prints the variables if the test fails

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -195,7 +195,11 @@ process_status(){
 # - second input to compare
 #
 equals(){
-  [[ "${1}" == "${2}" ]]; process_status "$?"
+  [[ "${1}" == "${2}" ]]; RET_CODE="$?"
+  if ! process_status "$RET_CODE" ; then
+    echo "       expected ${2}, got ${1}"
+  fi
+  return $RET_CODE
 }
 
 #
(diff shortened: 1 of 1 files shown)


lib/common.sh
```diff
@@ -196,6 +196,11 @@ process_status(){
 #
 equals(){
   [[ "${1}" == "${2}" ]]; process_status "$?"
+  RET_CODE="$?"
```
hardys: Since you're not using RESULT_STR, would it be cleaner to just not use process_status here and do something like:

```
equals(){
  if [[ "${1}" != "${2}" ]]; then
    echo "ERROR: expected ${2}, got ${1}"
    FAILS=$((FAILS+1))
  fi
}
```

maelk: process_status is printing the RESULT_STR that is set before calling the equals function, so we either use process_status or we need to add RESULT_STR in the output.

hardys: Ok I see now - it still looks a bit weird, because RET_CODE is saving the return code from process_status, not the original return-code from the test, would it be clearer to do something like:

```
equals(){
  [[ "${1}" != "${2}" ]]; RET_CODE="$?
  if ! process_status $RET_CODE; then
    echo "        expected ${2}, got ${1}"
  fi
}
```

maelk: ok, I'll fix it. thank you

lib/common.sh
```diff
@@ -206,7 +210,11 @@ equals(){
 # - String to look for the substring in
 #
 is_in(){
-  [[ "${2}" == *"${1}"* ]]; process_status "$?"
+  [[ "${2}" == *"${1}"* ]]; RET_CODE="$?"
+  if ! process_status "$?" ; then
```
hardys: This needs to be process_status $RET_CODE, $? here will always be 0 because the RET_CODE assignment is never going to fail:

```
$ cat noexist
cat: noexist: No such file or directory
$ echo $?
1

$ cat noexist; FOO=$?
cat: noexist: No such file or directory
$ echo $?
0
```

maelk:
fixed, thank you @hardys !

PR #56, opened 2019-09-18
move minikube initialization to step 01

maelk:
this is to optimize CI runs where 01 requirements are baked in

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -59,3 +59,23 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
   -e "virthost=$HOSTNAME" \
   -i vm-setup/inventory.ini \
   -b -vvv vm-setup/install-package-playbook.yml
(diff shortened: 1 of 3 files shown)


russellb:
I appreciate the CI optimization.  It seems weird to have this in a script called "install requirements" though.  It seems to fit better with the stuff in the 2nd script.  Or we could rename script 1?

maelk:
For the CI, we are trying to bake in all the expensive download only steps in the image. We do not want to run the 2nd script as it does mostly setup that needs to be validated. Renaming seems then the best option. Maybe something like 01_prepare_host.sh ?

russellb:
`prepare_host` sounds good.

maelk:
@russellb  ok, Updated, and created https://github.com/metal3-io/metal3-io.github.io/pull/92 and https://github.com/Nordix/airship-dev-tools/pull/74

PR #59, opened 2019-09-23
Fix libvirt group issue for Minikube

maelk:
The move of Minikube broke the deployment since the user is not part of the libvirt group. This commit adds the user to libvirt group.

Note : We are adding the provisioning interface to Minikube before creating the provisioning network. Any attempt to start again Minikube before running the 02_xx script will fail.


russellb:
Love seeing both CI jobs passing :-)

PR #60, opened 2019-09-24
Prevent missing group issue with Minikube in CI

maelk:
When starting minikube in the same script as we add the user to
libvirt group, it fails. This runs the Minikube commands in a
new login shell, and adds sudo for virsh where needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ verify:
 clean: delete_mgmt_cluster host_cleanup
 
 delete_mgmt_cluster:
-	minikube delete || true
+	sudo minikube delete || true
 
 host_cleanup:
 	./host_cleanup.sh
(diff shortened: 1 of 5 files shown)


maelk:
we cannot use newgrp, because that spawns a new shell. the script will not continue, even with exec. and minikube should not be started as root, so the best way we could come up with was using this "sudo su -l -c 'minikube start' ${USER}"

Makefile
```diff
@@ -15,7 +15,7 @@ verify:
 clean: delete_mgmt_cluster host_cleanup
 
 delete_mgmt_cluster:
-	minikube delete || true
+	sudo minikube delete || true
```
russellb: why not the `sudo su` approach on this one?

maelk: there is no need for it here, this can be run as sudo.

russellb: why does it need sudo at all, since we're still starting it as a regular user?

PR #62, opened 2019-09-25
Ensure clean before env setup

mkth1:


nordixinfra:

Makefile
```diff
@@ -1,4 +1,4 @@
-all: install_requirements configure_host launch_mgmt_cluster verify
+all: clean install_requirements configure_host launch_mgmt_cluster verify
```
hardys: I don't think this fits typical makefile usage, you'd always do e.g `make clean all` and just `make` could be run non-destructively in the case where something fails during the `make` process?

russellb: agreed - I'd suggest trying to detect an unclean environment in one of the scripts and giving a helpful message.

mkth1: @hardys the first approach was adding in the documentation to run make clean before consecutive make in case of error
https://github.com/metal3-io/metal3-io.github.io/pull/93/files

@russellb I did think about the script and helpful message but the script to detect error and giving helpful message would be more efficient if we would have more than one message to display, in our case it is always make clean if the we get error with make?

hardys: My point is that as a developer you might in some cases not want to `make clean` in the case of error.

Also for anyone who's used `make` in the context of software development this behaviour is probably unexpected, so personally I'd prefer to just document running `make clean` explicitly.

hardys: What is the specific problem you're trying to solve here, is there a script which fails when re-running?  Can we fix that so it's idempotent perhaps?

hardys:
@mkth1 how would you like to proceed with this one?  If you can describe the issues which prompted this change perhaps we can solve them in a different way and/or more clearly document that `make clean all` should be used in unclean environments?

As previously mentioned I think combining the `clean` target with the default `all` target is probably not what most users would expect, also ref https://www.gnu.org/prep/standards/html_node/Standard-Targets.html#Standard-Targets so perhaps we should close this without merging and look at alternatives?

russellb:
I'm going to close this for now.  I do think making the scripts more resilient so that running make a second time doesn't fail without cleaning, or at least gives more helpful output other than just something like "provisioning interface already exists".

Maybe you could file a bug about the error you see and we can track there?

PR #71, opened 2019-10-02
Add interface check for Minikube

maelk:
Check if the interfaces exist before adding them instead of
checking if Minikube VM exists

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -84,14 +84,19 @@ function init_minikube() {
     #If the vm exists, it has already been initialized
     if [[ "$(sudo virsh list --all)" != *"minikube"* ]]; then
       sudo su -l -c "minikube start" "$USER"
-      # The interface doesn't appear in the minikube VM with --live,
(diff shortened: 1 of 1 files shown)


01_prepare_host.sh
```diff
@@ -84,14 +84,19 @@ function init_minikube() {
...
-      # so just attach it and make it reboot. As long as the
-      # 02_configure_host.sh script does not run, the provisioning network does
-      # not exist. Attempting to start Minikube will fail until it is created.
+      sudo su -l -c "minikube stop" "$USER"
+    fi
+
+    # The interface doesn't appear in the minikube VM with --live,
+    # so just attach it and make it reboot. As long as the
+    # 02_configure_host.sh script does not run, the provisioning network does
+    # not exist. Attempting to start Minikube will fail until it is created.
```
russellb: I think this comment block should move down to before the `virsh attach-interface` command

maelk: fixed, thank you

PR #74, opened 2019-10-02
DNM - test out shellcheck prow job.

russellb:


russellb:
job failed because of real shellcheck warnings: https://prow.apps.ci.metal3.io/log?job=shellcheck&id=1179444780199841792

PR #78, opened 2019-10-07
Add v1alpha2 CAPI and CABPK controllers and CRDs

kashifest:
Adds CAPI v1alpha2 & CAPBPK controllers & CRDs in dev-env. The V1ALPHA2_SWITCH flag determines if we want to install the above mentioned entities in dev-env or not.

Fixes#69

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -27,10 +27,19 @@ M3PATH="${GOPATH}/src/github.com/metal3-io"
 BMOPATH="${M3PATH}/baremetal-operator"
 CAPBMPATH="${M3PATH}/cluster-api-provider-baremetal"
 
+CAPIPATH="${M3PATH}/cluster-api"
+CABPKPATH="${M3PATH}/cluster-api-bootstrap-provider-kubeadm"
+
 BMOREPO="${BMOREPO:-https://github.com/metal3-io/baremetal-operator.git}"
 BMOBRANCH="${BMOBRANCH:-master}"
 CAPBMREPO="${CAPBMREPO:-https://github.com/metal3-io/cluster-api-provider-baremetal.git}"
 CAPBMBRANCH="${CAPBMBRANCH:-master}"
+
+CAPIREPO="${CAPIREPO:-https://github.com/kubernetes-sigs/cluster-api.git}" 
+CAPIBRANCH="${CAPIBRANCH:-master}"
+CABPKREPO="${CABPKREPO:-https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm.git}" 
+CABPKBRANCH="${CABPKBRANCH:-master}"
+
 FORCE_REPO_UPDATE="${FORCE_REPO_UPDATE:-false}"
 
 BMO_RUN_LOCAL="${BMO_RUN_LOCAL:-false}"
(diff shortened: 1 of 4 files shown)


03_launch_mgmt_cluster.sh
```diff
@@ -113,8 +147,35 @@ function launch_cluster_api() {
...
+    popd
+}
+
 clone_repos
 sudo su -l -c 'minikube start' "${USER}"
 launch_baremetal_operator
 apply_bm_hosts
-launch_cluster_api
+launch_cluster_api_provider_baremetal
+if [ "${V1ALPHA2_SWITCH}" == true ]; then
```
maelk: V1ALPHA2_SWITCH should maybe be defined in lib/common.sh to default to false ? Then it would not be needed to be defined in 04_verify.sh

04_verify.sh
```diff
@@ -141,15 +143,26 @@ check_container(){
 }
 
 KUBECONFIG="${KUBECONFIG:-${HOME}/.kube/config}"
+V1ALPHA2_SWITCH=${V1ALPHA2_SWITCH:-"false"}
```
maelk: move it to lib/common.sh ?

03_launch_mgmt_cluster.sh
kashifest: Done

03_launch_mgmt_cluster.sh
```diff
@@ -113,9 +147,36 @@ function launch_cluster_api() {
...
 }
 
+#
+# Launch the cluster-api-controller manager (v1alpha2) in the system namespace.
+#
+
+function launch_core_cluster_api() {
+    pushd "${CAPIPATH}"
+      make generate
+      sed -i'' 's/capi-system/metal3/' config/default/kustomization.yaml
```
maelk: maybe the namespace could be created instead of doing this sed command

03_launch_mgmt_cluster.sh
```diff
@@ -113,9 +147,36 @@ function launch_cluster_api() {
...
+    popd
+}
+
+#
+# Launch the cluster-api-bootstrap-provider-kubeadm-controller manager (v1alpha2) in the metal3 namespace.
+#
+
+function launch_cluster_api_bootstrap_provider_kubeadm() {
+    pushd "${CABPKPATH}"
+      sed -i'' 's/cabpk-system/metal3/' config/default/kustomization.yaml
```
maelk: same here

03_launch_mgmt_cluster.sh
kashifest: Creating the namespace is not an issue since that is taken care of by the corresponding installation processes for each repo. I kept this under 'metal3' since other controllers and components were defined in the same namespace. Besides a different namespace for each controller would require changing the verify script also since then verification of deployments , statefulsets etc would require the namespace parameter.

03_launch_mgmt_cluster.sh
kashifest: Please see the previous comment

russellb:
Needs a rebase

kashifest:
> Needs a rebase

Done

russellb:
I'm fine with this.  Thanks for adding it and making it optional.  /lgtm

PR #90, opened 2019-10-16
Pre-pull images in Minikube

kashifest:
This PR pre-pulls images in minikube to reduce the baremetal-operator pod-initialization time.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -62,9 +62,13 @@ export SUSHY_TOOLS_IMAGE=${SUSHY_TOOLS_IMAGE:-"quay.io/metal3-io/sushy-tools"}
 
 # Ironic vars
 export IPA_DOWNLOADER_IMAGE=${IPA_DOWNLOADER_IMAGE:-"quay.io/metal3-io/ironic-ipa-downloader:master"}
-export IRONIC_IMAGE=${IRONIC_IMAGE:-"quay.io/metal3-io/ironic:master"}
+export IRONIC_IMAGE=${IRONIC_IMAGE:-"quay.io/metal3-io/ironic"}
+export IRONIC_INSPECTOR_IMAGE=${IRONIC_INSPECTOR_IMAGE:-"quay.io/metal3-io/ironic-inspector"}
 export IRONIC_DATA_DIR="$WORKING_DIR/ironic"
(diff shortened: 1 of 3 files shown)


lib/common.sh
```diff
@@ -62,10 +62,14 @@ export SUSHY_TOOLS_IMAGE=${SUSHY_TOOLS_IMAGE:-"quay.io/metal3-io/sushy-tools"}
 
 # Ironic vars
 export IPA_DOWNLOADER_IMAGE=${IPA_DOWNLOADER_IMAGE:-"quay.io/metal3-io/ironic-ipa-downloader:master"}
-export IRONIC_IMAGE=${IRONIC_IMAGE:-"quay.io/metal3-io/ironic:master"}
+export IRONIC_IMAGE=${IRONIC_IMAGE:-"quay.io/metal3-io/ironic"}
+export IRONIC_INSPECTOR_IMAGE=${IRONIC_INSPECTOR_IMAGE:-"quay.io/metal3-io/ironic-inspector"}
```
russellb: Can you retain the `master` tag, just so we're explicit about which git branch we're following?

russellb: OK, good point.  If that's what the operator yaml is using then let's keep them consistent.  You can leave it as is.

russellb:
lgtm, but another set of eyes from @hardys would be good

maelk:
lgtm to me too

kashifest:
> lgtm but now needs a rebase
> 
> Another nice enhancement would be the ability to cache images in a local registry and/or build them from source automatically - @derekhiggins added something to do that in https://github.com/openshift-metal3/dev-scripts/blob/master/04_setup_ironic.sh#L19 which can perhaps be refactored to work with both docker and podman then moved to metal3-dev-env.

Thanks @hardys . The rebase is done. I will check the link you provided

hardys:
> Thanks @hardys . The rebase is done. I will check the link you provided

Thanks!  The link was just FYI - there are two parts we might consider, first is to configure docker-distribution on the test host (which could be configured as a pull-through cache to avoid repeatedly downloading the container images on the ephemeral minikube vm)

The second is to automatically build images from git and push to the local registry, which would be useful for testing the ironic images etc.  I'll try to spend some time looking at if we can easily reuse some of what has already been done in the dev-scripts repo.

PR #97, opened 2019-10-18
Add iptables rule to allow mdns for ironic on provisioning

maelk:
Now using new version of ironic, IPA uses mdns to resolve the ironic endpoint. Adding port 5353 in iptables rules.

and small code refactoring for iptables commands. Also adding an out interface match on provisioning rules

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -111,15 +111,15 @@ fi
 # Add firewall rules to ensure the IPA ramdisk can reach httpd, Ironic and the
 # Inspector API on the cluster, and ironic can reach the agent (9999)
 for port in 6180 5050 6385 9999; do
-    if ! sudo iptables -C FORWARD -i provisioning -p tcp -m tcp --dport $port \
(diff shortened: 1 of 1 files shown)


hardys:
We may want to consider delaying this, as we discovered recently that the Ironic mdns implementation doesn't yet support ipv6.

So https://github.com/metal3-io/ironic-image/pull/106 was proposed yesterday to disable it again, until that is resolved.

What do you think, should we leave it enabled and use a special configuration for ipv6 testing (which several of us are starting to work on now), or should we avoid using mdns for the ironic endpoints until it can support both ipv4 and ipv6?

hardys:
@maelk ack thanks - my preference would be to land this, ask @stbenjam to rebase https://github.com/metal3-io/metal3-dev-env/pull/91 to include the changes here, but also to merge metal3-io/ironic-image#106 so we can continue to test with a single image for both ipv4 and ipv6

Then when Ironic mdns is ready for ipv6 we can re-enable it

PR #99, opened 2019-10-21
Add python3 openstack client

kashifest:
This PR adds opentstackclient using python3 in ubuntu environment

### DIFF
--- a/ubuntu_install_requirements.sh
+++ b/ubuntu_install_requirements.sh
@@ -18,7 +18,6 @@ sudo apt -y update
 # Install required packages
 
 sudo apt -y install \
-  python-pip \
   zlib1g-dev \
   libssl1.0-dev \
   wget
@@ -47,6 +46,7 @@ if [[ $PATH != *pyenv* ]]; then
 fi
 
 pyenv install -s 2.7.5
+pyenv install -s 3.7.5
 pyenv versions
 pyenv global 2.7.5
 # There are some packages which are newer in the tripleo repos
--- a/vm-setup/roles/packages_installation/defaults/main.yml
+++ b/vm-setup/roles/packages_installation/defaults/main.yml
@@ -9,6 +9,7 @@ packages:
      - patch 
      - psmisc 
      - python-pip 
+     - python3-pip
      - wget
      - libvirt-bin 
      - libvirt-clients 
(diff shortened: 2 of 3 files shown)


ubuntu_install_requirements.sh
```diff
@@ -19,6 +19,7 @@ sudo apt -y update
 
 sudo apt -y install \
   python-pip \
+  python3-pip \
```
hardys: Do you need to install this here, or can it just be installed via ansible (since the package install happens before any pip installs):

https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/packages_installation/tasks/main.yml#L5

ubuntu_install_requirements.sh
```diff
@@ -47,6 +48,7 @@ if [[ $PATH != *pyenv* ]]; then
 fi
 
 pyenv install -s 2.7.5
+pyenv install -s 3.4.0
```
hardys: I'm not familiar with python3 status on Ubuntu, but does 3.4 support align with a particular release?

For CentOS/RHEL we use python2 on CentOS/RHEL7 and python3 on CentOS/RHEL8 so I'm wondering if we could do something similar for Ubuntu - mixing python versions could get confusing/fragile otherwise perhaps?

kashifest: For now, I am working with upgrading the whole dev-env to python3. I just checked that eventually python latest is now 3.8. I will check and update the code again.

ubuntu_install_requirements.sh
hardys: Ah I see, so in a future commit https://github.com/metal3-io/metal3-dev-env/blob/master/ubuntu_install_requirements.sh#L91 will get switched to pip3 - thanks for clarifying

kashifest: Yes apparently

ubuntu_install_requirements.sh
```diff
@@ -47,6 +48,7 @@ if [[ $PATH != *pyenv* ]]; then
 fi
 
 pyenv install -s 2.7.5
+pyenv install -s 3.8.0
```
elfosardo: I don't think there's any Linux distro that at the moment uses Python 3.8 as default
I suggest to use 3.7, also to align with upstream Openstack development

kashifest: @elfosardo Thanks for the review. Your suggestion is addressed.

PR #101, opened 2019-10-22
Upgrade to python3 in dev-env for ubuntu

kashifest:
Upgrade python in dev-env from python2 to python3 for ubuntu

Solves Issue# 88

### DIFF
--- a/ubuntu_install_requirements.sh
+++ b/ubuntu_install_requirements.sh
@@ -18,7 +18,6 @@ sudo apt -y update
 # Install required packages
 
 sudo apt -y install \
-  python-pip \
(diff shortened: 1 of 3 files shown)


ubuntu_install_requirements.sh
```diff
@@ -18,7 +18,6 @@ sudo apt -y update
 # Install required packages
 
 sudo apt -y install \
-  python-pip \
```
hardys: Where does pip3 get installed now this is removed?

kashifest: I checked that pyenv is installing the pip3 so we don't need it

PR #102, opened 2019-10-22
Changes bmh naming

jan-est:
This patch changes the naming of the baremetalhosts from master-0 and worker-0 to node-XX. Amount of nodes is parameterized.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -52,8 +52,7 @@ WORKING_DIR=${WORKING_DIR:-"/opt/metal3-dev-env"}
 NODES_FILE=${NODES_FILE:-"${WORKING_DIR}/ironic_nodes.json"}
 NODES_PLATFORM=${NODES_PLATFORM:-"libvirt"}
 
-export NUM_MASTERS=${NUM_MASTERS:-"1"}
-export NUM_WORKERS=${NUM_WORKERS:-"1"}
+export NUM_NODES=${NUM_NODES:-"2"}
 export VM_EXTRADISKS=${VM_EXTRADISKS:-"false"}
 
 # VBMC and Redfish images
--- a/vm-setup/roles/common/tasks/main.yml
+++ b/vm-setup/roles/common/tasks/main.yml
@@ -6,21 +6,13 @@
 - name: Define vm_nodes if not already defined
   when: generate_vm_nodes
   block:
-    - name: Generate vm_nodes for "{{num_masters}}" masters
+    - name: Generate vm_nodes for "{{num_nodes}}" nodes
       set_fact:
         vm_nodes: "{{vm_nodes|default([]) + [
-                     {'name': ironic_prefix + 'master_%s'|format(item),
-                      'flavor': 'master',
+                     {'name': ironic_prefix + 'node_%s'|format(item),
+                      'flavor': 'node',
                       'virtualbmc_port': virtualbmc_base_port+item}]}}"
-      loop: "{{ range(0, num_masters|int)|list }}"
-    
-    - name: Generate vm_nodes for "{{num_workers}}" workers
-      set_fact:
-        vm_nodes: "{{vm_nodes|default([]) + [
-                     {'name': ironic_prefix + 'worker_%s'|format(item),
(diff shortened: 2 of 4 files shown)


lib/common.sh
```diff
@@ -52,8 +52,7 @@ WORKING_DIR=${WORKING_DIR:-"/opt/metal3-dev-env"}
 NODES_FILE=${NODES_FILE:-"${WORKING_DIR}/ironic_nodes.json"}
 NODES_PLATFORM=${NODES_PLATFORM:-"libvirt"}
 
-export NUM_MASTERS=${NUM_MASTERS:-"1"}
-export NUM_WORKERS=${NUM_WORKERS:-"1"}
+export NUM_NODES=${NUM_NODES:-"2"}
```
maelk: this needs to be modified also in config_example.sh

hardys:
We need to figure out a way to not break https://github.com/openshift-metal3/dev-scripts which also consumes these roles - I'm fine with changing the defaults here but we can perhaps make it possible with vars to deploy more than one "flavor" of nodes?

Ideally we'll expose some vars which can be overridden to retain the current behavior, while changing the defaults to match what is proposed here:

https://github.com/openshift-metal3/dev-scripts/blob/master/vm_setup_vars.yml

The alternative is we fork the roles into dev-scripts, but I was trying to share code where possible to minimise the maintenance overhead.

vm-setup/roles/common/tasks/main.yml
```diff
@@ -6,21 +6,13 @@
...
-      loop: "{{ range(0, num_masters|int)|list }}"
-    
-    - name: Generate vm_nodes for "{{num_workers}}" workers
-      set_fact:
-        vm_nodes: "{{vm_nodes|default([]) + [
-                     {'name': ironic_prefix + 'worker_%s'|format(item),
-                      'flavor': 'worker',
-                      'virtualbmc_port': virtualbmc_base_port+num_masters|int+item} ]}}"
-      loop: "{{ range(0, num_workers|int)|list }}"
+      loop: "{{ range(0, num_nodes|int)|list }}"
```
hardys: Could we add a loop here to iterate over the defined flavors, and set facts for the counts by looking for a num_$flavorname variable, with defaults defined e.g 0 when the variable isn't set?

jan-est:
> We need to figure out a way to not break https://github.com/openshift-metal3/dev-scripts which also consumes these roles - I'm fine with changing the defaults here but we can perhaps make it possible with vars to deploy more than one "flavor" of nodes?
> 
> Ideally we'll expose some vars which can be overridden to retain the current behavior, while changing the defaults to match what is proposed here:
> 
> https://github.com/openshift-metal3/dev-scripts/blob/master/vm_setup_vars.yml
> 
> The alternative is we fork the roles into dev-scripts, but I was trying to share code where possible to minimise the maintenance overhead.

Sorry Steven about my slow reply, been working on the other issues past week. I agree that having way to deploy different flavor of nodes is definitely an advantage. Do you have any ideas how we should do this or any specific requirements we need to take into account?

hardys:
> Do you have any ideas how we should do this or any specific requirements we need to take into account?

Yes, as previously mentioned I think instead of changing the currently hard-coded references to master/worker for another hard-coded reference, we should instead iterate over the flavors defined e.g https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/common/defaults/main.yml#L18

That way we can still override the default flavors and have whatever number/naming is needed for testing (which may be just one by default for metal3-dev-env).

If you like I'm happy to take a closer look and propose an alternative PR for consideration?

jan-est:
> If you like I'm happy to take a closer look and propose an alternative PR for consideration?

Yes, that would be great!

hardys:
@jan-est sorry for the delay looking into this, but please check https://github.com/metal3-io/metal3-dev-env/pull/129 which shows the templating approach I described, if you're happy we can go ahead with that, which includes this patch and my additions as an additional commit.

hardys:
This was merged via https://github.com/metal3-io/metal3-dev-env/pull/129 so closing this PR

PR #108, opened 2019-10-24
Fix Kustomize installation

kashifest:
The latest kustomize installation link doesn't seem to have a binary but an archive and this breaks the installation of kustomize in dev-env. Changing the download link to use Kustomize version 3.2.3 instead for a quick fix

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -33,7 +33,7 @@ if ! command -v kubectl 2>/dev/null ; then
 fi
 
 if ! command -v kustomize 2>/dev/null ; then
-    curl -Lo kustomize "$(curl --silent -L https://github.com/kubernetes-sigs/kustomize/releases/latest 2>&1 | awk -F'"' '/linux_amd64/ { print "https://github.com"$2; exit }')"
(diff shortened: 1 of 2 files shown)


01_prepare_host.sh
```diff
@@ -33,7 +33,7 @@ if ! command -v kubectl 2>/dev/null ; then
 fi
 
 if ! command -v kustomize 2>/dev/null ; then
-    curl -Lo kustomize "$(curl --silent -L https://github.com/kubernetes-sigs/kustomize/releases/latest 2>&1 | awk -F'"' '/linux_amd64/ { print "https://github.com"$2; exit }')"
+    curl -Lo kustomize https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2Fv3.2.3/kustomize_kustomize.v3.2.3_linux_amd64
```
maelk: Can you please make the version a variable ?

kashifest: Done

maelk:
/lgtm
needed to fix our CI pipeline

PR #112, opened 2019-10-28
Parameterize host image and use Cirros or Ubuntu

maelk:


maelk:
I'm putting a PR for shellcheck fixes together

hardys:
AFAICS this removes the centos option completely?  I'm wondering if we should just include cirros by default, and document how to override the config for other distros in the config_$user.sh?

Alternatively it'd be nice to leave the centos option as some of us are actively testing things on Centos

maelk:
I have added the Centos option. I will submit a PR to change the try-it.html

config_example.sh
```diff
@@ -83,3 +83,31 @@
...
+#
+#export IMAGE_OS="Cirros"
+
+# Image for target hosts deployment
+#
+#export IMAGE_NAME="CentOS-7-x86_64-GenericCloud-1907.qcow2"
+
+# Location of the image to download
+#
+#export IMAGE_LOCATION="http://cloud.centos.org/centos/7/images"
```
hardys: We should probably show the sample IMAGE_NAME/IMAGE_LOCATION as the defaults, e.g now cirros?  That would also be more consistent with the IMAGE_OS above and the IMAGE_USERNAME

maelk: True, will fix! thank you

maelk:
Here is the documentation update : https://github.com/metal3-io/metal3-io.github.io/pull/104

maelk:
The ubuntu test was a false positive. Re-running it

maelk:
/hold 
CI Metal3-dev-env broken on Ubuntu

maelk:
/hold cancel
CI fixed

PR #114, opened 2019-10-29
Temporary workaround for kubernetes/minikube#5751

maelk:
Will be removed when Minikube 1.5.1 is released


maelk:
The tests passed ! Modifying the commit to remove the hack to make them pass on an old image. I will generate new images when this is merged

maelk:
The build of the new images is in progress

PR #117, opened 2019-10-30
Increase minikube memory to 4096MB

maelk:
We are running out of memory due to ironic being in cluster


maelk:
Running test-integration and test-centos-integration would not have any effect since it requires to delete and recreate minikube to take effect

PR #118, opened 2019-10-30
create ssh keys in lib/common and set environment variable

maelk:
The environment variable will be used to populate the userdata
in v1alpha2 using envsubst, rather than the current templating
in v1alpha1

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -46,7 +46,13 @@ fi
 
 export EXTERNAL_SUBNET="192.168.111.0/24"
 
-export SSH_PUB_KEY=~/.ssh/id_rsa.pub
+export SSH_KEY=${SSH_KEY:-"${HOME}/.ssh/id_rsa"}
+export SSH_PUB_KEY=${SSH_PUB_KEY:-"${SSH_KEY}.pub"}
+# Generate user ssh key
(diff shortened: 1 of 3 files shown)


lib/common.sh
```diff
@@ -46,7 +46,14 @@ fi
 
 export EXTERNAL_SUBNET="192.168.111.0/24"
 
-export SSH_PUB_KEY=~/.ssh/id_rsa.pub
+export SSH_KEY_PATH=~/.ssh/id_rsa
+export SSH_PUB_KEY_PATH="${SSH_KEY_PATH}.pub"
```
hardys: These are hard-coded and won't allow override from the config as in `config_example.sh`

Should be like this I think?

```
export SSH_KEY_PATH=${SSH_KEY_PATH:-~/.ssh/id_rsa}
export SSH_PUB_KEY_PATH=${SSH_PUB_KEY_PATH:-${SSH_KEY_PATH}.pub}
```

maelk: indeed, I will fix this, thank you

maelk:
Hi @hardys ! Any further comments ?

hardys:
> Hi @hardys ! Any further comments ?

Sorry @maelk I forgot to revisit this, lgtm

PR #119, opened 2019-10-31
modify 05_test.sh to use primary interface for ssh

maelk:
This does not require networking setup on the vm. Fixes broken tests with cirros.
The default image is set back to centos as the userdata.sh generates user data for centos.

### DIFF
--- a/05_test.sh
+++ b/05_test.sh
@@ -132,6 +132,18 @@ check_bmh_status(){
 }
 
 #
+# Checks if an ip is in a range
+#
+# Inputs:
+# - IP address to check (X.X.X.X)
+# - IP range (X.X.X.X,X.X.X.X)
+#
+# Usage ip_in_range ip start end
+ip_in_range() {
+  python ip_range_check.py "$1" "$2"
+}
+
+#
 # Get the IP of a vm from the dhcp leases
 #
 # Inputs:
@@ -139,10 +151,21 @@ check_bmh_status(){
 #
 get_vm_ip(){
   local BMH_NAME="${1}"
-  sudo virsh net-dhcp-leases baremetal | grep "${BMH_NAME}" | awk '{print $5}' \
-    | cut -d '/' -f1
-}
+  local DHCP_RANGE
+  local CONFIGMAP
 
+  # Get configmap for ironic operator
+  CONFIGMAP=$(kubectl get configmap -n metal3 | grep ironic-bmo-configmap | awk '{print $1}')
+
+  # Get the DHCP start and end values
+  DHCP_RANGE=$(kubectl get configmap -n metal3 "$CONFIGMAP" -o jsonpath='{.data.DHCP_RANGE}')
+  #Compare the bmh ips to the dhcp range
+  for ip in $(kubectl get bmh -n metal3 "${BMH_NAME}" -o yaml | grep "ip:" | awk '{print $3}'); do
+    if ip_in_range "$ip" "$DHCP_RANGE"; then
+      echo "$ip"
+    fi
+  done
+}
 
 # provision the machines
 for name in $MACHINES_LIST; do
--- a/ip_range_check.py
+++ b/ip_range_check.py
@@ -1,0 +1,9 @@
+#!/usr/bin/python
+
+from ipaddress import IPv4Address
+import sys
+dhcp_range=unicode(sys.argv[2]).split(',')
+if IPv4Address(dhcp_range[0]) <= IPv4Address(unicode(sys.argv[1])) <= IPv4Address(dhcp_range[1]):
+  sys.exit(0)
+else:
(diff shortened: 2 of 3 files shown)


05_test.sh
```diff
@@ -139,11 +157,27 @@ check_bmh_status(){
...
+  local DHCP_RANGE
+  local DHCP_START
+  local DHCP_STOP
+  local CONFIGMAP
+
+  # Get configmap for ironic operator
+  CONFIGMAP=$(kubectl get configmap -n metal3 | grep ironic-bmo-configmap | awk '{print $1}')
+
+  # Get the DHCP start and end values
+  DHCP_RANGE=$(kubectl get configmap -n metal3 "$CONFIGMAP" -o yaml | grep "DHCP_RANGE" | awk '{print $2}')
```
hardys: Could we pass either a go template or jsonpath to the kubectl here instead of the grep/awk filtering?

05_test.sh
```diff
@@ -131,6 +131,24 @@ check_bmh_status(){
...
+ip_to_int(){
+  local IP="${1}"
+  local INT_IP=0
+  local iteration=3
+  for i in $(echo "$IP" | tr '.' ' '); do
+    ((INT_IP=INT_IP+(i*(2**(iteration*8)))))
+    ((iteration=iteration-1))
+  done
+  echo "${INT_IP}"
+}
```
hardys: I wonder if bash is the best tool for the job here - I can see this works but it's not super clear?

How about using python for the IP comparison, e.g 

```
# Usage ip_in_range ip start end
ip_in_range() {
python -c "
import ipaddress
import sys
if int(ipaddress.IPv4Address(u'"$2"')) <= int(ipaddress.IPv4Address(u'"$1"')) <= int(ipaddress.IPv4Address(u'"$3"')):
  sys.exit(0)
else:
  sys.exit(1)
"
}

DHCP_START="172.22.0.100"
DHCP_END="172.22.0.200"
IP="172.22.0.150"
if ip_in_range $IP $DHCP_START $DHCP_END; then
    echo "OK"
else
    echo "FAIL"
fi
```

05_test.sh
maelk: done, thanks

05_test.sh
```diff
@@ -139,11 +151,22 @@ check_bmh_status(){
 #
 get_vm_ip(){
   local BMH_NAME="${1}"
-  sudo virsh net-dhcp-leases baremetal | grep "${BMH_NAME}" | awk '{print $5}' \
```
hardys: I just realized that these two IPs aren't the same - previously we'd get the IP for the baremetal network, e.g the one which is used *after* provisioning for the deployed OS.

The changes below switch it to get the provisioning IP, which is potentially only up during provisioning?

I just wanted to ensure that is what you expect, as we do things like `ssh_to_machine` which I think should target the host OS not e.g the deploy ramdisk where it will fail?

maelk: yes, switching to the provisioning network is done on purpose. The provisioning interface is the primary interface, so it will be configured by default by cloud-init (except if some specific cloud-init is given). The problem with the baremetal network is that you need to run a specific cloud-init to get it configured and it depends on the image you deploy with. This makes it more generic, as it does not rely on what is in the cloud-init. If we would want to check that the cloud-init ran properly, we can add tests over ssh.
About the deploy ramdisk, the credentials will be incorrect, so it will fail to ssh there.

maelk:
By default (if using cloud-init), it will since it is the first interface.

05_test.sh
hardys: @maelk thanks for confirming - sounds reasonable but do we know why the baremetal nic is not coming up, despite the dnsmasq configuration having a reservation for it, is it just because of the nic ordering?

My concern is that we only create dhcp reservations in the libvirt dnsmasq for the baremetal network:

https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/common/tasks/main.yml#L33

So if we switch to the provisioning network by default then we require the ironic dnsmasq to allocate and IP, and IIRC that is not a persistent allocation we can necessarily rely on?

I guess for testing it should be fine, but I know from testing on OpenShift that often you would normally expect to deploy via a dedicated provisioning nic, then have an "external" or controlplane nic come up on booting the OS, and at this point there's no access via the provisioning nic.

hardys: As a side note, @dtantsur mentioned that Ironic does have support for network_data metadata so perhaps in future we can control the desired nic for cloud-init to configure via metadata when https://github.com/metal3-io/baremetal-operator/issues/329 is implemented.

https://specs.openstack.org/openstack/nova-specs/specs/liberty/implemented/metadata-service-network-info.html

Not looked into cloud-init support for that data yet though.

PR #121, opened 2019-11-01
WIP: Size down the node vms

maelk:
The second node cannot start deploying if the first node is
deployed with 8GB of ram

### DIFF
--- a/vm-setup/roles/common/defaults/main.yml
+++ b/vm-setup/roles/common/defaults/main.yml
@@ -9,8 +9,8 @@ libvirt_nic_model: virtio
 # These defaults are used if there are no flavor-specific
 # overrides configured.
 default_disk: 50
-default_memory: 8192
-default_vcpu: 4
+default_memory: 4096
+default_vcpu: 2
 num_masters: 1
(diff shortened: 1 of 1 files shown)


hardys:
> The second node cannot start deploying if the first node is
deployed with 8GB of ram

Does this mean in CI or something?

Note that we have to be careful with the memory allocation, because the ironic-python-agent stores the disk image in memory during deployment, thus the required RAM is influenced by the qcow disk image size.

maelk:
@hardys yes, I have just ran into this problem. 
The first issue is that in CI the vms have 16GB of ram, and when we deploy v1alpha2, Minikube's and the first node's memory consumption add to more than 8GB, hence libvirt refuses to start the second node. But when trying things out with this PR, ironic fails to deploy, because the image is too big to fit in ram when there are only 4GBs. (our image is currently centos with updated cloud-init, 2GB)
We'll try to switch to ubuntu (400MB) then ironic should be able to deploy with 4GB of ram

maelk:
Made a parameter in #125

PR #124, opened 2019-11-07
Delete registry container during metal3-dev-env cleanup

fmuyassarov:


hardys:
Part of the reason for having this container is to cache images over multiple test runs, to speed up testing and reduce bandwidth usage.

What's the motivation for removing the container here?

I'm thinking we should add a persistent volume mounted on the host if we're going to do this?

We already do have logic which will restart the container if for any reason it's not running:

https://github.com/metal3-io/metal3-dev-env/blob/master/02_configure_host.sh#L129

fmuyassarov:
As the port 5000 is in use, when we rerun (after cleanup) the test - it fails. See below:

- - - - -
+ sudo podman rm registry -f
ae692f9501cad25a854564b3b734b3ea2883385863825b98dde4c012b7bb5301
+ sudo podman run -d -p 5000:5000 --name registry docker.io/registry:latest
Error: cannot listen on the TCP port: listen tcp4 :5000: bind: address already in use
Makefile:7: recipe for target 'configure_host' failed
make: *** [configure_host] Error 126

fmuyassarov:
@hardys You were right, the reason why it failed was that the during the initial test, container runtime was set to Podman while in the next test it was running with Docker.  As such, Docker was trying to run docker registry container on port 5000 while it was already in use by the previous podman registry container (or vise versa with container runtime).

Sorry for confusion, I'm closing this PR.

PR #128, opened 2019-11-11
Fix local run config for BMO+Ironic container

mikkosest:
Trigger for run local Ironic containers was missing

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -25,6 +25,7 @@ export GOPATH
 
 M3PATH="${GOPATH}/src/github.com/metal3-io"
 BMOPATH="${M3PATH}/baremetal-operator"
+RUN_LOCAL_IRONIC_SCRIPT="${BMOPATH}/tools/run_local_ironic.sh"
 CAPBMPATH="${M3PATH}/cluster-api-provider-baremetal"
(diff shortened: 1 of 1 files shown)


hardys:
By default ironic now runs on the minikube cluster ref https://github.com/metal3-io/metal3-dev-env/pull/68 so I'm not sure we want to add this, or if we do it should probably be configurable?

/cc @stbenjam

mikkosest:
> By default ironic now runs on the minikube cluster ref #68 so I'm not sure we want to add this, or if we do it should probably be configurable?
> 
> /cc @stbenjam

I agree that this is only for debugging and verification needs. The code change is under BMO_RUN_LOCAL flag, so it is executed only when user really wants to run local BMO+Ironic
setup. 
    if [ "${BMO_RUN_LOCAL}" = true ]; then
      touch bmo.out.log
      touch bmo.err.log
      kustomize build "$kustomize_overlay_path" | kubectl apply -f-
      kubectl scale deployment metal3-baremetal-operator -n metal3 --replicas=0
      ${RUN_LOCAL_IRONIC_SCRIPT}

hardys:
> The code change is under BMO_RUN_LOCAL flag, so it is executed only when user really wants to run local BMO+Ironic
setup.

Ah sorry I see that now, the conditional was hidden by the github UI folding the previous code, looks good then

PR #129, opened 2019-11-11
Generate node data based on flavors map

hardys:
Proposed follow-up to https://github.com/metal3-io/metal3-dev-env/pull/102

Instead of a hard-coded reference to the default flavor, we can iterate over the flavor and the number of nodes by looking at the dynamic variable num_nodes where "node" is the default flavor, but can also be any other name, or multiple flavors e.g master/worker where num_masters and num_workers can be used as before to control the host count for those flavors.

This means that e.g https://github.com/openshift-metal3/dev-scripts can continue to use master/worker flavors as before, by overriding the flavor vars like:

```
num_masters: 3
num_workers: 1
flavors:
  master:
    memory: 16384
    disk: 20
    vcpu: 4
    extradisks: false

  worker:
    memory: 8192
    disk: 20
    vcpu: 4
    extradisks: false
```

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -20,8 +20,7 @@ fi
 
 ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -e "working_dir=$WORKING_DIR" \
-    -e "num_masters=$NUM_MASTERS" \
-    -e "num_workers=$NUM_WORKERS" \
+    -e "num_nodes=$NUM_NODES" \
     -e "extradisks=$VM_EXTRADISKS" \
     -e "virthost=$HOSTNAME" \
     -e "platform=$NODES_PLATFORM" \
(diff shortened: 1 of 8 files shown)


hardys:
Hmm both CI jobs failed like:

```
./05_test.sh
Logging to ./logs/05_test-2019-11-12-073947.log
OK - centos-0 machine CR created
OK - centos-1 machine CR created
OK - centos-0 machine CR exist
   - Waiting for task completion (up to 2400 seconds)  - Command: 'check_bmh_association centos-0'
Unable to connect to the server: http2: server sent GOAWAY and closed the connection; LastStreamID=1, ErrCode=NO_ERROR, debug=""
```

So I suspect something went wrong with the deployment of the BMH nodes - @maelk or @jan-est can you access any more detailed logs as it's difficult to see exactly what may have failed from the console output?

maelk:
I would recommend to run the tests again after adding a `echo  "${TMP_RET}"` on line 241 of lib/common.sh, in the end of the until loop of the iterate function. That will show which check is failing when it times out.

hardys:
Ok the CI failure seems to be because of a hard-coded reference to master/worker in the 05 test script, retrying.

hardys:
Also triggered CI on https://github.com/openshift-metal3/dev-scripts/pull/843

hardys:
@jan-est @maelk I fixed the failing test and CI now passed, so WIP removed - openshift-metal3/dev-scripts#843 also passed so I think this is ready to land unless you see any remaining issues.

PR #139, opened 2019-11-26
Modify v1alpha2 CRs to add noCloudProvider option

maelk:
and remove patching of the node in cloud-init to set the
ProviderID


maelk:
no need for integration tests on this one, since v1alpha2 is not part of the tests yet

PR #142, opened 2019-12-03
Move hard-coded networks into vars file

hardys:
This data can then be overridden to test with custom networks
defined


hardys:
@celebdor I think this may help with testing multiple networks - we can probably make use of the `virtualport_type` here to setup the ovs bridges as well:

https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/libvirt/tasks/network_setup_tasks.yml#L15

maelk:
please rebase to get the workaround in

PR #143, opened 2019-12-04
Run Baremetal operator in a loop locally

maelk:
Baremetal operator exits when the cluster is unreachable, that
happens during the tests. Running it in a loop to restart it
if this happens

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -110,7 +110,7 @@ function launch_baremetal_operator() {
       kustomize build "$kustomize_overlay_path" | kubectl apply -f-
       kubectl scale deployment metal3-baremetal-operator -n metal3 --replicas=0
       ${RUN_LOCAL_IRONIC_SCRIPT}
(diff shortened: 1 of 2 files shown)


maelk:
/hold
trying to increase the minikube CPUs to see if it impacts the problem

03_launch_mgmt_cluster.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -xe
 
+CUR_DIR="$(dirname "$(readlink -f "${0}")")/"
+
```
hardys: There is already `$SCRIPTDIR` in lib/common.sh which can be used for this?

hardys:
This seems like a reasonable temporary workaround to unblock CI, but it'd be good to figure out why the minikube VM becomes unresponsive - one nit noted but otherwise lgtm as a workaround.

maelk:
/hold cancel
Increasing the cpus to 4 do not help with the issue, we'll go with this workaround

PR #144, opened 2019-12-04
increase number of cpus of Minikube to 4

maelk:


maelk:
not using this workaround, going with #143

PR #146, opened 2019-12-09
Add missing ovs role

bcrochet:
With the ability to pull the networks out into a separate config file,
comes the ability to specify an openvswitch network. A necessary role
was missing. This pulls in that missing role. It installs the necessary
ovs packages, and enables the service.


hardys:
Looks good @bcrochet thanks - I don't recall why we had the parts/ovs path in tripleo-quickstart, should we just squash the path so it's roles/ovs or roles/openvswitch perhaps?

PR #153, opened 2019-12-12
Change V1ALPHA2_SWITCH variable to CAPI_VERSION

maelk:
in order to support v1alpha3

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -99,8 +99,13 @@ export BAREMETAL_OPERATOR_IMAGE=${BAREMETAL_OPERATOR_IMAGE:-"quay.io/metal3-io/b
 # Config for OpenStack CLI
 export OPENSTACK_CONFIG=$HOME/.config/openstack/clouds.yaml
 
-# v1alpha2 var
+# CAPI version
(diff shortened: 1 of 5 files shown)


lib/common.sh
```diff
@@ -99,8 +99,13 @@ export BAREMETAL_OPERATOR_IMAGE=${BAREMETAL_OPERATOR_IMAGE:-"quay.io/metal3-io/b
...
 export OPENSTACK_CONFIG=$HOME/.config/openstack/clouds.yaml
 
-# v1alpha2 var
+# CAPI version
 export V1ALPHA2_SWITCH=${V1ALPHA2_SWITCH:-"false"}
+if [ "${V1ALPHA2_SWITCH}" == true ]; then
+  export CAPI_VERSION=${CAPI_VERSION:-"v1alpha2"}
+else
+  export CAPI_VERSION=${CAPI_VERSION:-"v1alpha2"}
+fi
```
fmuyassarov: Isn't it setting the same value for CAPI_VERSION no matter if condition met or not?

maelk: good catch, fixed!

PR #157, opened 2019-12-17
WIP testing with CentOS8

hardys:
This aims to remove the RDO repo dependency for both RHEL8 and CentOS8 - the packages we need from there have reduced since e.g vbmc is now containerized and the qemu-kvm changes needed are in the main el8 repos now.

The main TODO item is how we handle install of openstackclient but for now this is handled via pip, similar to what is already done on Ubuntu.

Marking this WIP as I've only tested on RHEL8 so far, @yrobla is planning to try with CentOS and provide feedback on any additional fixes needed.

### DIFF
--- a/centos_install_requirements.sh
+++ b/centos_install_requirements.sh
@@ -20,28 +20,26 @@ sudo yum -y update
 source /etc/os-release
 # VERSION_ID can be "7" or "8.x" so strip the minor version
 DISTRO="${ID}${VERSION_ID%.*}"
-if [ ! -f /etc/yum.repos.d/epel.repo ] ; then
-    if [[ $DISTRO == "rhel7" ]]; then
-        sudo yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
-    elif [[ $DISTRO == "centos7" ]]; then
-        sudo yum -y install epel-release --enablerepo=extras
-    fi
-fi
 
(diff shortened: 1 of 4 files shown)


centos_install_requirements.sh
```diff
@@ -20,39 +20,34 @@ sudo yum -y update
...
-fi
 
 if [[ $DISTRO == "centos7" ]]; then
-    sudo yum -y install epel-release dnf --enablerepo=extras
+  sudo yum -y install epel-release dnf --enablerepo=extras
+  # Install tripleo-repos, used to get a more recent version of some packages on CentOS7
+  sudo dnf -y --repofrompath="current-tripleo,https://trunk.rdoproject.org/${DISTRO}-master/current-tripleo" install "python*-tripleo-repos" --nogpgcheck
+  sudo tripleo-repos current-tripleo
+  # There are some packages which are newer in the tripleo repos
+  sudo yum -y update
```
elfosardo: let's use dnf here

elfosardo: let's use dnf here

hardys:
> are we missing the centos8 part in vm-setup/roles/packages_installation/tasks/main.yml ?

Yes good point, I thought @yrobla already pushed a fix for that but it looks like we are missing that, I'll rebase and update, thanks!

hardys:
Replaced by https://github.com/metal3-io/metal3-dev-env/pull/223

PR #160, opened 2019-12-20
Enable testing with an IPv6 EXTERNAL_SUBNET

hardys:
This allows initial testing with an external/baremetal network
configured for ipv6

### DIFF
--- a/vm-setup/roles/libvirt/templates/network.xml.j2
+++ b/vm-setup/roles/libvirt/templates/network.xml.j2
@@ -1,7 +1,14 @@
 {% set nat_port_range = item.nat_port_range|default([1024, 65535]) %}
 {% set netmask = item.netmask|default("") %}
 {% set prefix = item.prefix|default("") %}
+{% if item.dns.options is defined %}
+<network xmlns:dnsmasq='http://libvirt.org/schemas/network/dnsmasq/1.0'>
+  <dnsmasq:options>
+    <dnsmasq:option value='{{ item.dns.options }}'/>
+  </dnsmasq:options>
+{% else %}
 <network>
+{% endif %}
   <name>{{ item.name }}</name>
   <bridge name='{{ item.bridge }}'/>
 {% if item.forward_mode is defined %}
--- a/vm-setup/roles/libvirt/templates/network.xml.j2
+++ b/vm-setup/roles/libvirt/templates/network.xml.j2
@@ -1,5 +1,6 @@
 {% set nat_port_range = item.nat_port_range|default([1024, 65535]) %}
-{% set netmask = item.netmask|default('255.255.255.0') %}
+{% set netmask = item.netmask|default("") %}
+{% set prefix = item.prefix|default("") %}
 <network>
   <name>{{ item.name }}</name>
   <bridge name='{{ item.bridge }}'/>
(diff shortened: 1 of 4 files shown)


vm-setup/roles/libvirt/templates/network.xml.j2
```diff
@@ -25,7 +30,11 @@
 {% set numflavor = lookup('vars', 'num_' + flavor + 's')|default(0)|int %}
 {% for num in range(0, numflavor) %}
 {% set ironic_name = ironic_prefix + flavor + "_" + num|string %}
+{% if item.address|ipv6 != False %}
+      <host id='01:{{ node_mac_map.get(ironic_name).get(item.name)}}' name='{{flavor}}-{{num}}' ip='{{item.dhcp_range[0]|ipmath(ns.index|int)}}'/>
```
derekhiggins: This is where I started hitting problems with DHCPv6, I couldn't see a way for libvirt to do IP based on MAC addresses. The ID its looking for is the ClientID/DUID , you we don't know them at this point..... I ended up with a script to change the dnsmasq config to insert back in the MAC but this was just a POC not a workable solution.

hardys: Interesting, thanks @derekhiggins - I copied the DUID format from the virsh net-dhcp-leases in the ipv4 setup, which is just the mac with this prefix.

hardys: So digging into this it seems dnsmasq creates the DUID derived from the MAC:

http://thekelleys.org.uk/gitweb/?p=dnsmasq.git;a=blob;f=src/dhcp6.c;h=d7b4e8bc489df137d58cf8365d38f19960e22206;hb=HEAD#l615

So I think we should be able to use that behavior to predict the DUID in the libvirt conf - @derekhiggins is that something you already tried?

derekhiggins: I might be wrong but AIUI the client DUID is generated by the dhcp client and dnsmasq doesn't have any say in its value. so I'm guessing that the link above is how dnsmasq generates a server DUID.

hardys: @derekhiggins ah yeah of course that makes sense, I think we reach a similar conclusion though, looking at the dhclient conf on my test box we see:

```
# DHCP Client Configuration file.
#   see /usr/share/doc/dhclient/dhclient.conf.example
#   see dhclient.conf(5) man page
#
# Send client identifier as "hardware-type.link-layer address" (e.g. "1.c2.23.7d.c3.52.2c")
# Required in environments where a bridge might be clobbering the forwarded
# packet's MAC address (common in Wifi, Docsis, or ADSL bridging scenarios)
# see dhcp-options(5) man page for 'dhcp-client-identifier'
# see dhcp-eval(5) man page for 'hardware'
send dhcp-client-identifier = hardware;
```

AFAICT that means the client ID (DUID) will be 01 (ethernet network type) followed by the MAC address, so we can use that behavior to predict the DUID provided the client is configured like this (and from observation of the virsh net-dhcp-leases it appears to be)

russellb:
I started working on building on top of this to support a dual-stack (v4 + v6) baremetal network.  So far I haen't tried to make the IPv6 part optional.  definitely still WIP ... https://github.com/russellb/metal3-dev-env/commits/ipv6_ctlplane

vm-setup/roles/libvirt/templates/network.xml.j2
hardys: Ok to resolve this thread we discovered you can tell NetworkManager to use a duid-ll as mentioned in a subsequent commit on this branch

hardys:
@russellb @derekhiggins @stbenjam I think this is ready for review now, several of us have tested this with ipv6 and I've also re-tested with ipv4 and all seems to work OK.

PR #163, opened 2019-12-23
Fix Networking in Ubuntu ControlPlane CR

kashifest:
This PR fixes the networking interface issue in ubuntu controlplane and worker CR so that the networking survive while the node reboots. Previously, networking did not survive reboot

### DIFF
--- a/crs/v1alpha2/controlplane_ubuntu.yaml
+++ b/crs/v1alpha2/controlplane_ubuntu.yaml
@@ -81,16 +81,17 @@ spec:
                   192.168.111.249
               }
           }
-      - path: /etc/network/interfaces
+      - path : /etc/netplan/50-cloud-init.yaml
         owner: root:root
         permissions: '0644'
-        content: |
-          # interfaces(5) file used by ifup(8) and ifdown(8)
-          auto lo
-          iface lo inet loopback
-          ## To configure a dynamic IP address
-          auto enp2s0
-          iface enp2s0 inet dhcp
+        content: | 
+          network:
+              ethernets:
+                  enp1s0:
+                      dhcp4: true
+                  enp2s0:
+                      dhcp4: true
+              version: 2
       - path: /tmp/akeys
         owner: root:root
         permissions: '0644'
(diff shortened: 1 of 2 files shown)


crs/v1alpha2/controlplane_ubuntu.yaml
```diff
@@ -81,16 +81,27 @@ spec:
                   192.168.111.249
               }
           }
-      - path: /etc/network/interfaces
+          vrrp_instance VI_2 {
```
maelk: Why is there a second vrrp config?

maelk:
No need to run the integration tests, those CR files for v1a2 are not tested

crs/v1alpha2/controlplane_ubuntu.yaml
kashifest: This ensures the Ironic Endpoint IP 172.22.0.2 is also alive in target-cluster controlplane. This is added in 03.. scripts in minikube and this IP needs to be also on the controlplane so that after pivoting the ironic endpoint is reachable.

maelk: This is problematic because as long as you have Minikube cluster still up and running, you will have ip address conflict

maelk:
The failure is probably linked to the duplicated ironic ip address in the network. Which cluster controls the node deployment? Minikube ? if so, there should not be the ironic ip on the target control node since it conflicts. 
Another issue will be when you have multiple masters. How do you make sure that the ironic pod is on the node that has the ironic ip address at the moment ? you would need multiple ironic pods, one per master, with a daemon set maybe ?

kashifest:
> The failure is probably linked to the duplicated ironic ip address in the network. Which cluster controls the node deployment? Minikube ? if so, there should not be the ironic ip on the target control node since it conflicts.
> Another issue will be when you have multiple masters. How do you make sure that the ironic pod is on the node that has the ironic ip address at the moment ? you would need multiple ironic pods, one per master, with a daemon set maybe ?

@maelk you were right ... the failure was indeed because of  duplicated ironic ip address ... I have removed the second VRRP now and the PR should be ok now. Please review again

PR #166, opened 2019-12-30
Fix host_cleanup.sh to set num_nodes

dhawal55:
Set num_nodes to $NUM_NODES so host_cleanup.sh cleans up all VMs


hardys:
Thanks, this lgtm, it was missed in #102 which got merged via #129

PR #167, opened 2020-01-07
[WIP] test Centos 8 images

elfosardo:
Upstream images in metal3-io are being converted to use Centos 8
as base image.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -60,7 +60,7 @@ for IMAGE_VAR in IRONIC_IMAGE IPA_DOWNLOADER_IMAGE VBMC_IMAGE SUSHY_TOOLS_IMAGE 
     sudo "${CONTAINER_RUNTIME}" pull "$IMAGE"
 done
 
-# Download IPA and CentOS 7 Images
+# Download IPA and CentOS 8 Images
 mkdir -p "$IRONIC_IMAGE_DIR"
 pushd "$IRONIC_IMAGE_DIR"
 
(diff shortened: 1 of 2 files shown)


01_prepare_host.sh
```diff
@@ -60,7 +60,7 @@ for IMAGE_VAR in IRONIC_IMAGE IPA_DOWNLOADER_IMAGE VBMC_IMAGE SUSHY_TOOLS_IMAGE
     sudo "${CONTAINER_RUNTIME}" pull "$IMAGE"
 done
 
-# Download IPA and CentOS 7 Images
+# Download IPA and CentOS 8 Images
```
hardys: I don't think we care about testing deployment of centos8 in this PR, instead we need to set e.g IRONIC_LOCAL_IMAGE to point to your fork:

https://github.com/metal3-io/metal3-dev-env/blob/master/02_configure_host.sh#L135

maelk:
The CI is broken right now, we're working on it. Something with the ironic-image change, maybe ? So please don't run the /test-*integration

elfosardo:
@maelk  which change in ironic-image ?

hardys:
@elfosardo seems to be an issue from https://github.com/metal3-io/ironic-image/pull/120 which should be fixed via  https://github.com/metal3-io/ironic-image/pull/123

maelk:
Unfortunately, that was not the cause of the problem :/ it was an issue, but not the one breaking the CI. We're still investigating.

elfosardo:
I believe the CI has been fixed, can we test this now?

elfosardo:
actually the change for the centos8 base image in ironic-image has been merged already, so test on the field!
But we might use this to test this other change https://github.com/metal3-io/ironic-image/pull/119

elfosardo:
All the Centos8 migration PRs have been merged

PR #173, opened 2020-01-10
Change dockerfile location depending on repo

maelk:
when building the local images
/cc @fmuyassarov 
/cc @stbenjam


maelk:
This is a quick fix for the CI to run again for baremetal-operator. It does not impact if the image is not built locally so we will push it in and test it from BMO

maelk:
Testing : https://github.com/metal3-io/baremetal-operator/pull/386

PR #182, opened 2020-01-16
Add a branch checkout in the image building step

maelk:
To build the local images from a branch if needed when cloned

For example :
IRONIC_LOCAL_IMAGE=https://github.com/metal3-io/ironic-image.git
IRONIC_LOCAL_IMAGE_BRANCH=master
to build the image from the master branch of the ironic repo (defaults to master)

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -133,7 +133,9 @@ fi
 
 # Support for building local images
 for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
-  IMAGE=${!IMAGE_VAR}
+  BRANCH_IMAGE_VAR="${IMAGE_VAR}_BRANCH"
+  IMAGE="${!IMAGE_VAR}"
+  BRANCH="${!BRANCH_IMAGE_VAR:-master}"
 
   case ${IMAGE_VAR%_LOCAL_IMAGE} in
     'BAREMETAL_OPERATOR')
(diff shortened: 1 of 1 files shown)


maelk:
This PR does not consider the cases when the repo is cloned by is not on the correct branch to follow what was done before, if the folder exists, we directly use it.

02_configure_host.sh
```diff
@@ -148,8 +150,14 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
   if [[ "$IMAGE" =~ "://" ]] ; then
     REPOPATH=~/${IMAGE##*/}
     # Clone to ~ if not there already
-    [ -e "$REPOPATH" ] || git clone "$IMAGE" "$REPOPATH"
-    cd "$REPOPATH" || exit
+    if [ -e "$REPOPATH" ]; then
+       cd "$REPOPATH" || exit
```
kanwar-saad: please one kind of variable name either "$VAR" or $VAR or ${VAR} or "${VAR}". It is different everywhere now. I prefer "${VAR}"

02_configure_host.sh
```diff
@@ -148,8 +150,14 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
   if [[ "$IMAGE" =~ "://" ]] ; then
     REPOPATH=~/${IMAGE##*/}
     # Clone to ~ if not there already
-    [ -e "$REPOPATH" ] || git clone "$IMAGE" "$REPOPATH"
-    cd "$REPOPATH" || exit
+    if [ -e "$REPOPATH" ]; then
```
kanwar-saad: it can be like this

if dir not exist
  then clone
end if

cd dir
switch branch

PR #187, opened 2020-01-20
Install tripleo current repos for Centos 8

elfosardo:
The correct path for Centos 8 TripleO repos is current and not
current-tripleo.

### DIFF
--- a/centos_install_requirements.sh
+++ b/centos_install_requirements.sh
@@ -47,8 +47,13 @@ sudo yum -y install \
   wget
 
 # Install tripleo-repos, used to get a recent version of python-virtualbmc
-sudo dnf -y --repofrompath="current-tripleo,https://trunk.rdoproject.org/${DISTRO}-master/current-tripleo" install "python*-tripleo-repos" --nogpgcheck
-sudo tripleo-repos current-tripleo
+if [[ $DISTRO == "centos8" ]]; then
+  sudo dnf -y --repofrompath="current,https://trunk.rdoproject.org/${DISTRO}-master/current" install "python*-tripleo-repos" --nogpgcheck
+  sudo tripleo-repos current
+else
+  sudo yum -y --repofrompath="current-tripleo,https://trunk.rdoproject.org/${DISTRO}-master/current-tripleo" install "python*-tripleo-repos" --nogpgcheck
+  sudo tripleo-repos current-tripleo
+fi
 
 # There are some packages which are newer in the tripleo repos
 sudo yum -y update


hardys:
Thanks @elfosardo this looks OK but I took a different approach in https://github.com/metal3-io/metal3-dev-env/pull/157

Since we now have the vbmc running in a container, the only thing we actually install from RDO AFAICS is ironicclient/openstackclient, so I'm wondering if we should just stop enabling the RDO repos (which is what I did in my PR, instead pip installing the client).

Another approach could be to have an ironic-client container which just contains the client, and this would also be useful for debugging on a cluster where the client may not be installed anywhere?

I actually proposed https://github.com/openshift/ironic-image/pull/17 a while back to add the client to ironic-image, so one option would be to do that in the metal3 image, but perhaps a new smaller image would be preferable?

@maelk any thoughts on the preferred option here, and what would work best for other platforms?

elfosardo:
@hardys I'm aware of your PR to migrate to Centos8, this was mainly intended as temporary fix for an issue raised in #180
I'm ok with adding a small image with the client, it'll be useful for testing/debugging

maelk:
@hardys When it comes to the client, I think the easiest would be to install it with pip in metal3-dev-env. It's the most straightforward from a user perspective. That being said, I really like the idea of having a container image that we could use to debug in certain situations (like on a management cluster that is not in metal3-dev-env)

hardys:
@maelk yes agree pip is the easiest, and that's what I did in #157 - some folks have objected to using pip system-wide in the past, but it seems reasonable in this case (we could look at using pip --user in future).

If we're OK with that approach I'll rebase/re-test  #157 and we then potentially won't need the RDO repos anymore.  If this PR is needed as an immediate fix I'm fine with it though.

elfosardo:
@hardys  I would probably be one of those folks, I never recommend to run pip system wide, better create a virtualenv or restrict to user.
Let's rebase #157 and see how it goes, I can find some time to test it.
This one is not needed immediately.

hardys:
@elfosardo when you get time please see https://github.com/metal3-io/metal3-dev-env/pull/223 and https://github.com/metal3-io/metal3-dev-env/pull/215 which I'm proposing as an alternative - with that approach we remove the RDO repo requirement completely for CentOS8

hardys:
@elfosardo should we close this one and go with https://github.com/metal3-io/metal3-dev-env/pull/223 ?

elfosardo:
obsoleted by #223

PR #196, opened 2020-01-24
fix ipaddress import failure

Jaakko-Os:
due to missing import, ipcalc was also failing


maelk:
What about CentOS ? Would the same problem happen ?

Jaakko-Os:
> What about CentOS ? Would the same problem happen ?

seems to be fine in centos since the integration tests pass

maelk:
the integration tests are also passing for Ubuntu, even if ipcalc is not installed properly. Did you verify if ipcalc is installed on centOS properly ?

Jaakko-Os:
No. Closing this PR.

PR #197, opened 2020-01-24
Add v1alpha3 CAPBM version

maelk:
and change v1alpha2 CAPBM to release branch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,9 @@ test:
 test_v1a2:
 	./05_test_v1a2.sh
 
+test_v1a3:
+	./05_test_v1a2.sh
+
 lint:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -26,6 +26,9 @@ test:
 test_v1a2:
 	./05_test_v1a2.sh
 
+test_v1a3:
+	./05_test_v1a2.sh
```
Jaakko-Os: Script name should not be version specific if used like this.

kashifest: Is it a typo or are we using v1a2 test scripts for v1a3 also?

maelk: This is just to keep the CI jobs green. Once the v1a3 target exists, we'll change it

PR #201, opened 2020-01-27
Default CAPI_VERSION set to v1alpha2

### LINKED ISSUE #200: Default CAPI_VERSION is set to `v1alpha1`
https://github.com/metal3-io/metal3-dev-env/blob/master/lib/common.sh#L88
https://github.com/metal3-io/metal3-dev-env/blob/master/config_example.sh#L87

According to https://metal3.io/try-it.html#setup the default value should be `v1alpha2`

e-minguez:
Fix #200 

TBH, I'm not sure if you want to switch to v1alpha2 as the default API version, but just in case and in order to be 'compilant' with the documentation.

### DIFF
--- a/config_example.sh
+++ b/config_example.sh
@@ -82,9 +82,9 @@
 
 # Select the Cluster API version
 # Accepted values : v1alpha1 v1alpha2
-# default: v1alpha1
+# default: v1alpha2
 #
-#export CAPI_VERSION=v1alpha1
+#export CAPI_VERSION=v1alpha2
 
 # Configure provisioning network for single-stack ipv6
 #PROVISIONING_IPV6=false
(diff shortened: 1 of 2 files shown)


russellb:
I think making the default the newest version that works makes sense as the default.  @maelk

maelk:
I am not sure about this @russellb. I would rather update the docs for now to reflect the current configuration, and then once v1alpha3 is released, we can review this decision again. v1alpha1 is the default interface that is used downstream, in Openshift for example.

russellb:
> I am not sure about this @russellb. I would rather update the docs for now to reflect the current configuration, and then once v1alpha3 is released, we can review this decision again. v1alpha1 is the default interface that is used downstream, in Openshift for example.

OK, I'm fine with waiting for v1alpha3.

maelk:
The default version was changed to v1alpha3

PR #202, opened 2020-01-27
Fix when kubectl already installed and it is not latest

alosadagrande:
Signed-off-by: Alberto Losada <alosadag@redhat.com>

This PR fixes issue #98 

I faced this issue when run the metal3-dev-env scripts in my CentOS 7 when kubernetes-client rpm package (v1.5.2) was already installed.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -26,10 +26,15 @@ if ! command -v docker-machine-driver-kvm2 2>/dev/null ; then
     sudo mv docker-machine-driver-kvm2 /usr/local/bin/.
 fi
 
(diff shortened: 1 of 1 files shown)


alosadagrande:
Just a clarification:

KUBECTL_PATH variable was created since kubectl when installed from rpm packages on CentOS 7 it is placed at /usr/bin/kubectl. So, if there is a kubectl installation, I want to know where it is located so I can overwrite the binary. Otherwise I'll have two kubectl binaries in the same server. Probably kubectl version  won't be updated until new shell is launched again,e .g. won't be updated during the script execution.

PR #204, opened 2020-01-28
Adds calico as SDN to the new cluster

e-minguez:
The default kubeadmconfigs deploy everything but the SDN which means that the cluster is not fully functional and for a proper cluster you need to:

* ssh into the master host and add an SDN
or
* get the kubeconfig (by ssh'ing into the master host) to get the kubeconfig, then add an SDN

The calico SDN manifest (https://docs.projectcalico.org/v3.11/manifests/calico.yaml) almost fits the IPv4 range (`CALICO_IPV4POOL_CID=192.168.0.0/16`) but I believe this is ignored because it is created via the `--cluster-cidr` flag here https://github.com/metal3-io/metal3-dev-env/blob/master/crs/v1alpha2/cluster.yaml#L11

### DIFF
--- a/crs/v1alpha2/controlplane_centos.yaml
+++ b/crs/v1alpha2/controlplane_centos.yaml
@@ -54,6 +54,7 @@ spec:
     - mkdir -p /home/centos/.kube
     - cp /etc/kubernetes/admin.conf /home/centos/.kube/config
     - chown centos:centos /home/centos/.kube/config
+    - KUBECONFIG=/etc/kubernetes/admin.conf kubectl apply -f https://docs.projectcalico.org/v3.11/manifests/calico.yaml
   files:
     - path: /etc/keepalived/keepalived.conf
       content: |
--- a/crs/v1alpha2/controlplane_ubuntu.yaml
+++ b/crs/v1alpha2/controlplane_ubuntu.yaml
@@ -58,6 +58,7 @@ spec:
     - mkdir -p /home/ubuntu/.kube
     - cp /etc/kubernetes/admin.conf /home/ubuntu/.kube/config
     - chown ubuntu:ubuntu /home/ubuntu/.kube/config
+    - KUBECONFIG=/etc/kubernetes/admin.conf kubectl apply -f https://docs.projectcalico.org/v3.11/manifests/calico.yaml
   files:
       - path: /etc/keepalived/keepalived.conf
         content: |


russellb:
Can you please expand on the commit message to explain some of the reasoning behind this change?  Thanks.

russellb:
@maelk which SDN backend are you targeting for the clusters you create with cluster-api?

crs/v1alpha2/controlplane_centos.yaml
```diff
@@ -54,6 +54,7 @@ spec:
     - mkdir -p /home/centos/.kube
     - cp /etc/kubernetes/admin.conf /home/centos/.kube/config
     - chown centos:centos /home/centos/.kube/config
+    - KUBECONFIG=/etc/kubernetes/admin.conf kubectl apply -f https://docs.projectcalico.org/v3.11/manifests/calico.yaml
```
Xenwar: would this work if kubeadm init was run with a different pod network ?
It seems to assume the pod network as shown here https://docs.projectcalico.org/getting-started/kubernetes/quickstart.

Would it be better to download the file and do replacement ?

e-minguez: According to the manifest:
```
# The default IPv4 pool to create on startup if none exists. Pod IPs will be
# chosen from this range. Changing this value after installation will have
# no effect. This should fall within `--cluster-cidr`.
```
So I'm assuming as the CIDR is set here https://github.com/metal3-io/metal3-dev-env/blob/master/crs/v1alpha2/cluster.yaml that would be used.

maelk: I think this might cause a problem. It needs to fall within `--cluster-cidr`, but cluster CIDR is 192.168.0.0/18 and calico default pool is 192.168.0.0/16, so by default it does not fall within. So I don't think it would work as is. the sed step should be probably be applied

maelk:
@russellb This is up to the user. The SDN is deployed as part of the cluster add-ons, there is nothing forced by cluster-API

russellb:
OK, my personal preference for selfish reasons would be to use ovn-kubernetes :-)

or just leave this alone and document how people can customize their deployed clusters further.  For the purposes of the dev env, I'm not sure it's that important to start configuring addons.  At that point, we're past anything metal3 is involved in.

maelk:
agreed. we'll look into addons when we start working with clusterctl, but for now we should probably leave it out.

russellb:
Thanks @maelk 

/close

PR #207, opened 2020-01-31
Make code python 3 compatible

stannum-l:
Now python 2 is EOL, this patch set upgrades scripts to be python3
compatible.

Signed-off-by: Tin Lam <tin@irrational.io>

### DIFF
--- a/vm-setup/library/generate_macs.py
+++ b/vm-setup/library/generate_macs.py
@@ -55,7 +55,7 @@ def generate_baremetal_macs(nodes, networks):
                  random.randint(0x00, 0xff),
                  random.randint(0x00, 0xff),
                  random.randint(0x00, 0xff)]
-    base_mac = ':'.join(map(lambda x: "%02x" % x, base_nums))
+    base_mac = ':'.join(["%02x" % x for x in base_nums])
 
     start = random.randint(0x00, 0xff)
     if (start + (count * 2)) > 0xff:


vm-setup/library/generate_macs.py
```diff
@@ -55,7 +55,7 @@ def generate_baremetal_macs(nodes, networks):
                  random.randint(0x00, 0xff),
                  random.randint(0x00, 0xff),
                  random.randint(0x00, 0xff)]
-    base_mac = ':'.join(map(lambda x: "%02x" % x, base_nums))
+    base_mac = ':'.join(["%02x" % x for x in base_nums])
```
russellb: Is this required?

```
Python 3.7.6 (default, Dec 19 2019, 22:52:49) 
[GCC 9.2.1 20190827 (Red Hat 9.2.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import random
>>> base_nums = [0x00,
...                  random.randint(0x00, 0xff),
...                  random.randint(0x00, 0xff),
...                  random.randint(0x00, 0xff),
...                  random.randint(0x00, 0xff)]
>>> base_mac = ':'.join(map(lambda x: "%02x" % x, base_nums))
>>> print(base_mac)
00:62:67:bf:de
```

stannum-l: With the `':'.join()` this may just be a readability issue and not required, but from previous porting py 2 and py 3 effort, I noticed that `map(lambda ...` returns a `list` in python 2, but a `map` in python 3, which caused issues.

maelk:
@russellb any further comments on this ? Can we merge it ?

russellb:
fine with me
/approve
/lgtm

PR #213, opened 2020-02-10
v1alpha3 modifications for integration test

Jaakko-Os:
capbm and capi use 'capxxx-system' namespace instead of metal3 namespace.
    this helps getting over certificate issues when accessing
    capbm and capi services.
    
    Parameterization added to avoid file duplication in v1a2 and v1a3.


03_launch_mgmt_cluster.sh
```diff
@@ -177,9 +188,16 @@ function launch_cluster_api_provider_baremetal() {
...
-      [ "${CAPI_VERSION}" == "v1alpha3" ]; then
+    if [ "${CAPI_VERSION}" == "v1alpha3" ]; then
+      ./examples/generate.sh -f
+      kustomize_overlay_capbm "$kustomize_overlay_path" \
+        "$CAPBMPATH/examples/provider-components"
+      kubectl apply -f ./examples/_out/cert-manager.yaml
+      kubectl wait --for=condition=Available --timeout=300s -n cert-manager deployment cert-manager
+      kubectl wait --for=condition=Available --timeout=300s -n cert-manager deployment cert-manager-cainjector
+      kubectl wait --for=condition=Available --timeout=300s -n cert-manager deployment cert-manager-webhook
+      kubectl apply -f ./examples/_out/provider-components.yaml
```
maelk: this is duplicated with line 215. It should not be here, since the image modifications happen after

Jaakko-Os: removed

PR #214, opened 2020-02-10
Use python3 to build ubuntu based Metal3-dev-env

fmuyassarov:
Use Pyhon3 as Python2 EOL has passed.


ubuntu_install_requirements.sh
```diff
@@ -20,7 +20,6 @@ sudo apt -y update
 # ansible uses default python2 (python-pip) to run on the local machine
 sudo apt -y install \
   python3-pip \
-  python-pip \
   python-setuptools \
```
elfosardo: probably need to change this too

ubuntu_install_requirements.sh
```diff
@@ -20,7 +20,6 @@ sudo apt -y update
 # ansible uses default python2 (python-pip) to run on the local machine
```
elfosardo: maybe update this comment ?

ubuntu_install_requirements.sh
fmuyassarov: > probably need to change this too

Yes, thanks

ubuntu_install_requirements.sh
fmuyassarov: > maybe update this comment ?

sure

lib/network.sh
```diff
@@ -13,7 +13,7 @@ function network_address() {
   network=$2
   record=$3
 
-  result=$(python -c "import ipaddress; import itertools; print(next(itertools.islice(ipaddress.ip_network(u\"$network\").hosts(), $record - 1, None)))")
+  result=$(python3 -c "import ipaddress; import itertools; print(next(itertools.islice(ipaddress.ip_network(u\"$network\").hosts(), $record - 1, None)))")
```
hardys: We already set alternatives to python3 on CentOS/RHEL8 - could we perhaps add that for Ubuntu to avoid the need for changing these?

https://github.com/metal3-io/metal3-dev-env/blob/master/centos_install_requirements.sh#L38

hardys: Note #223 adds CentOS 8 support which we'll need to move to only python3

fmuyassarov: > We already set alternatives to python3 on CentOS/RHEL8 - could we perhaps add that for Ubuntu to avoid the need for changing these?
> 
> https://github.com/metal3-io/metal3-dev-env/blob/master/centos_install_requirements.sh#L38

Yes, absolutely

fmuyassarov:
@hardys , @elfosardo  I updated my commit, can you please have a look when you have time. 
So, now this patch is intending to switch to python3 only for Ubuntu based metal-dev-env.

maelk:
/approve
Let's wait for the move to centos 8 and then do the same for Centos

maelk:
This will need a rebase to run the tests properly due to the Centos change

fmuyassarov:
> You will need to fix the conflict and do a rebase to run the tests properly due to the Centos change

Rebased

vm-setup/roles/packages_installation/defaults/main.yml
```diff
@@ -47,32 +44,32 @@ packages:
          - OVMF
          - patch
          - psmisc
-         - python-pip
-         - python-requests
-         - python-setuptools
+         - python3-pip
+         - python3-requests
+         - python3-setuptools
```
hardys: These packages won't exist on centos7 so we shouldn't change them, we probably need to deprecate centos7 suppport or figure out a way for this to temporarily work with python2 as well as python3 (on Centos/RHEL8 and Ubuntu)

lib/network.sh
```diff
@@ -12,8 +12,15 @@ function network_address() {
   resultvar=$1
   network=$2
   record=$3
+  OS=$(awk -F= '/^ID=/ { print $2 }' /etc/os-release | tr -d '"')
+
+  if [ "${OS}" == "ubuntu" ]
+  then
+    result=$(python3 -c "import ipaddress; import itertools; print(next(itertools.islice(ipaddress.ip_network(u\"$network\").hosts(), $record - 1, None)))")
+  else
+    result=$(python -c "import ipaddress; import itertools; print(next(itertools.islice(ipaddress.ip_network(u\"$network\").hosts(), $record - 1, None)))")
```
hardys: Did you look at the alternatives approach I mentioned to avoid this conditional?

vm-setup/roles/packages_installation/defaults/main.yml
```diff
@@ -21,15 +21,11 @@ packages:
...
-    pip: # ansible uses these packages in python2 to run on the local machine
-     - libvirt-python==5.10.0
-     - lxml
-     - six
     pip3:
      - virtualbmc
      - lxml
      - netaddr
-     - libvirt-python
+     - libvirt-python==5.10.0
```
hardys: Is it worth adding a comment saying why this has to be pinned?

vm-setup/roles/packages_installation/defaults/main.yml
hardys: Oh I think this was addressed in the rebase commit, it may be worth squashing that into the original commit then force-pushing if you do any other updates

lib/network.sh
fmuyassarov: yes, done

vm-setup/roles/packages_installation/defaults/main.yml
fmuyassarov: When latest release of `libvirt-python` came out integration CI started to fail. Then, we decided to switch to `v5.10.0`  to avoid CI failure.  
I thought that we need the same in case of python3 as well, but apparently after my test, the latest versionof  `libvirt-python` works fine in dev-env.

fmuyassarov: So, switching back to the latest

PR #216, opened 2020-02-11
Ansible changes to have a single source of CRs

fmuyassarov:
Use  ` /vm-setup/roles/v1aX_integration_test/templates/`  folder for v1alphaX deployment tests rather than copy of those CRs from` /crs `and remove the `/crs` folder.


scripts/v1alphaX/provision_cluster.sh
```diff
@@ -1,17 +1,19 @@
...
 # shellcheck disable=SC1090
 source "${METAL3_DIR}/lib/images.sh"
 
-make_v1alphaX_machine controlplane "${CAPI_VERSION}" | kubectl delete -n metal3 -f -
-
+ANSIBLE_FORCE_COLOR=true ansible-playbook \
+   -e "metal3_dir=$SCRIPTDIR" \
+   -e "IMAGE_OS=$IMAGE_OS" \
+   -e "DEFAULT_HOSTS_MEMORY=$DEFAULT_HOSTS_MEMORY" \
+   -e "CAPI_VERSION=$CAPI_VERSION" \
```
maelk: Isn't the variables in the playbook defaulting to the environment variable content ?

PR #61, opened 2019-09-25
Add a switch to use redfish as BMC driver

kashifest:
Add a switch for using redfish bmc driver

PR #218, opened 2020-02-12
Generate a correct ironic_nodes.json

dtantsur:
The current format is some awkward hybrid of ironic and BMO formats.
This changes brings it back to the ironic native format.

### DIFF
--- a/vm-setup/roles/libvirt/tasks/vm_setup_tasks.yml
+++ b/vm-setup/roles/libvirt/tasks/vm_setup_tasks.yml
@@ -133,8 +133,19 @@
 # task *must* occur after the above vm tasks, because if
 # `vm_nodes` is defined the template depends on the
 # `node_mac_map` variable.
-- name: Write ironic node json files
+- name: Write metal3 node json files
   template:
     src: ../templates/ironic_nodes.json.j2
     dest: "{{ nodes_file }}"
     force: no
+  vars:
+    nodes_json_format: metal3
+
+- name: Write ironic-compatible node json files
+  template:
+    src: ../templates/ironic_nodes.json.j2
+    dest: "{{ ironic_nodes_file }}/"
+    force: no
+  vars:
+    nodes_json_format: ironic
+  when: ironic_nodes_file
--- a/vm-setup/roles/libvirt/tasks/vm_teardown_tasks.yml
+++ b/vm-setup/roles/libvirt/tasks/vm_teardown_tasks.yml
@@ -93,7 +93,13 @@
         state: absent
       when: pool_check is success
 
-    - name: Remove ironic_nodes.json
+    - name: Remove {{ nodes_file }}
       file:
         path: "{{ nodes_file }}"
         state: absent
+
+    - name: Remove ironic native nodes file
+      file:
+        path: "{{ ironic_nodes_file }}"
+        state: absent
+      when: ironic_nodes_file
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -26,6 +26,7 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -e "manage_baremetal=$MANAGE_BR_BRIDGE" \
     -e "provisioning_url_host=$PROVISIONING_URL_HOST" \
     -e "nodes_file=$NODES_FILE" \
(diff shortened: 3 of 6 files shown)


dtantsur:
/hold

testing

dtantsur:
/unhold

seems to work

hardys:
Thanks for this - it would be useful for this to match the ironic format for testing with openstackclient.

I think we will need to modify the locations which read the BMC credentials though, e.g 

https://github.com/metal3-io/metal3-dev-env/blob/master/lib/common.sh#L206..L207

and 

https://github.com/openshift-metal3/dev-scripts/blob/master/utils.sh#L143..L144

I can see this was changed in https://github.com/metal3-io/metal3-dev-env/pull/61 cc @kashifest and I guess it was done like this to avoid exposing an interface change via NODES_FILE.

It should be fairly simple to prefix the username/password key with the driver though I think?

dtantsur:
If we use this file for something other than ironic, we should probably keep it as it is and/or create an actual ironic file.

hardys:
> If we use this file for something other than ironic, we should probably keep it as it is and/or create an actual ironic file.

Yeah perhaps that would be simpler, we could copy this template and generate another file, or make the template configurable

kashifest:
> Thanks for this - it would be useful for this to match the ironic format for testing with openstackclient.
> 
> I think we will need to modify the locations which read the BMC credentials though, e.g
> 
> https://github.com/metal3-io/metal3-dev-env/blob/master/lib/common.sh#L206..L207
> 
> and
> 
> https://github.com/openshift-metal3/dev-scripts/blob/master/utils.sh#L143..L144
> 
> I can see this was changed in #61 cc @kashifest and I guess it was done like this to avoid exposing an interface change via NODES_FILE.
> 
> It should be fairly simple to prefix the username/password key with the driver though I think?

@hardys  Yes, you are correct. But then, I only kept the original variables and not any driver specific variables to avoid any further changes. I also agree that we can make the template configurable.

dtantsur:
Should we rename the current file then? to something like `inventory.json`? I'd prefer to avoid keeping `ironic` in something that is not compatible with ironic.

hardys:
> Should we rename the current file then? to something like `inventory.json`? I'd prefer to avoid keeping `ironic` in something that is not compatible with ironic.

Maybe rename the current file to `metal3_nodes.json` and also write `ironic_nodes.json` for compatibility with ironic?

maelk:
This will need a rebase to run the tests properly due to the Centos change

vm-setup/roles/libvirt/tasks/vm_teardown_tasks.yml
```diff
@@ -93,7 +93,12 @@
         state: absent
       when: pool_check is success
 
-    - name: Remove ironic_nodes.json
+    - name: Remove metal3_nodes.json
```
hardys, line 96: The name should use the variable name e.g `Remove {{ nodes_file }}`

vm-setup/roles/libvirt/tasks/vm_setup_tasks.yml
```diff
@@ -133,8 +133,14 @@
...
   template:
-    src: ../templates/ironic_nodes.json.j2
+    src: ../templates/metal3_nodes.json.j2
     dest: "{{ nodes_file }}"
     force: no
+
+- name: Write ironic-compatible node json files
+  template:
+    src: ../templates/ironic_nodes.json.j2
+    dest: "{{ working_dir }}/ironic_nodes.json"
```
hardys, line 145: We should make this a variable or this will overwrite nodes_file in the case where it's set to ironic_nodes.json and also dev-scripts writes out the existing file to a per-cluster directory, so we'll want to do the same for this file.

hardys:
A couple of small nits but otherwise lgtm as an interim solution - eventually it would be nice to avoid the duplication of templates but I don't see an easy way to do that without breaking the changes introduced in #61

dtantsur:
hey @hardys, could you take another look?

dtantsur:
I assume that nobody cares about this change.

maelk:
what do we need the proper format for ? how is it used in metal3-dev-env ?

dtantsur:
It's used if you need to test ironic separately from k8s (I did have this case for a lab back when I filed this).

Issue #224, opened 2020-02-15
PROVISIONING_NETMASK is not calculated 


Arvinderpal:
ipcalc has no flag named `--netmask`, at least for the version running on Ubuntu 18.04.1 LTS. This results in PROVISIONING_NETMASK not being calculated properly. 

```
+++ ipcalc --netmask 172.22.0.0/24
+++ cut -d= -f2
++ export 'PROVISIONING_NETMASK=Unknown option: --netmask'
++ PROVISIONING_NETMASK='Unknown option: --netmask'
```
Alternative is to just write a simple bash function to do this calculation. I'll send a PR for that. 

OS: Ubuntu 18.04.1 LTS

PR #225, opened 2020-02-15
Replace ipcalc with python script

Arvinderpal:
Fixes: https://github.com/metal3-io/metal3-dev-env/issues/224

Output of `./01_prepare_host.sh`:
```
+++ python -c 'import ipaddress; print(ipaddress.ip_network(u"172.22.0.0/24").netmask)'                                                                                                                            
++ export PROVISIONING_NETMASK=255.255.255.0                                                                                                                                                                       
++ PROVISIONING_NETMASK=255.255.255.0 
```

### DIFF
--- a/lib/network.sh
+++ b/lib/network.sh
@@ -39,7 +39,7 @@ fi
 
 # shellcheck disable=SC2155
 export PROVISIONING_CIDR=$(python -c "import ipaddress; print(ipaddress.ip_network(u\"$PROVISIONING_NETWORK\").prefixlen)")
-export PROVISIONING_NETMASK=${PROVISIONING_NETMASK:-$(ipcalc --netmask "$PROVISIONING_NETWORK" | cut -d= -f2)}
+export PROVISIONING_NETMASK=${PROVISIONING_NETMASK:-$(python -c "import ipaddress; print(ipaddress.ip_network(u\"$PROVISIONING_NETWORK\").netmask)")}
 
 network_address PROVISIONING_IP "$PROVISIONING_NETWORK" 1
 network_address CLUSTER_PROVISIONING_IP "$PROVISIONING_NETWORK" 2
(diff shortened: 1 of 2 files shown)


lib/network.sh
```diff
@@ -18,6 +18,15 @@ function network_address() {
   export resultvar
 }
 
+# REF: https://forum.archive.openwrt.org/viewtopic.php?id=47986&p=1#p220781
+function cdr2mask ()
+{
+   # Number of args to shift, 255..255, first non-255 byte, zeroes
+   set -- $(( 5 - ($1 / 8) )) 255 255 255 255 $(( (255 << (8 - ($1 % 8))) & 255 )) 0 0 0
```
hardys: This won't work with ipv6 - I think a better approach would be to use python, e.g similar to:

https://github.com/metal3-io/metal3-dev-env/blob/master/lib/network.sh#L41

hardys:
Thanks for the PR - one suggestion which should make this compatible with ipv6

lib/network.sh
Arvinderpal: Ohh. Didn't realize that. Changed it to python. PTAL

Arvinderpal:
I should also note that this does not break anything at the moment since the underlying ubuntu specific code does not make use of `PROVISIONING_NETMASK`. I was walking through the outputs of the metal3-dev-env and happened to see this issue.

hardys:
lgtm subject to CI passing, thanks!

maelk:
This makes the bmo deployment fail

hardys:
> This makes the bmo deployment fail

I don't see many clues in the logs, any idea why?  AFAICS this does generate the same output as the previous ipcalc:

```
$ export PROVISIONING_NETWORK="172.22.0.0/24"
$ ipcalc --netmask "$PROVISIONING_NETWORK" | cut -d= -f2
255.255.255.0
$ export PROVISIONING_NETMASK=${PROVISIONING_NETMASK:-$(python -c "import ipaddress; print(ipaddress.ip_network(u\"$PROVISIONING_NETWORK\").netmask)")}
$ echo $PROVISIONING_NETMASK 
255.255.255.0
```

Not tested this e2e locally though.

maelk:
Due to CentOS changes, this will need a rebase

PR #228, opened 2020-02-18
Include tasks to download centos-updated.qcow2 & some Ansible vars modifications

alosadagrande:
Signed-off-by: Alberto Losada <alosadag@redhat.com>

There are two main changes:

1. Ansible variable now includes a lookup for the IMAGE_NAME_CENTOS env var since an specific image with a proper cloud-init version needs to be used as OS for the provisioning process of the new cluster.

`IMAGE_NAME_CENTOS: "{{ lookup('env', 'IMAGE_NAME_CENTOS') | default('centos-updated.qcow2', true)}}"`

I think, the default image for CentOS should be the centos-updated, since it is the only that works when provisioning new clusters.

2. A few tasks have been added in order to verify that the centos-updated.qcow2 image and its md5 hash are available for the new bmh.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/provision.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/provision.yml
@@ -20,6 +20,26 @@
 
   - name: Template CentOS machine crs
     block:
+      - name: Verify specific centos image containing newer version of cloud-init is downloaded
+        get_url:
+          url: "{{ IMAGE_LOCATION_CENTOS }}"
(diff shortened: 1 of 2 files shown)


fmuyassarov:
@alosadagrande 
The CI is failing because the CityCloud which is a cloud provider is upgrading an Openstack release as Mael mentioned on the slack.
Hopefully, CityCloud infra will become stable soon and we'll be able to run the tests smoothly.

PR #229, opened 2020-02-19
Ansible changes to separate provisioning scripts

fmuyassarov:
This changes allow users to separately provision/deprovision cluster, controlplane & worker machines.
Currently, `provision.yml` provisions cluster, controlplane & worker at once, while there are some cases where users want to do it step by step or to deploy a single node cluster.


scripts/v1alphaX/deprovision_worker.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+METAL3_DIR="$(dirname "$(readlink -f "${0}")")/../.."
+RUN_CI_TEST=false
+PROVISION_CLUSTER=false
+PROVISION_CONTROLPLANE=false
+PROVISION_WORKER=false
+DEPROVISION_CONTROLPLANE=false
+DEPROVISION_WORKER=true
+DEPROVISION_CLUSTER=false
```
hardys: Instead of setting shell variables to influence conditionals in one playbook, would it perhaps be cleaner to just have separate provision/deprovision playbooks? Similar to how we do the setup/teardown in vm-setup?

fmuyassarov: Originally we had only provision & deprovision & verify playbooks which were used by CI.
However, for the local provisioning we had other shell scripts under /scripts/v1alphaX/..
We thought that it would be sustainable to reuse the Ansible playbooks also for local provisioning/deprovisioning rather than having the shell scripts, and this is the currently how it is on the master.

But the problem of previous provision/deprovision playbooks is that they had some tasks which are not really necessary when provisioning locally, some of them are just verification tasks that are needed only when triggered by the CI. In that sense, I had to break down the provision & deprovision playbooks to decouple those tasks that would be called only when running by the CI but still be under the same role.

Apart from that, current /scripts/v1alphaX/provision_cluster.sh is calling the Ansible playbook that provisions a cluster and two machines at once, meaning that user can't really control the number of machines ha wants to provision. So, I thought by reusing a single role and adding some extra shell variables and by breaking down provision/deprovision playbooks as it is proposed here, we could reuse the same role/playbooks and allow users to provision/deprovision either a single node or two nodes cluster.

maelk: If I understood correctly, @hardys ' point was not about moving to ansible playbooks vs. shell, but to have multiple playbooks (one for provisioning, one for deprovisioning etc. instead of a single v1aX_integration_test.yml that covers all scenarios

hardys: Yes that's right @maelk it just seemed like that would be cleaner than a single playbook with the variables to control provision/deprovision behavior?

fmuyassarov: I agree @hardys @maelk . I just didn't understand you correctly.
Updated the PR, please review it.

PR #233, opened 2020-02-21
Do not install Docker on CentOS8

maelk:
Currently docker installation fails in CentOS8 with 
```
Error: 
 Problem: package docker-ce-3:19.03.6-3.el7.x86_64 requires containerd.io >= 1.2.2-3, but none of the providers can be installed
  - cannot install the best candidate for the job
  - package containerd.io-1.2.10-3.2.el7.x86_64 is excluded
  - package containerd.io-1.2.2-3.3.el7.x86_64 is excluded
  - package containerd.io-1.2.2-3.el7.x86_64 is excluded
  - package containerd.io-1.2.4-3.1.el7.x86_64 is excluded
  - package containerd.io-1.2.5-3.1.el7.x86_64 is excluded
  - package containerd.io-1.2.6-3.3.el7.x86_64 is excluded
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
make: *** [Makefile:4: install_requirements] Error 1
```
Since Docker is not supported in CentOS8, this PR adds a failure message instead of installing Docker in CentOS8 only

### DIFF
--- a/centos_install_requirements.sh
+++ b/centos_install_requirements.sh
@@ -59,20 +59,25 @@ fi
 if [[ "${CONTAINER_RUNTIME}" == "podman" ]]; then
   sudo yum -y install podman
 else
-  sudo yum install -y yum-utils \
-    device-mapper-persistent-data \
-    lvm2
-  sudo yum-config-manager \
-    --add-repo \
-    https://download.docker.com/linux/centos/docker-ce.repo
-    cat <<EOF > daemon.json
+  if [[ $DISTRO == "centos7" ]]; then
+    sudo yum install -y yum-utils \
+      device-mapper-persistent-data \
+      lvm2
+    sudo yum-config-manager \
+      --add-repo \
+      https://download.docker.com/linux/centos/docker-ce.repo
+      cat <<EOF > daemon.json
 {
   "insecure-registries" : ["192.168.111.1:5000"]
 }
 EOF
-  sudo chown root:root daemon.json
-  sudo yum install -y docker-ce docker-ce-cli containerd.io
-  sudo mkdir -p /etc/docker
-  sudo mv daemon.json /etc/docker/daemon.json
-  sudo systemctl restart docker
+    sudo chown root:root daemon.json
+    sudo yum install -y docker-ce docker-ce-cli containerd.io
+    sudo mkdir -p /etc/docker
+    sudo mv daemon.json /etc/docker/daemon.json
+    sudo systemctl restart docker
+  else
+    echo "Only Podman is supported in CentOS/RHEL8"
+    exit 1
+  fi
 fi
--- a/ip_range_check.py
+++ b/ip_range_check.py
@@ -2,8 +2,9 @@
 
 from ipaddress import IPv4Address
 import sys
-dhcp_range=unicode(sys.argv[2]).split(',')
-if IPv4Address(dhcp_range[0]) <= IPv4Address(unicode(sys.argv[1])) <= IPv4Address(dhcp_range[1]):
+import six
+dhcp_range=six.text_type(sys.argv[2]).split(',')
+if IPv4Address(dhcp_range[0]) <= IPv4Address(six.text_type(sys.argv[1])) <= IPv4Address(dhcp_range[1]):
   sys.exit(0)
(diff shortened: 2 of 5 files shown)


maelk:
This requires https://github.com/metal3-io/project-infra/pull/55

centos_install_requirements.sh
```diff
@@ -59,20 +59,25 @@ fi
...
-  sudo mkdir -p /etc/docker
-  sudo mv daemon.json /etc/docker/daemon.json
-  sudo systemctl restart docker
+    sudo chown root:root daemon.json
+    sudo yum install -y docker-ce docker-ce-cli containerd.io
+    sudo mkdir -p /etc/docker
+    sudo mv daemon.json /etc/docker/daemon.json
+    sudo systemctl restart docker
+  else
+    echo "Only Podman is supported in CentOS8"
```
hardys: Nit - We could say CentOS/RHEL8 as both are supported and will hit this path if someone overrides CONTAINER_RUNTIME

hardys:
lgtm, thanks, minor nit mentioned as this also applies when testing on RHEL.

vm-setup/roles/packages_installation/defaults/main.yml
```diff
@@ -87,3 +87,4 @@ packages:
          - python3-netaddr
          - virt-install
          - network-scripts
+         - firewalld
```
kashifest: why is this needed?

fmuyassarov: there is a [task](https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/firewall/tasks/firewalld.yaml) that involves firewalld service and in CentOS8 firewalld isn't preinstalled

kashifest: Aha. Ok. I guess it comes pre-installed with CentOS7

kashifest:
lgtm

vm-setup/roles/packages_installation/defaults/main.yml
hardys: When you do a local minimal install firewalld is pre-installed, but @maelk confirmed that in CI we use the cloud images, which don't seem to include it.

It seems that we also need to restart libvirtd in this case, so we may need to add a libvirtd restart task to ensure things are correctly configured when firewalld is installed/enabled after libvirtd has already started.

I tested on a CI VM and restarted both libvirtd then firewalld after a `make` failed with a timeout, but I think the important part was the libvirtd restart.

maelk:
/hold
testing

vm-setup/roles/packages_installation/defaults/main.yml
maelk: I added the restarts, let's see if this helps.

05_test.sh
```diff
@@ -127,7 +127,7 @@ check_bmh_status(){
 #
 # Usage ip_in_range ip start end
 ip_in_range() {
-  python ip_range_check.py "$1" "$2"
+  python3 ip_range_check.py "$1" "$2"
```
hardys: This should be transparent due to https://github.com/metal3-io/metal3-dev-env/blob/master/centos_install_requirements.sh#L35 - do we need something similar for ubuntu perhaps?

ip_range_check.py
```diff
@@ -1,9 +1,9 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
hardys: If we hard-code this, it won't work on a default CentOS7 install, should we use the alternatives approach mentioned above to avoid that, even if we expect to deprecate CentOS7 and python2 support soon?

05_test.sh
maelk: The problem is not for centos8. It was running with python3, and that's what we want. But for others, such as Ubuntu, and Centos7, we need to be specific as long as there is python2

ip_range_check.py
maelk: The code would anyways not work with python2 (needs unicode)

hardys:
lgtm when CI passes, thanks!

PR #242, opened 2020-03-04
Use RDO packages when installing virtualbmc

dtantsur:
Packages are safer since they pass CI testing, avoiding issues like
https://bugs.launchpad.net/pyghmi/+bug/1865987 that affects us now.
    
This change brings the virtualbmc image in sync with how ironic
images are built and also potentially reduces its size because
the development packages are no longer installed.

### DIFF
--- a/resources/vbmc/Dockerfile
+++ b/resources/vbmc/Dockerfile
@@ -1,7 +1,10 @@
-FROM registry.hub.docker.com/library/python:3.7
+FROM docker.io/centos:centos8
 
-RUN apt update && \
-    apt install -y libvirt-dev && \
-    pip3 install virtualbmc
+RUN dnf install -y python3 python3-requests && \
+    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b master current && \
+    dnf update -y && \
+    dnf install -y python3-virtualbmc && \
+    dnf clean all && \
+    rm -rf /var/cache/{yum,dnf}/*
 
-CMD /usr/local/bin/vbmcd --foreground
+CMD /usr/bin/vbmcd --foreground


resources/vbmc/Dockerfile
```diff
@@ -1,7 +1,8 @@
 FROM registry.hub.docker.com/library/python:3.7
 
 RUN apt update && \
-    apt install -y libvirt-dev && \
-    pip3 install virtualbmc
+    apt install -y curl libvirt-dev && \
+    curl -Lf -o /tmp/uc.txt https://releases.openstack.org/constraints/upper/master && \
+    pip3 install virtualbmc -c /tmp/uc.txt
```
hardys: This seems reasonable, but I wonder if we might also consider switching to the RDO packages, since we already use those for the ironic images:

https://github.com/metal3-io/ironic-image/blob/master/Dockerfile#L31..L34

Then we'd likely get a more tested virtualbmc version than relying on pip install where there is no CI/promotion pipeline?

Any thoughts on that @elfosardo @maelk ?

resources/vbmc/Dockerfile
```diff
@@ -1,7 +1,8 @@
 FROM registry.hub.docker.com/library/python:3.7
 
 RUN apt update && \
-    apt install -y libvirt-dev && \
-    pip3 install virtualbmc
+    apt install -y curl libvirt-dev && \
+    curl -Lf -o /tmp/uc.txt https://releases.openstack.org/constraints/upper/master && \
```
elfosardo: instead of using curl we can use -c with the url directly

dtantsur: Oh, it works? Nice!

resources/vbmc/Dockerfile
dtantsur: No objections from my side, unless people will be sad about switching to CentOS.

elfosardo: the CI/promotion of vbmc comes from the openstack testing infrastructure, so it's not like we don't get any testing
not against using RDO pkgs, will probably get better integration testing with that and the centos8 repos are very close to be finally stable, so we can definitely consider that

dtantsur:
@hardys @elfosardo updated to use packages

PR #243, opened 2020-03-05
Update dev-env with new CAPM3 repo

fmuyassarov:
A new repo https://github.com/metal3-io/cluster-api-provider-metal3 is created and this patch 
- updates variables to clone CAPM3 in case CAPI_VERSION=v1alpha3
- fixes [#200](https://github.com/metal3-io/metal3-dev-env/issues/200)
- updates kinds in CR templates

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -28,20 +28,23 @@ export GOPATH
 M3PATH="${GOPATH}/src/github.com/metal3-io"
 BMOPATH="${M3PATH}/baremetal-operator"
 RUN_LOCAL_IRONIC_SCRIPT="${BMOPATH}/tools/run_local_ironic.sh"
-CAPM3PATH="${M3PATH}/cluster-api-provider-baremetal"
-
-BMOREPO="${BMOREPO:-https://github.com/metal3-io/baremetal-operator.git}"
-BMOBRANCH="${BMOBRANCH:-master}"
-CAPM3REPO="${CAPM3REPO:-https://github.com/metal3-io/cluster-api-provider-baremetal.git}"
+CAPM3PATH="${M3PATH}/cluster-api-provider-metal3"
 
 if [ "${CAPI_VERSION}" == "v1alpha3" ]; then
   CAPM3BRANCH="${CAPM3BRANCH:-master}"
+  CAPM3REPO="${CAPM3REPO:-https://github.com/metal3-io/cluster-api-provider-metal3.git}"
 elif [ "${CAPI_VERSION}" == "v1alpha2" ]; then
+  CAPM3PATH="${M3PATH}/cluster-api-provider-baremetal"
   CAPM3BRANCH="${CAPM3BRANCH:-release-0.2}"
+  CAPM3REPO="${CAPM3REPO:-https://github.com/metal3-io/cluster-api-provider-baremetal.git}"
 elif [ "${CAPI_VERSION}" == "v1alpha1" ]; then
+  CAPM3PATH="${M3PATH}/cluster-api-provider-baremetal"
   CAPM3BRANCH="${CAPM3BRANCH:-v1alpha1}"
+  CAPM3REPO="${CAPM3REPO:-https://github.com/metal3-io/cluster-api-provider-baremetal.git}"
 fi
 
+BMOREPO="${BMOREPO:-https://github.com/metal3-io/baremetal-operator.git}"
+BMOBRANCH="${BMOBRANCH:-master}"
 FORCE_REPO_UPDATE="${FORCE_REPO_UPDATE:-false}"
 
 BMO_RUN_LOCAL="${BMO_RUN_LOCAL:-false}"
--- a/04_verify.sh
+++ b/04_verify.sh
@@ -105,7 +105,7 @@ check_k8s_entity() {
     fi
     process_status $?
 
-    # Check the replicas
+    # Check the replicabaremetalclusters.s
     if [[ "${BMO_RUN_LOCAL}" != true ]] && [[ "${CAPM3_RUN_LOCAL}" != true ]]
     then
       RESULT_STR="${name} ${TYPE} replicas correct"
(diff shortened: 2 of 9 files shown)


03_launch_mgmt_cluster.sh
```diff
@@ -28,20 +28,23 @@ export GOPATH
...
+  CAPM3PATH="${M3PATH}/cluster-api-provider-baremetal"
   CAPM3BRANCH="${CAPM3BRANCH:-release-0.2}"
+  CAPM3REPO="${CAPM3REPO:-https://github.com/metal3-io/cluster-api-provider-baremetal.git}"
 elif [ "${CAPI_VERSION}" == "v1alpha1" ]; then
+  CAPM3PATH="${M3PATH}/cluster-api-provider-baremetal"
   CAPM3BRANCH="${CAPM3BRANCH:-v1alpha1}"
+  CAPM3REPO="${CAPM3REPO:-https://github.com/metal3-io/cluster-api-provider-baremetal.git}"
 fi
 
+BMOREPO="${BMOREPO:-https://github.com/metal3-io/baremetal-operator.git}"
```
Xenwar: also replace all occurences of BMO.
Now it is a mix of both

03_launch_mgmt_cluster.sh
```diff
@@ -28,20 +28,23 @@ export GOPATH
...
-CAPM3PATH="${M3PATH}/cluster-api-provider-baremetal"
-
-BMOREPO="${BMOREPO:-https://github.com/metal3-io/baremetal-operator.git}"
-BMOBRANCH="${BMOBRANCH:-master}"
-CAPM3REPO="${CAPM3REPO:-https://github.com/metal3-io/cluster-api-provider-baremetal.git}"
 
 if [ "${CAPI_VERSION}" == "v1alpha3" ]; then
+  CAPM3PATH="${M3PATH}/cluster-api-provider-metal3"
   CAPM3BRANCH="${CAPM3BRANCH:-master}"
+  CAPM3REPO="${CAPM3REPO:-https://github.com/metal3-io/cluster-api-provider-metal3.git}"
```
kashifest: May be default it to something and once it is not default change it rather than default the env variable for each different case?

vm-setup/roles/v1aX_integration_test/templates/controlplane_ubuntu.yaml
```diff
@@ -128,11 +128,19 @@ spec:
       name: {{ CLUSTER_NAME }}-controlplane-0
   infrastructureRef:
     apiVersion: infrastructure.cluster.x-k8s.io/{{ CAPI_VERSION }}
+{% if CAPI_VERSION == 'v1alpha3' %}
```
kashifest: may be if CAPI_VERSION != v1alpha2 && CAPI_VERSION !=v1alpha1, since this change is here to stay I suppose. So, kind will be the same even for v1alpha4 etc.

vm-setup/roles/v1aX_integration_test/templates/controlplane_ubuntu.yaml
```diff
@@ -128,11 +128,19 @@ spec:
...
+    kind: Metal3Machine
+{% else %}
     kind: BareMetalMachine
+{% endif %}
     name: {{ CLUSTER_NAME }}-controlplane-0
 ---
 apiVersion: infrastructure.cluster.x-k8s.io/{{ CAPI_VERSION }}
+{% if CAPI_VERSION == 'v1alpha3' %}
```
kashifest: Same as above

vm-setup/roles/v1aX_integration_test/templates/workers_centos.yaml
```diff
@@ -32,10 +32,18 @@ spec:
       infrastructureRef:
         name: {{ CLUSTER_NAME }}-md-0
         apiVersion: infrastructure.cluster.x-k8s.io/{{ CAPI_VERSION }}
+{% if CAPI_VERSION == 'v1alpha3' %}
```
kashifest: Same as above

vm-setup/roles/v1aX_integration_test/templates/workers_centos.yaml
```diff
@@ -32,10 +32,18 @@ spec:
...
+        kind: Metal3MachineTemplate
+{% else %}
         kind: BareMetalMachineTemplate
+{% endif %}
 ---
 apiVersion: infrastructure.cluster.x-k8s.io/{{ CAPI_VERSION }}
-kind: BareMetalMachineTemplate
+{% if CAPI_VERSION == 'v1alpha3' %}
```
kashifest: Same as above

vm-setup/roles/v1aX_integration_test/templates/workers_ubuntu.yaml
```diff
@@ -32,10 +32,18 @@ spec:
       infrastructureRef:
         name: {{ CLUSTER_NAME }}-md-0
         apiVersion: infrastructure.cluster.x-k8s.io/{{ CAPI_VERSION }}
+{% if CAPI_VERSION == 'v1alpha3' %}
```
kashifest: Same as above

vm-setup/roles/v1aX_integration_test/templates/workers_ubuntu.yaml
```diff
@@ -32,10 +32,18 @@ spec:
...
+        kind: Metal3MachineTemplate
+{% else %}
         kind: BareMetalMachineTemplate
+{% endif %}
 ---
 apiVersion: infrastructure.cluster.x-k8s.io/{{ CAPI_VERSION }}
-kind: BareMetalMachineTemplate
+{% if CAPI_VERSION == 'v1alpha3' %}
```
kashifest: Same as above

vm-setup/roles/v1aX_integration_test/templates/controlplane_ubuntu.yaml
fmuyassarov: good point!

vm-setup/roles/v1aX_integration_test/templates/cluster.yaml
```diff
@@ -12,17 +12,23 @@ spec:
     serviceDomain: "cluster.local"
   infrastructureRef:
     apiVersion: infrastructure.cluster.x-k8s.io/{{ CAPI_VERSION }}
+{% if CAPI_VERSION == 'v1alpha2' %}
```
kashifest: doesn't it mean v1alpha1 now would have kind: Metal3Cluster

kashifest: Ah for that we have different templates I suppose

vm-setup/roles/v1aX_integration_test/templates/controlplane_ubuntu.yaml
fmuyassarov: Since these CR templates are only for v1a2/v1a3/v1ax,  I would rather go for for example:
```
{% if CAPI_VERSION == 'v1alpha2' %}
kind: BareMetalMachine
{% else %}
kind: Metal3Machine
{% endif %}
```
because in case of v1alpha1 we use completely different [templates ](https://github.com/metal3-io/metal3-dev-env/blob/master/scripts/v1alpha1/create_machine.sh)

03_launch_mgmt_cluster.sh
fmuyassarov: Agree, updated

vm-setup/roles/v1aX_integration_test/templates/cluster.yaml
fmuyassarov: yes

04_verify.sh
```diff
@@ -196,6 +196,16 @@ EXPTD_V1ALPHA2_CRDS="clusters.cluster.x-k8s.io \
   baremetalhosts.metal3.io \
   baremetalmachines.infrastructure.cluster.x-k8s.io \
   baremetalmachinetemplates.infrastructure.cluster.x-k8s.io"
+EXPTD_V1ALPHA3_CRDS="clusters.cluster.x-k8s.io \
```
kashifest: I am now thinking is this the correct way or should we have two sets one for common CRDs and one for extra CRDs which are version specific, otherwise this list will be exhaustive? Just a suggestion

04_verify.sh
```diff
@@ -253,7 +263,14 @@ RESULT_STR="Fetch CRDs"
 CRDS="$(kubectl --kubeconfig "${KUBECONFIG}" get crds)"
 process_status $? "Fetch CRDs"
 
-if [[ "${CAPI_VERSION}" == "v1alpha2" ]] || [[ "${CAPI_VERSION}" == "v1alpha3" ]]; then
+if [ "${CAPI_VERSION}" == "v1alpha3" ]; then
```
kashifest: Similar comment as above. May be we should populate the list of CRDs based on the version once and then run the check once only rather than having similar code for all the versions.

04_verify.sh
fmuyassarov: good point, done

04_verify.sh
fmuyassarov: done

maelk:
This needs the project-infra PR to be in to pass.

vm-setup/roles/v1aX_integration_test/templates/cluster.yaml
```diff
@@ -12,17 +12,23 @@ spec:
...
     kind: BareMetalCluster
+{% else %}
+    kind: Metal3Cluster
     name: {{ CLUSTER_NAME }}
```
maelk: The name is needed for v1alpha2 also!

PR #247, opened 2020-03-09
Update sushy container image

derekhiggins:
We need sushy-tools==0.8.0 which includes a fix for
the case expected when setting boot mode to UEFI.

We need this patch , https://review.opendev.org/#/c/688458/6/sushy_tools/emulator/resources/systems/libvirtdriver.py

### DIFF
--- a/resources/sushy-tools/Dockerfile
+++ b/resources/sushy-tools/Dockerfile
@@ -2,6 +2,6 @@ FROM registry.hub.docker.com/library/python:3.7
 
 RUN apt update && \
     apt install -y libvirt-dev && \
-    pip3 install sushy-tools==0.7.0 libvirt-python
+    pip3 install sushy-tools==0.8.0 libvirt-python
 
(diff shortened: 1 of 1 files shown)


maelk:
Can you please try a rebase ? It looks like the base branch is old

hardys:
> Can you please try a rebase ? It looks like the base branch is old

@maelk I noticed this is a common request - would it make sense to adjust CI so that it rebases any PR branch on the latest HEAD?

I guess we really want to prove that situation anyway, since if we merge the branch it's going to be on top of HEAD and not whatever commit the branch was originally created from?

Clearly there will be cases where the branch won't rebase cleanly, but it's probably fine to just fail in that case since since any conflicts have to be resolved to be mergeable anyway?

maelk:
@hardys Yes definitely! It is on my backlog, we already discussed it with Russell if I recall correctly. I just did not have time to do that yet. I'll try to prioritize it.

PR #250, opened 2020-03-11
Rename test scripts to reflect v1a1 deprecation

maelk:
V1a1 is now deprecated, and v1a3 default. So the test scripts and Makefile should reflect this.


Makefile
```diff
@@ -23,11 +23,14 @@ host_cleanup:
...
+test_v1a1:
+	./05_test_v1alpha1.sh
+
 test_v1a2:
-	./05_test_v1aX.sh
+	./05_test.sh
 
 test_v1a3:
-	./05_test_v1aX.sh
+	./05_test.sh
```
fmuyassarov: Do we need those? If I understood right based on [#64](https://github.com/metal3-io/project-infra/pull/64), we will have only two targets, one for v1alpha1, and another for v1alpha2/3/4/..

maelk: yes, but for backwards compatibility, we should keep them around for a while

PR #253, opened 2020-03-12
Add ability to specity external SRV records

bcrochet:
If multicast is not available, mDNS won't work. It is then necessary
to specify external SRV records for dnsmasq to use.


vm-setup/roles/common/defaults/main.yml
```diff
@@ -63,3 +63,4 @@ networks:
       forwarders:
         - domain: "apps.{{ cluster_domain }}"
           addr: "{% if baremetal_network_cidr|ipv6 != False %}::1{% else %}127.0.0.1{% endif %}"
+      srvs: "{{dns_externalsrvs | default([])}}"
```
hardys: Note for other reviewers, the support for this already exists here:

https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/libvirt/templates/network.xml.j2#L65

PR #255, opened 2020-03-13
Use Kubeadm Kubelet extraArgs to set providerID in node templates

maelk:
This patch adds an extra arg to kubelet in the KubeadmConfigs to
set the providerID, instead of setting `noCloudProvider` to let
CAPM3 set it itself based on the label.


jan-est:
This looks good to me

PR #256, opened 2020-03-13
Make openstackclient.sh more flexible

hardys:
Currently this is hard-coded so it always uses OS_CLOUD=metal3
and can only consume the _clouds_yaml from metal3-dev-env.

To enable compatibility with https://github.com/openshift-metal3/dev-scripts/
Check $PWD for an alternative _clouds_yaml, and allow for overriding
OS_CLOUD from the environment.

dev-scripts reuses some parts of metal3-dev-env for testing metal3
enabled openshift so this will make openstackclient.sh useful in
both repos.


hardys:
Part of the fix for https://github.com/openshift-metal3/dev-scripts/issues/971

PR #257, opened 2020-03-13
Set API version explicitly to v1alpha2 when condition is met.

fmuyassarov:
Set API version explicitly to v1alpha2 when v1alpha2 condition is met

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/templates/cluster.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/cluster.yaml
@@ -1,6 +1,6 @@
 ---
 {% if CAPI_VERSION == 'v1alpha2' %}
-apiVersion: cluster.x-k8s.io/{{ CAPI_VERSION }}
+apiVersion: cluster.x-k8s.io/v1alpha2
 {% else %}
 apiVersion: cluster.x-k8s.io/v1alpha3
 {% endif %}
--- a/vm-setup/roles/v1aX_integration_test/templates/workers_centos.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/workers_centos.yaml
@@ -1,5 +1,5 @@
 {% if CAPI_VERSION == 'v1alpha2' %}
-apiVersion: cluster.x-k8s.io/{{ CAPI_VERSION }}
+apiVersion: cluster.x-k8s.io/v1alpha2
 {% else %}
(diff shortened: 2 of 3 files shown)


maelk:
Should we rather aim at using CentOS8, so that we don't rely on our own image ?

fmuyassarov:
> Should we rather aim at using CentOS8, so that we don't rely on our own image ?

Yes, agree. But not sure if provisioning is going to work without any failures, but I'll test with the CentOS8 image that we have in CityCloud.

fmuyassarov:
/hold 
testing with CentOS8

lib/images.sh
```diff
@@ -11,8 +11,8 @@ elif [[ "${IMAGE_OS}" == "FCOS" ]]; then
   export IMAGE_LOCATION=${IMAGE_LOCATION:-https://builds.coreos.fedoraproject.org/prod/streams/testing/builds/30.20191014.0/x86_64/}
   export IMAGE_USERNAME=${IMAGE_USERNAME:-core}
 elif [[ "${IMAGE_OS}" == "Centos" ]]; then
-  export IMAGE_NAME=${IMAGE_NAME:-CentOS-7-x86_64-GenericCloud-1907.qcow2}
-  export IMAGE_LOCATION=${IMAGE_LOCATION:-http://cloud.centos.org/centos/7/images}
+  export IMAGE_NAME=${IMAGE_NAME:-centos-updated.qcow2}
+  export IMAGE_LOCATION=${IMAGE_LOCATION:-http://artifactory.nordix.org/artifactory/airship/images/centos.qcow2}
```
maelk: This needs to be changed to CentOS8

maelk:
please update the description if we go with CentOS8

lib/images.sh
fmuyassarov: yes, I'll for now I just want to make sure it is going to work with CentOS8

fmuyassarov:
> please update the description if we go with CentOS8

sure, will do that

alosadagrande:
Related issue #263 when creating CentOS 8 control-plane VM in the target cluster

fmuyassarov:
> Related: issue #263 when creating CentOS 8 control-plane VM in the target cluster.
> 
> The idea is to replace centos-update.qcow2 (based in CentOS 7) with CentOS 8.

Thanks @alosadagrande . Can you then please open a PR once you finalize your tests

PR #261, opened 2020-03-17
Remove duplicated Ansible tasks

fmuyassarov:
Couple of the Ansible tasks are duplicated and were forgotten to be removed.
These particular tasks are now located in [download_image.yml](https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/tasks/download_image.yml)

This patchset:
1. removes duplicated tasks
2. adds tasks to download ubuntu image and creates its checksum on the fly

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/download_image.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/download_image.yml
@@ -2,13 +2,13 @@
   - name: Get the facts about local CentOS image
     stat:
       path: "{{ IRONIC_IMAGE_DIR }}/{{ IMAGE_NAME_CENTOS }}"
-    register: image_path
+    register: image_path_centos
 
   - debug:
       msg: "Local image of CentOS is found"
     when:
       - IMAGE_OS == "Centos"
-      - image_path.stat.exists == True
+      - image_path_centos.stat.exists == True
 
   - name: Download CentOS image.
     block:
@@ -21,7 +21,7 @@
           dest: "{{ IRONIC_IMAGE_DIR }}/{{ IMAGE_NAME_CENTOS }}"
           mode: 0664
 
-      - name: Calculate md5sum of the specific image
+      - name: Calculate md5sum of the centos image
         stat:
          path: "{{ IRONIC_IMAGE_DIR }}/{{ IMAGE_NAME_CENTOS }}"
          checksum_algorithm: md5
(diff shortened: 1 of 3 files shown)


Jaakko-Os:
looks good!

maelk:
looks good, but should we also take this opportunity to remove the condition on the image download tasks so that it would apply for all images (for example you instantiated metal3-dev-env with Centos target host image, but then want to deploy Ubuntu. This is not supported until now.

fmuyassarov:
> looks good, but should we also take this opportunity to remove the condition on the image download tasks so that it would apply for all images (for example you instantiated metal3-dev-env with Centos target host image, but then want to deploy Ubuntu. This is not supported until now.

Done

vm-setup/roles/v1aX_integration_test/tasks/download_image.yml
```diff
@@ -6,9 +6,7 @@
 
   - debug:
       msg: "Local image of CentOS is found"
```
maelk: This is not centos only anymore

vm-setup/roles/v1aX_integration_test/tasks/download_image.yml
```diff
@@ -6,9 +6,7 @@
...
-    when:
-      - IMAGE_OS == "Centos"
-      - image_path.stat.exists == True
+    when: image_path.stat.exists == True
 
   - name: Download CentOS image.
```
maelk: same here

maelk:
small nits, otherwise lgtm

vm-setup/roles/v1aX_integration_test/tasks/download_image.yml
fmuyassarov: I think it still downloads only the CentOS, the only the difference is that there is no condition to run this playbook, no?

maelk: Yes, it should be for all images

maelk:
Sorry, my bad, since the variables are OS specific, then you need to keep the condition, and then add a new block for Ubuntu. Sorry I read the code too fast.

fmuyassarov:
> Sorry, my bad, since the variables are OS specific, then you need to keep the condition, and then add a new block for Ubuntu. Sorry I read the code too fast.

No problem, now it is done for both centos&ubuntu. Thanks

vm-setup/roles/v1aX_integration_test/tasks/download_image.yml
```diff
@@ -35,5 +35,44 @@
...
       - IMAGE_OS == "Centos"
-      - image_path.stat.exists == False
+
+  - name: Get the facts about local Ubuntu image
+    stat:
+      path: "{{ IRONIC_IMAGE_DIR }}/{{ IMAGE_NAME }}"
+    register: image_path_ubuntu
+
+  - debug:
+      msg: "Local image of CentOS is found"
```
maelk: Ubuntu

Issue #263, opened 2020-03-18
Control-plane creation error when using CentOS 8 as IMAGE_NAME_CENTOS


alosadagrande:
I have been testing the official cloud CentOS 8 image as IMAGE_NAME_CENTOS 
 in metal3-dev-env without success. 

Some issues/ facts:

- I am using cloud image [CentOS-8-GenericCloud-8.1.1911-20200113.3.x86_64.qcow2](https://cloud.centos.org/centos/8/x86_64/images/CentOS-8-GenericCloud-8.1.1911-20200113.3.x86_64.qcow2)
- CentOS 8 ships cloud-init 18.5, so it is the same version as the centos-update.qcow2 that works perfectly OK.
- Found some issues when moving from CentOS7 to 8 with docker dependencies. They were addressed by adding --nobest option to yum (worth researching cri-o or containerd support in kubeadm. Probably once we get this fixed with docker):

```yaml    
    preKubeadmCommands:
      - ifup eth1
      - yum update -y
      - yum install yum-utils -y
      - yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
      - setenforce 0
      - sed -i 's/^SELINUX=enforcing$/SELINUX=permissive/' /etc/selinux/config
      - yum install docker-ce docker-ce-cli --disableexcludes=kubernetes --nobest -y
      - >-
        yum install gcc kernel-headers kernel-devel keepalived
        device-mapper-persistent-data lvm2
        kubelet kubeadm kubectl --disableexcludes=kubernetes -y
      - usermod -aG docker centos
      - systemctl enable --now docker keepalived kubelet
```

Then everytime I run the script provision_controlplane.sh I always stuck at the fact that the /tmp/kubeadm.yaml file is not created in the control-plane VM. So I understand that kubeadm init command cannot be executed and provisioning is not finished.

From the cloud-init logs I can see:

```
2020-03-18 16:31:01,467 - util.py[WARNING]: Running module write-files (<module 'cloudinit.config.cc_write_files' from '/usr/lib/python3.6/site-packages/cloudinit/config/cc_write_files.py'>) failed
2020-03-18 16:31:01,467 - util.py[DEBUG]: Running module write-files (<module 'cloudinit.config.cc_write_files' from '/usr/lib/python3.6/site-packages/cloudinit/config/cc_write_files.py'>) failed
Traceback (most recent call last):
[...]

PR #264, opened 2020-03-20
Target cluster OS set to Centos8 instead of CentOS 7

alosadagrande:
This PR focuses on set CentOS 8 as the default option when selecting Centos as base image for the target cluster. It replaces CentOS 7 as the default CentOS version.

- Linked to issue #263 

Notes:
==========

- Changes in the lib/images.sh to set CentOS 8 as the default
- Changes in the templates for worker and control-plane since CentOS 8 does not run latest docker. Basically, included --nobest option which installs Docker 18.9.1 (docker-ce-18.09.1-3.el7.x86_64). Additionally configure to work with the official cloud image instead of the modified centos-update.qcow2.
- The default OS_IMAGE is changed to Centos instead of Ubuntu. Since the default image that is downloaded with make scripts is CentOS - now, version 8 qcow2 image. It does not make sense to download Centos cloud image with make scripts and then download the Ubuntu image using the Ansible playbook. The ironic folder ends up with two images. 

I set the default to Centos in the Ansible vars yaml file since Centos image is downloaded during metal3-dev-env configuration time. However, it can be changed to set Ubuntu the default, basically changing in both sides (images.sh and Ansible provision playbook).
- Added an extra task during cluster provisioning in case CentOS 7 is the OS for the server where the metal3-dev-env is set up. It should be pointed in the documentation that CentOS 8 is the suggested one along with Ubuntu 18.04.

Extra details:
===============

- I have been able to deploy the metal3-dev-env on a physical server running CentOS 8 and CentOS 7 both provisioning the target control-plane and workers with the cloud CentOS 8 image. 

```sh
[centos@node-2 ~]$ kubectl  get nodes -o wide
NAME     STATUS   ROLES    AGE   VERSION   INTERNAL-IP      EXTERNAL-IP   OS-IMAGE                KERNEL-VERSION                CONTAINER-RUNTIME
node-1   Ready    <none>   10h   v1.17.4   192.168.111.21   <none>        CentOS Linux 8 (Core)   4.18.0-147.3.1.el8_1.x86_64   docker://18.9.1
node-2   Ready    master   15h   v1.17.4   192.168.111.22   <none>        CentOS Linux 8 (Core)   4.18.0-147.3.1.el8_1.x86_64   docker://18.9.1
node-3   Ready    <none>   10h   v1.17.4   192.168.111.23   <none>        CentOS Linux 8 (Core)   4.18.0-147.3.1.el8_1.x86_64   docker://18.9.1
node-4   Ready    <none>   10h   v1.17.4   192.168.111.24   <none>        CentOS Linux 8 (Core)   4.18.0-147.3.1.el8_1.x86_64   docker://18.9.1
```

I think it's time to set the default CentOS version to major release 8 in both the target cluster and the metal3-dev-env host that runs libvirtd. Try-it document must be updated accordingly once the PR is merged.

In the near future, centos-update.qcow2 can be decomissioned since now downloads the public cloud image from Centos download page.

### DIFF
--- a/lib/images.sh
+++ b/lib/images.sh
@@ -11,8 +11,8 @@ elif [[ "${IMAGE_OS}" == "FCOS" ]]; then
   export IMAGE_LOCATION=${IMAGE_LOCATION:-https://builds.coreos.fedoraproject.org/prod/streams/testing/builds/30.20191014.0/x86_64/}
   export IMAGE_USERNAME=${IMAGE_USERNAME:-core}
 elif [[ "${IMAGE_OS}" == "Centos" ]]; then
-  export IMAGE_NAME=${IMAGE_NAME:-CentOS-7-x86_64-GenericCloud-1907.qcow2}
-  export IMAGE_LOCATION=${IMAGE_LOCATION:-http://cloud.centos.org/centos/7/images}
+  export IMAGE_NAME=${IMAGE_NAME:-CentOS-8-GenericCloud-8.1.1911-20200113.3.x86_64.qcow2}
+  export IMAGE_LOCATION=${IMAGE_LOCATION:-https://cloud.centos.org/centos/8/x86_64/images/}
   export IMAGE_USERNAME=${IMAGE_USERNAME:-centos}
 else
   export IMAGE_NAME=${IMAGE_NAME:-cirros-0.4.0-x86_64-disk.img}
--- a/vm-setup/roles/v1aX_integration_test/tasks/provision_cluster.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/provision_cluster.yml
@@ -1,12 +1,21 @@
 ---
-  - name: Install openshift client
+  - name: Install openshift client in CentOS 8 using pip3
     pip:
       executable: "pip3"
       name: openshift
       state: present
     become: true
     become_user: root
-    when: ansible_facts['distribution'] == "CentOS"
+    when: ansible_facts['distribution'] == "CentOS" and ansible_facts['distribution_major_version'] == '8'
+
+  - name: Install openshift client in CentOS 7 using pip
+    pip:
+      executable: "pip"
+      name: openshift
+      state: present
+    become: true
+    become_user: root
+    when: ansible_facts['distribution'] == "CentOS" and ansible_facts['distribution_major_version'] == '7'
 
 # Populate cr with environment variables
   - name: Template cluster.yaml
--- a/vm-setup/roles/v1aX_integration_test/templates/controlplane_centos.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/controlplane_centos.yaml
@@ -51,10 +51,10 @@ spec:
     - yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
     - setenforce 0
     - sed -i 's/^SELINUX=enforcing$/SELINUX=permissive/' /etc/selinux/config
+    - yum install docker-ce docker-ce-cli --disableexcludes=kubernetes --nobest -y
     - >-
       yum install gcc kernel-headers kernel-devel keepalived
       device-mapper-persistent-data lvm2
-      docker-ce-18.09.9 docker-ce-cli-18.09.9 containerd.io
       kubelet kubeadm kubectl --disableexcludes=kubernetes -y
     - usermod -aG docker centos
     - systemctl enable --now docker keepalived kubelet
@@ -105,8 +105,8 @@ spec:
         gpgcheck=1
         repo_gpgcheck=0
         gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
-    - path: /home/centos/.ssh/authorized_keys
-      owner: centos:centos
+    - path: /tmp/akeys
+      owner: root:root
       permissions: '0600'
       content: {{ SSH_PUB_KEY_CONTENT }}
 {% else %}
@@ -139,15 +139,17 @@ spec:
         extraArgs:
     preKubeadmCommands:
       - ifup eth1
+      - mv /tmp/akeys /home/centos/.ssh/authorized_keys
+      - chown centos:centos /home/centos/.ssh/authorized_keys
       - yum update -y
       - yum install yum-utils -y
       - yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
       - setenforce 0
       - sed -i 's/^SELINUX=enforcing$/SELINUX=permissive/' /etc/selinux/config
+      - yum install docker-ce docker-ce-cli --disableexcludes=kubernetes --nobest -y
       - >-
         yum install gcc kernel-headers kernel-devel keepalived
         device-mapper-persistent-data lvm2
-        docker-ce-18.09.9 docker-ce-cli-18.09.9 containerd.io
         kubelet kubeadm kubectl --disableexcludes=kubernetes -y
       - usermod -aG docker centos
       - systemctl enable --now docker keepalived kubelet
@@ -198,8 +200,8 @@ spec:
           gpgcheck=1
           repo_gpgcheck=0
           gpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg
-      - path: /home/centos/.ssh/authorized_keys
-        owner: centos:centos
+      - path: /tmp/akeys
+        owner: root:root
         permissions: '0600'
         content: {{ SSH_PUB_KEY_CONTENT }}
 ---
--- a/vm-setup/roles/v1aX_integration_test/templates/workers_centos.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/workers_centos.yaml
@@ -76,14 +76,15 @@ spec:
             provider-id: 'metal3://{{ '{{ ds.meta_data.uuid }}' }}'
       preKubeadmCommands:
         - ifup eth1
+        - mv /tmp/akeys /home/centos/.ssh/authorized_keys
+        - chown centos:centos /home/centos/.ssh/authorized_keys
         - yum update -y
         - yum install yum-utils device-mapper-persistent-data lvm2 -y
         - yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
         - setenforce 0
         - sed -i 's/^SELINUX=enforcing$/SELINUX=permissive/' /etc/selinux/config
-        - >-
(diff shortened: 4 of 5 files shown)


alosadagrande:
I am not sure if this issue #265 is related or not to CentOS 8 replacement.

vm-setup/roles/v1aX_integration_test/tasks/provision_cluster.yml
```diff
@@ -1,12 +1,21 @@
...
       name: openshift
       state: present
     become: true
     become_user: root
-    when: ansible_facts['distribution'] == "CentOS"
+    when: ansible_facts['distribution'] == "CentOS" and ansible_facts['distribution_major_version'] == '8'
+
+  - name: Install openshift client in CentOS 7 using pip
+    pip:
+      executable: "pip"
```
fmuyassarov: Could you please share why is it required to use pip instead of pip3?

lib/images.sh
```diff
@@ -11,8 +11,8 @@ elif [[ "${IMAGE_OS}" == "FCOS" ]]; then
   export IMAGE_LOCATION=${IMAGE_LOCATION:-https://builds.coreos.fedoraproject.org/prod/streams/testing/builds/30.20191014.0/x86_64/}
   export IMAGE_USERNAME=${IMAGE_USERNAME:-core}
 elif [[ "${IMAGE_OS}" == "Centos" ]]; then
-  export IMAGE_NAME=${IMAGE_NAME:-CentOS-7-x86_64-GenericCloud-1907.qcow2}
-  export IMAGE_LOCATION=${IMAGE_LOCATION:-http://cloud.centos.org/centos/7/images}
+  export IMAGE_NAME=${IMAGE_NAME:-CentOS-8-GenericCloud-8.1.1911-20200113.3.x86_64.qcow2}
+  export IMAGE_LOCATION=${IMAGE_LOCATION:-https://cloud.centos.org/centos/8/x86_64/images/}
```
fmuyassarov: why isn't value the same as in vm-setup/roles/v1aX_integration_test/vars/main.yml?

fmuyassarov:
@alosadagrande Also note that, even if the IMAGE_OS is set to be Ubuntu by default, at least the CI still going to set it to be Ubuntu. And, since CI pipeline is the first occurrence of this variable, it is going to be inherited all the way till 05 scripts, meaning that IMAGE_OS variable will have Ubuntu value.
@maelk I think we need to change the JJB pipelines for this variable?

maelk:
We need to keep the CI with Ubuntu, due to RAM restrictions. The Centos image is too big.

lib/images.sh
alosadagrande: I replaced the previous value of centos-update to be the same, see: https://github.com/metal3-io/metal3-dev-env/pull/264/files#diff-ef7461d6dcc137f60cbd6cc5f23fbef5R29

maelk:
Either or is fine by me. The CI sets the variable properly anyways to override the default behavior so it can be set to one or the other

fmuyassarov:
It is fine for me too, since it is going to be replaced anyway by the CI. I just wanted to let you know that if you see later that Ubuntu was chosen for the target it is because of the CI. So, both are fine for me too.

lib/images.sh
fmuyassarov: Ah sorry, my bad

vm-setup/roles/v1aX_integration_test/tasks/provision_cluster.yml
alosadagrande: Yes, this issue happens when the physical server which is running libvirtd is CentOS7. In this case, Ansible is running python2, therefore if you use pip3 to install the openshift client (provision_cluster playbook) since the ansible_python_interpreter is set to python2/pip2 it won't find the kubernetes module:

```sh
TASK [v1aX_integration_test : Provision cluster] **************************************************************************************************************
task path: /home/alosadag/metal3-dev-env/vm-setup/roles/v1aX_integration_test/tasks/provision_cluster.yml:17
<localhost> ESTABLISH LOCAL CONNECTION FOR USER: alosadag
<localhost> EXEC /bin/sh -c 'echo ~alosadag && sleep 0'
<localhost> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /home/alosadag/.ansible/tmp/ansible-tmp-1584957211.06-58949636475436 `" && echo ansible-tmp-1584957211.06-58949636475436="` echo /home/alosadag/.ansible/tmp/ansible-tmp-1584957211.06-58949636475436 `" ) && sleep 0'
Using module file /usr/lib/python2.7/site-packages/ansible/modules/clustering/k8s/k8s.py
<localhost> PUT /home/alosadag/.ansible/tmp/ansible-local-45010DrKySv/tmpL17g6P TO /home/alosadag/.ansible/tmp/ansible-tmp-1584957211.06-58949636475436/AnsiballZ_k8s.py
<localhost> EXEC /bin/sh -c 'chmod u+x /home/alosadag/.ansible/tmp/ansible-tmp-1584957211.06-58949636475436/ /home/alosadag/.ansible/tmp/ansible-tmp-1584957211.06-58949636475436/AnsiballZ_k8s.py && sleep 0'
<localhost> EXEC /bin/sh -c '/usr/bin/python /home/alosadag/.ansible/tmp/ansible-tmp-1584957211.06-58949636475436/AnsiballZ_k8s.py && sleep 0'
<localhost> EXEC /bin/sh -c 'rm -f -r /home/alosadag/.ansible/tmp/ansible-tmp-1584957211.06-58949636475436/ > /dev/null 2>&1 && sleep 0'
The full traceback is:
Traceback (most recent call last):
  File "/tmp/ansible_k8s_payload_tnA_vG/ansible_k8s_payload.zip/ansible/module_utils/k8s/common.py", line 33, in <module>
    import kubernetes
ImportError: No module named kubernetes
    "changed": false, 
    "error": "No module named kubernetes", 
    "invocation": {
        "module_args": {
            "api_key": null, 
            "api_version": "v1", 
            "append_hash": false, 
            "apply": false, 
            "ca_cert": null, 
            "client_cert": null, 
            "client_key": null, 
            "context": null, 
            "force": false, 
            "host": null, 
            "kind": null, 
            "kubeconfig": null, 
            "merge_type": null, 
            "name": null, 
            "namespace": "metal3", 
            "password": null, 
            "proxy": null, 
            "resource_definition": null, 
            "src": "/tmp/cluster.yaml", 
            "state": "present", 
            "username": null, 
            "validate": null, 
            "validate_certs": null, 
            "wait": false, 
            "wait_condition": null, 
            "wait_sleep": 5, 
            "wait_timeout": 120
        }
    }, 
    "msg": "Failed to import the required Python library (openshift) on eko1.cloud.lab.eng.bos.redhat.com's Python /usr/bin/python. Please read module documentation and install in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter"
}
```

I found two ways to fix it. One is the selecting the executable based on the major OS version. The other is setting the ansible_python_interpreter to python3 in the inventory:

```
$ cat  vm-setup/inventory.ini 

[virthost]
localhost ansible_python_interpreter=/usr/bin/python3
```

Which one do you think is better fix? @fmuyassarov @maelk

alosadagrande: Note as well, that CentOS 7 should be decomissioned in favour of CentOS 8, right?

fmuyassarov: From my point of view, I think it is better to use: `ansible_python_interpreter=/usr/bin/python3`, because if you add this it is anyway generic and will not hurt other OS based metal3-dev-env (where the libvirt is running) cases such as Ubuntu/Centos8 based.

alosadagrande: Ok dropped the commit that includes the conditional statement and added the ansible_python_interpreter to python3 in the inventory

alosadagrande:
> It is fine for me too, since it is going to be replaced anyway by the CI. I just wanted to let you know that if you see later that Ubuntu was chosen for the target it is because of the CI. So, both are fine for me too.

Verify that the Ansible role grabs the exported variables correctly via lookup. I am not sure, but I think it failed to me a couple of times. As I say, not sure, probably I was doing something wrong or in the wrong terminal :)

vm-setup/roles/v1aX_integration_test/tasks/provision_cluster.yml
alosadagrande: mmmm found the following error when running make. Probably it is safer to include the conditional instead of changing the interpreter. Otherwise we will need to replace yum by dnf and probably more things. 

```
 TASK [packages_installation : Install packages on rhel7] ******************************************************************************************************
    "changed": false, 
    "invocation": {
        "module_args": {
            "allow_downgrade": false, 
            "autoremove": false, 
            "bugfix": false, 
            "conf_file": null, 
            "disable_excludes": null, 
            "disable_gpg_check": false, 
            "disable_plugin": [], 
            "disablerepo": [], 
            "download_dir": null, 
            "download_only": false, 
            "enable_plugin": [], 
            "enablerepo": [], 
            "exclude": [], 
            "install_repoquery": true, 
            "install_weak_deps": true, 
            "installroot": "/", 
            "list": null, 
            "lock_timeout": 30, 
            "name": [
                "curl", 
                "dnsmasq", 
                "NetworkManager", 
                "nmap", 
                "OVMF", 
                "patch", 
                "psmisc", 
                "vim-enhanced", 
                "wget", 
                "ansible", 
                "bind-utils", 
                "jq", 
                "libguestfs-tools", 
                "redhat-lsb-core", 
                "unzip", 
                "genisoimage", 
                "qemu-kvm", 
                "libvirt", 
                "libvirt-devel", 
                "libvirt-daemon-kvm", 
                "libvirt-python", 
                "libguestfs-tools", 
                "python-lxml", 
                "polkit-pkla-compat", 
                "python-netaddr", 
                "virt-install", 
                "python-six"
            ], 
            "releasever": null, 
            "security": false, 
            "skip_broken": false, 
            "state": "present", 
            "update_cache": false, 
            "update_only": false, 
            "use_backend": "auto", 
            "validate_certs": true
        }
    }, 
    "msg": "The Python 2 bindings for rpm are needed for this module. If you require Python 3 support use the `dnf` Ansible module instead.. The Python 2 yum module is needed for this module. If you require Python 3 support use the `dnf` Ansible module instead."
}

```

fmuyassarov: Yes now I see. I had the same, one solution would be first to set alternatives to python3 and then modify inventory accordingly. 
But perhaps better to use your initial commit, which adds  conditional statement, otherwise there will be a need for more changes.
Sorry, for making extra work.

PR #266, opened 2020-03-21
Add machine status check in test playbook

maelk:
Verify that machines are in Running phase.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/verify.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/verify.yml
@@ -17,6 +17,16 @@
     delay: 20
     until: provisioned_bmh.stdout_lines ==  ["node-0", "node-1"]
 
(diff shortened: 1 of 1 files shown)


vm-setup/roles/v1aX_integration_test/tasks/verify.yml
```diff
@@ -17,6 +17,16 @@
     delay: 20
     until: provisioned_bmh.stdout_lines ==  ["node-0", "node-1"]
 
+  - name: Wait until both machines become running.
+    shell: |
+        kubectl get machines -n metal3 -o json | jq -r '.items[]
+        | select (.status.Phase == "Running")
```
fmuyassarov: I think it should be small ''p'' not capital in `.status.Phase`

maelk: good catch!

PR #269, opened 2020-03-24
Remove openstackclient container after use

derekhiggins:
Remove old openstackclient containers as they arn't reused

### DIFF
--- a/openstackclient.sh
+++ b/openstackclient.sh
@@ -11,5 +11,5 @@ else
 fi
 
 sudo "${CONTAINER_RUNTIME}" run -ti --net=host \
-  -v "${MOUNTDIR}:/etc/openstack" \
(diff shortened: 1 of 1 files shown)


maelk:
trying again. If it does not work, we can merge manually

maelk:
I removed the v1a2 centos job from the requirements as we do not support v1a2 anymore and it seems broken. Running the v1a3 centos test. if it passes, I will merge manually

hardys:
@maelk any idea how we get tide to merge this, seems like test-centos-integration did succeed, and test-v1a2-centos-integration is no longer needed?

maelk:
I'll merge it manually, the v1a2 test is broken. We could either amend the commit to discard the old tests, or just merge manually

PR #273, opened 2020-03-26
Freeze the version of kubelet, kubectl and kubeadm

### LINKED ISSUE #271: Freeze kubeadm, kubelet and kubectl version
Provisioning a control plane is failing for the reason shown below. 
Currently, we need to make changes such as [this](https://github.com/metal3-io/metal3-dev-env/pull/270) in order to start using the new version. 

However, this does not make the metal3-dev-env statble. 

Possible solution.
- Install specific versions of kubeadm, kubelet and kubectl (instead of latest).
- Verify that irrespective of new kuberenetes release the environment is usable

```
error execution phase preflight: [preflight] Some fatal errors occurred:
	[ERROR KubeletVersion]: the kubelet version is higher than the control plane version. This is not a supported version skew and may lead to a malfunctional cluster. Kubelet version: "1.18.0" Control plane version: "1.17.0"
[preflight] If you know what you are doing, you can make a check non-fatal with `--ignore-preflight-errors=...`
To see the stack trace of this error execute with --v=5 or higher
cp: cannot stat '/etc/kubernetes/admin.conf': No such file or directory
```

Xenwar:
As we have it now, ```KUBERNETES_VERSION``` is fixed in time. However, the binaries (kubelet, kubeadm and kubectl), are always the latest and this causes a mismatch between the two.

The purpose of this PR is then making both freeze in time and prevent auto update/upgrade of both. This allows developers to keep using specific version, if they wish to do so.

In future works, we intend to use the metal3-dev-env for testing upgrade workflow, where the kuberenetes and binaries upgrade is done independently. To that end, we have added a new variable. `KUBERNETES_BINARIES_VERSION` which defaults to the value of ```KUBERNETES_VERSION```
 
This PR makes the versions both of `KUBERNETES_VERSION` and `KUBERNETES_BINARIES_VERSION` set to ```1.17.0``` only for testing purpose.

Fixes #271


lib/common.sh
```diff
@@ -111,6 +111,7 @@ export DEFAULT_HOSTS_MEMORY=${DEFAULT_HOSTS_MEMORY:-8192}
 export CLUSTER_NAME=${CLUSTER_NAME:-"test1"}
 export CLUSTER_APIENDPOINT_IP=${CLUSTER_APIENDPOINT_IP:-"192.168.111.249"}
 export KUBERNETES_VERSION=${KUBERNETES_VERSION:-"v1.18.0"}
+export K3_BINARIES_VERSION=${K3_BINARIES_VERSION:-"1.18.0"}
```
fmuyassarov: why not to use KUBERNETES_VERSION variable ?

Xenwar: They are not in the same format. (v)
:-"v1.18.0"
:-"1.18.0"

fmuyassarov: I see, but then you should change also K3_BINARIES_VERSION to be `default('1.18.0', true) }}"` in vm-setup/roles/v1aX_integration_test/vars/main.yml

maelk: you could make KUBERNETES_VERSION to be 1.18.0 and then wherever used add the v in front, that would be better than creating this duplicated variable

maelk:
why K3 btw ?

maelk:
now you are missing the setting of the variable in the lib/common.sh. Kubernetes version is set there, we should have both at the same place. Also, if you add something into lib/common.sh, it should be considered whether that should be added in config_example.sh or not and on the try-it.md of the website

Xenwar:
@maelk 
Added the value in 
lib/common.sh
config_example.sh (this may not be needed as the Kubernetes version is not there either)
try-it.html

maelk:
I think it is obvious that kubernetes binaries cover kubeadm kubectl and kubelet

vm-setup/roles/v1aX_integration_test/vars/main.yml
```diff
@@ -7,7 +7,8 @@ CLUSTER_APIENDPOINT_HOST: "{{ lookup('env', 'CLUSTER_APIENDPOINT_HOST') }}"
 CLUSTER_APIENDPOINT_IP: "{{ lookup('env', 'CLUSTER_APIENDPOINT_IP') }}"
 CLUSTER_NAME: "{{ lookup('env', 'CLUSTER_NAME') | default('test1', true) }}"
 KUBECONFIG_PATH: "/home/ubuntu/.kube/config"
-KUBERNETES_VERSION: "{{ lookup('env', 'KUBERNETES_VERSION') | default('v1.18.0', true) }}"
+KUBERNETES_VERSION: "{{ lookup('env', 'KUBERNETES_VERSION') | default('v1.17.0', true) }}"
```
kashifest: any specific reason for not using 1.18?

Xenwar: See the description. only for testing purpose.

Xenwar:
> I think it is obvious that kubernetes binaries cover kubeadm kubectl and kubelet

I agree. We can remove it. It is needed for upgrading automated tests. Even in that it is only test case.

vm-setup/roles/v1aX_integration_test/vars/main.yml
kashifest: Ah sorry, missed that.

Xenwar:
> lgtm, just a nit, it could still be KUBERNETES_BINARIES_VERSION, and could default to `KUBERNETES_BINARIES_VERSION="${KUBERNETES_BINARIES_VERSION:-${KUBERNETES_VERSION#'v'}}"` that will just remove the v from the K8S version

Changed accordingly.

vm-setup/roles/v1aX_integration_test/vars/main.yml
```diff
@@ -7,7 +7,9 @@ CLUSTER_APIENDPOINT_HOST: "{{ lookup('env', 'CLUSTER_APIENDPOINT_HOST') }}"
...
+KUBERNETES_BINARIES_VERSION: "{{ lookup('env', 'KUBERNETES_BINARIES_VERSION') | default(lookup('env', 'KUBERNETES_VERSION' | replace('v','')))}}"
```
maelk: what if KUBERNETES_BINARIES_VERSION and KUBERNETES_VERSION are both not set ? you should also default the second lookup for KUBERNETES_VERSION the same way you did line 10

vm-setup/roles/v1aX_integration_test/vars/main.yml
```diff
@@ -7,7 +7,9 @@ CLUSTER_APIENDPOINT_HOST: "{{ lookup('env', 'CLUSTER_APIENDPOINT_HOST') }}"
 CLUSTER_APIENDPOINT_IP: "{{ lookup('env', 'CLUSTER_APIENDPOINT_IP') }}"
 CLUSTER_NAME: "{{ lookup('env', 'CLUSTER_NAME') | default('test1', true) }}"
 KUBECONFIG_PATH: "/home/ubuntu/.kube/config"
-KUBERNETES_VERSION: "{{ lookup('env', 'KUBERNETES_VERSION') | default('v1.18.0', true) }}"
+KUBERNETES_VERSION: "{{ lookup('env', 'KUBERNETES_VERSION') | default('v1.17.0', true) }}"
+KUBERNETES_BINARIES_VERSION: "{{ lookup('env', 'KUBERNETES_BINARIES_VERSION') | default(lookup('env', 'KUBERNETES_VERSION' | replace('v','')) | default('1.17.0', true) )}}"
```
maelk: There might be an issue in the first default (parenthesis on the lookup). Maybe it could be :
```
default(lookup('env', 'KUBERNETES_VERSION') | default('v1.17.0', true) | replace('v','') , true)
```

Xenwar: Done

vm-setup/roles/v1aX_integration_test/vars/main.yml
Xenwar: Done

config_example.sh
```diff
@@ -86,6 +86,11 @@
 #
 #export CAPI_VERSION=v1alpha3
 
+#export KUBERNETES_VERSION="1.18.0"
```
maelk: missing a v (v1.18.0)

vm-setup/roles/v1aX_integration_test/templates/controlplane_ubuntu.yaml
```diff
@@ -156,7 +170,7 @@ spec:
       - curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
       - echo 'deb https://apt.kubernetes.io/ kubernetes-xenial main' > /etc/apt/sources.list.d/kubernetes.list
       - apt update -y
-      - apt install docker-ce docker-ce-cli containerd.io kubelet kubeadm kubectl -y
+      - apt install docker-ce docker-ce-cli containerd.io kubelet={{KUBERNETES_BINARIES_VERSION}} kubeadm={{KUBERNETES_BINARIES_VERSION}} kubectl={{KUBERNETES_BINARIES_VERSION}} -y
```
maelk: this should be replaced too, no ?

vm-setup/roles/v1aX_integration_test/templates/workers_centos.yaml
```diff
@@ -84,7 +84,7 @@ spec:
         - setenforce 0
         - sed -i 's/^SELINUX=enforcing$/SELINUX=permissive/' /etc/selinux/config
         - yum install docker-ce docker-ce-cli --disableexcludes=kubernetes --nobest -y
-        - yum install kubelet kubeadm kubectl --disableexcludes=kubernetes -y
+        - yum install kubelet={{KUBERNETES_BINARIES_VERSION}} kubeadm={{KUBERNETES_BINARIES_VERSION}} kubectl={{KUBERNETES_BINARIES_VERSION}} --disableexcludes=kubernetes -y
```
maelk: this should be replaced too, no ?

vm-setup/roles/v1aX_integration_test/templates/controlplane_centos.yaml
```diff
@@ -204,6 +204,19 @@ spec:
...
+        permissions: '0755'
+        content: |
+          #!/bin/bash 
+          RELEASE="${KUBERNETES_BINARIES_VERSION}"
+          # Get executables
+          curl -L --remote-name-all https://storage.googleapis.com/kubernetes-release/release/${RELEASE}/bin/linux/amd64/{kubeadm,kubelet,kubectl}
+          chmod +x {kubeadm,kubelet,kubectl}
+          # Move binaries to a folder in PATH
+          mv {kubeadm,kubelet,kubectl} /usr/local/bin
+          curl -sSL "https://raw.githubusercontent.com/kubernetes/kubernetes/${RELEASE}/build/debs/kubelet.service" | sed "s:/usr/bin:/opt/bin:g" > /etc/systemd/system/kubelet.service
```
maelk: kubelet is not copied to /opt/bin, you copied it to /usr/local/bin. this issue is in all files

vm-setup/roles/v1aX_integration_test/templates/controlplane_centos.yaml
```diff
@@ -204,6 +204,19 @@ spec:
...
+          mkdir -p /etc/systemd/system/kubelet.service.d
+          curl -sSL "https://raw.githubusercontent.com/kubernetes/kubernetes/${RELEASE}/build/debs/10-kubeadm.conf" | sed "s:/usr/bin:/opt/bin:g" > /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
```
maelk: same here

vm-setup/roles/v1aX_integration_test/templates/controlplane_ubuntu.yaml
Xenwar: indeed. 
I have replaced all on my text editor. 
Thanks, will check again.

vm-setup/roles/v1aX_integration_test/templates/controlplane_centos.yaml
Xenwar: It should not make a difference as long as the binaries are in the path. 
the remove /opt/bin on purpose.

maelk: it does make a difference. if you open https://raw.githubusercontent.com/kubernetes/kubernetes/v1.17.3/build/debs/kubelet.service, you will get this content 
```
[Unit]
Description=kubelet: The Kubernetes Node Agent
Documentation=http://kubernetes.io/docs/

[Service]
ExecStart=/usr/bin/kubelet
Restart=always
StartLimitInterval=0
RestartSec=10

[Install]
WantedBy=multi-user.target
```
by running this sed command, you replace /usr/bin/kubelet by /opt/bin/kubelet. whether kubelet is in your path or not, systemd will look for /opt/bin/kubelet. that does not exist, hence the CI failure

maelk: your sed command is a search and replace. replace `/usr/bin` by `/opt/bin`

Xenwar: Good point. 
But, the CI should not reach that level yet. There were two other issues that would prevent it reaching it there. Working on those.

Xenwar:
> ok for me, but I think we should rather go with K8S 1.18.0 now that the tests are passing

Thanks. 
I agree. Forgot changing it as described in the description.

Xenwar:
please do wait for test result with 1.18.0 (current code) before giving lgtm or approval

Xenwar:
The test with v1.18.0 is failing due to missing files, unlike previous releases. 
An issue is created in kuberentes, https://github.com/kubernetes/kubernetes/issues/89757. 

I suggest we hold this until the issue is resolved.

PR #278, opened 2020-03-31
Fix target nodes networking setup

### LINKED ISSUE #276: Ubuntu target nodes networking setup fails
When deploying the Ubuntu target cluster, the networking setup done with netplan fails with the following error :
```
/etc/netplan/60-ironicendpoint.yaml:6:20: Error in network definition: ironicendpoint: interface 'enp1s0' is not defined
      interfaces: [enp1s0]
```
/assign @kashifest 
/kind bug
/priority important-soon

kashifest:
Fixes #276

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/templates/controlplane_ubuntu.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/controlplane_ubuntu.yaml
@@ -87,7 +87,7 @@ spec:
                   {{ CLUSTER_APIENDPOINT_HOST }}
               }
           }
-      - path: /etc/netplan/50-cloud-init.yaml
+      - path: /etc/netplan/51-cloud-config.yaml
         owner: root:root
         permissions: '0644'
         content: |
@@ -95,12 +95,14 @@ spec:
               ethernets:
                   enp2s0:
                       dhcp4: true
+                  enp1s0:
+                      dhcp4: false
               version: 2
       - path: /tmp/akeys
         owner: root:root
         permissions: '0644'
         content: {{ SSH_PUB_KEY_CONTENT }}
-      - path : /etc/netplan/60-ironicendpoint.yaml
+      - path : /etc/netplan/52-ironicendpoint.yaml
         owner: root:root
         permissions: '0644'
         content: |
@@ -188,7 +190,7 @@ spec:
                     {{ CLUSTER_APIENDPOINT_HOST }}
                 }
             }
-        - path: /etc/netplan/50-cloud-init.yaml
+        - path: /etc/netplan/51-cloud-config.yaml
           owner: root:root
           permissions: '0644'
           content: |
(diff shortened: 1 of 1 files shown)


kashifest:
> Don't you need the same changes in [here ](https://github.com/metal3-io/metal3-dev-env/blob/577161a49426a5593a8caf6f1daf0b516e3d4bcd/vm-setup/roles/v1aX_integration_test/templates/controlplane_ubuntu.yaml#L196)as well?

Indeed, changed accordingly

vm-setup/roles/v1aX_integration_test/templates/controlplane_ubuntu.yaml
```diff
@@ -95,6 +95,8 @@ spec:
               ethernets:
                   enp2s0:
                       dhcp4: true
+                  enp1s0:
+                      dhcp4: false
```
maelk: I would rather add it in the 60, so that it is with the bridge. 50- comes from cloud-init normally, and might be overwritten

Xenwar:
test-centos-integration

vm-setup/roles/v1aX_integration_test/templates/controlplane_ubuntu.yaml
kashifest: Sorry, I did not get you, what do you mean by  50- comes from cloud-init normally, and might be overwritten? I think the naming is just to tell which file to be applied before. In our case, 50 would be applied before 60. I don't think 50- would be overwritten automatically, unless we do some changes explicitly on the interface in cloud-init

maelk: I meant that the 50-cloud-init.yaml file is normally generated by cloud-init.

maelk: so creating another file called the same way should be done only if you want to overwrite the cloud-init output. And there might be unexpected cases where cloud-init could regenerate that file and overwrite your changes

kashifest: That holds true for the whole file then, not this change I suppose

kashifest: I have changed the file names @maelk

maelk: how does the numbering apply ? because we want this configuration to have a higher priority than the 50-cloud-init.yaml one

kashifest: I just found these in netplan 
"Configuration is a merge of all /etc/netplan/\*.yaml. The netplan config show command will read the current on-disk configuration and merge any configs in /etc/netplan/\*.yaml merge them and produce a YAML formatted output to stdout."

kashifest: netplan’s configuration files use the YAML format. All /{lib,etc,run}/netplan/*.yaml are considered. Lexicographically later files (regardless of in which directory they are) amend (new mapping keys) or override (same mapping keys) previous ones. A file in /run/netplan completely shadows a file with same name in /etc/netplan, and a file in either of those directories shadows a file with the same name in /lib/netplan.

kashifest: changed it to 51 and 52 just to be on the safe side

maelk: good. Maybe we could later on actually merge those two.

PR #279, opened 2020-03-31
Remove nodes_file on tear down

derekhiggins:
And only create a new one if it doesn't exist. This to ensure
we don't create a new one without also defining new VM's.

Also only run teardown if vm_platform == "libvirt" to avoid
removing a user defined nodes_file.


derekhiggins:
/hold
Should be merged after
https://github.com/openshift-metal3/dev-scripts/pull/998
Fixes: https://github.com/openshift-metal3/dev-scripts/issues/932

PR #280, opened 2020-03-31
Add optional dual-stack baremetal network support

russellb:
metal3-dev-env previously included support for an IPv4 or an IPv6
baremetal libvirt network.  This patch adds the ability to also have
dual-stack (IPv4+IPv6).

There is some impact to the configuration in this change.  IPv4-only
is the default.  To use IPv6 or IPv4+IPv6, you must set the new
IP_STACK configuration variable.  It can be set to "v4", "v6", or
"v4v6".

Instead of EXTERNAL_SUBNET, the configuration now supports
EXTERNAL_SUBNET_V4 and EXTERNAL_SUBNET_V6 to allow specifying IPv4 and
IPv6 subnets at the same time.

The network.xml.j2 template is changed more than strictly necessary,
because I found the nesting of conditionals and loops difficult to
follow without any indentation.

I've validated this by ensuring that the 04_verify.sh script still
completes successfully with "v4" or "v4v6" IP_STACK mode.  With either
"v6" or "v4v6", I validated that I could configure one of the virtual
bare metal hosts with DHCPv6 and successfully get a lease.

More work is needed to actually provision an IPv6 or IPv4+IPv6
kubernetes cluster, but these changes are a prerequisite.

### DIFF
--- a/lib/network.sh
+++ b/lib/network.sh
@@ -68,6 +68,13 @@ network_address dhcp_range_end "$PROVISIONING_NETWORK" 100
 
 export CLUSTER_DHCP_RANGE=${CLUSTER_DHCP_RANGE:-"$dhcp_range_start,$dhcp_range_end"}
 
+EXTERNAL_SUBNET=${EXTERNAL_SUBNET:-""}
+if [[ -n "${EXTERNAL_SUBNET}" ]]; then
+    echo "EXTERNAL_SUBNET has been removed in favor of EXTERNAL_SUBNET_V4 and EXTERNAL_NETWORK_V6."
+    echo "Please update your configuration to drop the use of EXTERNAL_SUBNET."
+    exit 1
+fi
+
 export IP_STACK=${IP_STACK:-"v4"}
 if [[ "${IP_STACK}" == "v4" ]]; then
     export EXTERNAL_SUBNET_V4=${EXTERNAL_SUBNET_V4:-"192.168.111.0/24"}
--- a/lib/network.sh
+++ b/lib/network.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-
 #
 # Get the nth address from an IPv4 or IPv6 Network
 #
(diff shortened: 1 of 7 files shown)


russellb:
/hold

need to do some more testing, and probably adapt our openshift scripts a bit

russellb:
seems to work in the tests here, and I'm pretty sure https://github.com/openshift-metal3/dev-scripts/pull/999 is all that's needed for the openshift external consumer of the ansible bits here.  CI is currently broken on the dev-scripts repo, so I won't wait for it ...

russellb:
Another quick way to check this is to try different `IP_STACK` values, and then run `sudo virsh net-dumpxml baremetal` after script `02_...sh` runs.  You should see the DHCP pool configured for v4, v6, or both.

lib/network.sh
```diff
@@ -69,6 +68,19 @@ network_address dhcp_range_end "$PROVISIONING_NETWORK" 100
 
 export CLUSTER_DHCP_RANGE=${CLUSTER_DHCP_RANGE:-"$dhcp_range_start,$dhcp_range_end"}
 
-export EXTERNAL_SUBNET=${EXTERNAL_SUBNET:-"192.168.111.0/24"}
+export IP_STACK=${IP_STACK:-"v4"}
+if [[ "${IP_STACK}" == "v4" ]]; then
+    export EXTERNAL_SUBNET_V4=${EXTERNAL_SUBNET_V4:-"192.168.111.0/24"}
```
hardys, line 73: Would it be worth still supporting EXTERNAL_SUBNET for backwards compatibility, in the default v4 case?

Something like:

```
    export EXTERNAL_SUBNET_V4=${EXTERNAL_SUBNET_V4:-${EXTERNAL_SUBNET:-"192.168.111.0/24"}}
```

Or alternatively perhaps we should fail with a helpful message if folks still have EXTERNAL_SUBNET set in their `env`?

hardys:
lgtm, just one comment/question about the impact of removing EXTERNAL_SUBNET - I guess that may be a commonly overridden variable so may be worth checking for stale config and exiting with a clear error?

lib/network.sh
russellb: I'll push a change to add a helpful message.  Previously EXTERNAL_SUBNET could be v4 or v6, so it can't be used as the default without a little bit more validation than what you suggested.  I don't think requiring a config change is too much if there's a helpful message.

Thanks for the suggestions!

PR #282, opened 2020-04-06
Replace old baremetal provider name references with metal3

Xenwar:


Xenwar:
The tests have failed due to missing quotes. 
eg
```. 
In ./01_prepare_host.sh line 102:
sudo "${CONTAINER_RUNTIME}" run -d --net host --privileged --name ipa-downloader ${POD_NAME} \
                                                                                 ^---------^ SC2086: Double quote to prevent globbing and word splitting.
```
However, other those variables were not introduced in this PR. 
We should ask why previous/other tests passed.

maelk:
/approve cancel
Those variables should not be in quotes. That breaks the functionality as it considers the content of the variable as one string if quoted. A better alternative would be to add an exception for this rule

Xenwar:
> /approve cancel
> Those variables should not be in quotes. That breaks the functionality as it considers the content of the variable as one string if quoted. A better alternative would be to add an exception for this rule

Would it also work if, somehow, escape the quotes ?

maelk:
if you escape the quotes they will appear as quotes in the docker/podman command line, making it a string also. IMO, if you want to fix this, you should add an exception for that rule.

Xenwar:
> if you escape the quotes they will appear as quotes in the docker/podman command line, making it a string also. IMO, if you want to fix this, you should add an exception for that rule.

ok. Creating a separate PR for that.

Xenwar:
A fix for the shellcheck error. https://github.com/metal3-io/metal3-dev-env/pull/283
This PR can be approved after that.

PR #283, opened 2020-04-06
Add exception for missing double quotes

Xenwar:


russellb:
Why not just add the missing quotes instead of adding exceptions?

Issue #285, opened 2020-04-07
Running metal3-dev-env removes old GOPATH


derekhiggins:
If running as root, then GOPATH is cleaned up up each time vm-setup/install-package-playbook.yml is applied, zapping away any local edits.

PR #286, opened 2020-04-07
Set go_install_clean_full

derekhiggins:
With go_install_clean=true and go_install_clean_full=false
old golang installs are cleaned up without cleaning up GOPATH

New param added to fubarhouse/ansible-role-golang here
https://github.com/fubarhouse/ansible-role-golang/pull/152

Fixed: #285

### DIFF
--- a/vm-setup/install-package-playbook.yml
+++ b/vm-setup/install-package-playbook.yml
@@ -11,3 +11,4 @@
       vars:
         go_version: 1.14
         go_install_clean: true
+        go_install_clean_full: false


vm-setup/install-package-playbook.yml
```diff
@@ -11,3 +11,4 @@
       vars:
         go_version: 1.14
         go_install_clean: true
+        go_install_clean_full: false
```
hardys, line 14: Do we need to force galaxy to upgrade the role?

```
+(./01_install_requirements.sh:18): main(): ansible-galaxy install -r vm-setup/requirements.yml
[WARNING]: - fubarhouse.golang (2.9.0) is already installed - use --force to change version to
master
```

PR #289, opened 2020-04-08
Install python3 in CentOS target cluster and use dnf

maelk:
The CentOS verifications are failing for CentOS target cluster since python3 is not installed. This fixes it.

/cc @fmuyassarov 
/cc @Xenwar


maelk:
Skipping the integration tests are they are currently not testing those files.

PR #292, opened 2020-04-09
Remove v1alpha1 and v1alpha2 versions of CAPM3

maelk:
This PR removes the v1alpha1 and v1alpha2 specific code since those two versions were deprecated.


maelk:
From the Openshift point of view, is this change problematic ? 
/cc @russellb 
/cc @hardys 
/cc @stbenjam 
and anyone else who would want to voice their opinion :)

hardys:
> From the Openshift point of view, is this change problematic ?
> /cc @russellb
> /cc @hardys
> /cc @stbenjam
> and anyone else who would want to voice their opinion :)

I think it should be fine - the https://github.com/openshift-metal3/dev-scripts only uses the vm-setup roles that create the VMs so I don't think we'll be impacted by this change.

That said we can run CI to be sure when https://github.com/openshift-metal3/dev-scripts/pull/1013 lands (currently we're pinned to a commit before https://github.com/metal3-io/metal3-dev-env/pull/280)

fmuyassarov:
Just a small nit:
1. I think we won't need the following in the Makefile?
```
test_v1a1:
	./05_test_v1alpha1.sh

test_v1a2:
	./05_test.sh

test_v1a3:
	./05_test.sh
```

2. Shouldn't it be CentOS 8 instead of CentOS 7 in README?

maelk:
Thanks, I'll take care of this on Tuesday

maelk:
@fmuyassarov fixed, thank you

fmuyassarov:
> > From the Openshift point of view, is this change problematic ?
> > /cc @russellb
> > /cc @hardys
> > /cc @stbenjam
> > and anyone else who would want to voice their opinion :)
> 
> I think it should be fine - the https://github.com/openshift-metal3/dev-scripts only uses the vm-setup roles that create the VMs so I don't think dev-scripts will be impacted by this change.
> 
> That said we can run CI to be sure when [openshift-metal3/dev-scripts#1013](https://github.com/openshift-metal3/dev-scripts/pull/1013) lands (currently dev-scripts is pinned to a commit before #280)

@hardys 
Can we merge this PR? Because I see from comments in https://github.com/openshift-metal3/dev-scripts/pull/1017 that the CI has passed.

hardys:
> @hardys
> Can we merge this PR? Because I see from comments in [openshift-metal3/dev-scripts#1017](https://github.com/openshift-metal3/dev-scripts/pull/1017) that the CI has passed.

That doesn't prove it works with this PR, I'm manually testing to confirm and we can merge if all is OK, sorry for the delay responding.

PR #293, opened 2020-04-14
Kind access to local Docker registry made possible by using containerd config

mikkosest:


mikkosest:
> This could be just a firewall issue, please try to add the rule we discussed about and let's see if this is actually needed.

Should third rule below allow access, still does not work without kind config, why not to use kind config, which is given as example in kind page?

ACCEPT     all  --  anywhere             192.168.122.0/24     ctstate RELATED,ESTABLISHED
ACCEPT     all  --  192.168.122.0/24     anywhere            
ACCEPT     all  --  anywhere             anywhere

03_launch_mgmt_cluster.sh
```diff
@@ -161,6 +161,19 @@ function launch_baremetal_operator() {
...
 
+function launch_kind() {
+  reg_ip="$(docker inspect -f '{{.NetworkSettings.IPAddress}}' registry)"
+  cat <<EOF | kind create cluster --name kind --image=kindest/node:"${KUBERNETES_VERSION}" --config=-
+  kind: Cluster
+  apiVersion: kind.x-k8s.io/v1alpha4
+  containerdConfigPatches:
+  - |-
+    [plugins."io.containerd.grpc.v1.cri".registry.mirrors."192.168.111.1:5000"]
+      endpoint = ["http://${reg_ip}:5000"]
```
maelk, line 172: Is there really a need to go with the Docker IP instead of the 192.168.111.1 here ?

maelk:
ok, let's go with this then.

PR #297, opened 2020-04-27
Replace CAPI and CAPM3 deployment with clusterctl in metal3-dev-env

mikkosest:


03_launch_mgmt_cluster.sh
```diff
@@ -79,6 +82,38 @@ function clone_repos() {
...
+kind: Kustomization
+resources:
+- components.yaml
+EOF
+  wget https://github.com/metal3-io/cluster-api-provider-metal3/releases/download/"${CAPM3_RELEASE}"/cluster-template.yaml
+  wget https://github.com/metal3-io/cluster-api-provider-metal3/releases/download/"${CAPM3_RELEASE}"/infrastructure-components.yaml -O components.yaml
+  wget https://github.com/metal3-io/cluster-api-provider-metal3/releases/download/"${CAPM3_RELEASE}"/metadata.yaml
+  kustomize edit set image quay.io/metal3-io/cluster-api-provider-metal3:"${CAPM3_RELEASE}"=192.168.111.1:5000/localimages/cluster-api-provider-metal3:"${CAPM3_RELEASE}"
+  kustomize build . > infrastructure-components.yaml
+  rm components.yaml kustomization.yaml
```
maelk, line 115: you should not be using the released version of capm3. The goal of those tests are to test a branch. Instead of doing it this way, you should generate the release manifests by doing `make release-manifests` . Then copy what is in the `out` folder into the override folder of clusterctl.

For the image, you can override it in the clusterctl config file, you don't need an intermediary kustomization. And the image should be overridden conditionally

03_launch_mgmt_cluster.sh
```diff
@@ -93,10 +128,16 @@ function update_images(){
     # Strip the tag for image replacement
     OLD_IMAGE="${!OLD_IMAGE_VAR%:*}"
     #shellcheck disable=SC2086
-    kustomize edit set image $OLD_IMAGE=$LOCAL_IMAGE
+    if [ -n "${CAPM3_LOCAL_IMAGE}" ]; then
+      patch_clusterctl
+    else
+      kustomize edit set image $OLD_IMAGE=$LOCAL_IMAGE
+    fi
```
maelk, line 135: this would rather depend on the version, no ?

03_launch_mgmt_cluster.sh
```diff
@@ -194,29 +234,29 @@ kind: Kustomization
...
+    popd
 
     pushd "$kustomize_overlay_path"
     update_images
     popd
-    kustomize build "$kustomize_overlay_path" | kubectl apply -f-
+
+    pushd "${CAPIPATH}"
+    cd bin
+    ./clusterctl init --infrastructure=metal3
```
maelk, line 259: if you do it this way, it will download the last release of metal3. We want to acutally test the branch we are on (can be master for daily jobs) so we need to use the override feature.

vm-setup/roles/v1aX_integration_test/tasks/generate_templates.yml
```diff
@@ -0,0 +1,59 @@
...
+
+  - name: Copy kustomization skeleton
+    copy:
+      src: "{{ CRS_PATH }}/kustomization.yaml"
+      dest: "{{ TEMP_GEN_DIR }}"
+
+  - name: Map replicas values
+    template:
+      src: "{{ CRS_PATH }}/replicas.yaml"
+      dest: "{{ TEMP_GEN_DIR }}/replicas.yaml"
```
maelk, line 31: the number of replicase can be given as variables to clusterctl config. I am not sure what this whole kustomization part does.

03_launch_mgmt_cluster.sh
mikkosest: Mostly fixed, only image overwrite is a problem

03_launch_mgmt_cluster.sh
mikkosest: Changed these lines completely due to to other changes

03_launch_mgmt_cluster.sh
mikkosest: Fixed

vm-setup/roles/v1aX_integration_test/tasks/generate_templates.yml
mikkosest: Fixed. Kustomization is need to split and merge manifests when separate provision steps are cluster, controlplane, worker

03_launch_mgmt_cluster.sh
mikkosest: Completely fixed

lib/common.sh
```diff
@@ -41,6 +41,8 @@ CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-"podman"}
 # Hostname format
 NODE_HOSTNAME_FORMAT=${NODE_HOSTNAME_FORMAT:-"node-%d"}
 
+CAPM3_RELEASE=${CAPM3_RELEASE:-"v0.3.1"}
```
maelk, line 44: is there a way to retrieve this automatically from Github ? otherwise we will need to update it everytime we release a version

03_launch_mgmt_cluster.sh
```diff
@@ -29,10 +29,13 @@ M3PATH="${M3PATH:-${GOPATH}/src/github.com/metal3-io}"
 BMOPATH="${BMOPATH:-${M3PATH}/baremetal-operator}"
 RUN_LOCAL_IRONIC_SCRIPT="${BMOPATH}/tools/run_local_ironic.sh"
 CAPM3PATH="${CAPM3PATH:-${M3PATH}/cluster-api-provider-metal3}"
+CAPIPATH="${CAPIPATH:-${M3PATH}/cluster-api}"
 
 if [ "${CAPI_VERSION}" == "v1alpha3" ]; then
   CAPM3BRANCH="${CAPM3BRANCH:-release-0.3}"
   CAPM3REPO="${CAPM3REPO:-https://github.com/metal3-io/cluster-api-provider-metal3.git}"
+  CAPIBRANCH="${CAPIBRANCH:-v0.3.3}"
```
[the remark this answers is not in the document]
mikkosest: Done

lib/common.sh
mikkosest: Done

03_launch_mgmt_cluster.sh
```diff
@@ -29,13 +29,22 @@ M3PATH="${M3PATH:-${GOPATH}/src/github.com/metal3-io}"
 BMOPATH="${BMOPATH:-${M3PATH}/baremetal-operator}"
 RUN_LOCAL_IRONIC_SCRIPT="${BMOPATH}/tools/run_local_ironic.sh"
 CAPM3PATH="${CAPM3PATH:-${M3PATH}/cluster-api-provider-metal3}"
+CAPM3REPO="${CAPM3REPO:-https://github.com/metal3-io/cluster-api-provider-metal3.git}"
```
maelk, line 32: you do not have to add the .git at the end. If you do not provide it, then you can use CAPM3REPO as the base url for CAPM3RELEASEPATH

03_launch_mgmt_cluster.sh
```diff
@@ -29,13 +29,22 @@ M3PATH="${M3PATH:-${GOPATH}/src/github.com/metal3-io}"
...
+CAPM3RELEASEPATH="${CAPM3RELEASEPATH:-https://github.com/metal3-io/cluster-api-provider-metal3/releases/latest}"
+CAPM3_RELEASE=$(curl -ksL "${CAPM3RELEASEPATH}" \
+                   | grep -m 1 "releases/download" \
+                   | cut -d \" -f 2 | tr -d \" \
```
maelk, line 36: if you use jq, you can reduce this to a single command after a pipe

03_launch_mgmt_cluster.sh
```diff
@@ -29,13 +29,22 @@ M3PATH="${M3PATH:-${GOPATH}/src/github.com/metal3-io}"
...
+                   | sed 's/.*releases\/download\///' \
+                   | sed 's/\/.*//')
+
+CAPIPATH="${CAPIPATH:-${M3PATH}/cluster-api}"
 
 if [ "${CAPI_VERSION}" == "v1alpha3" ]; then
   CAPM3BRANCH="${CAPM3BRANCH:-release-0.3}"
-  CAPM3REPO="${CAPM3REPO:-https://github.com/metal3-io/cluster-api-provider-metal3.git}"
+  CAPIBRANCH="${CAPIBRANCH:-v0.3.4}"
```
maelk, line 44: you could do the same for the release as for CAPM3

mikkosest: Done

03_launch_mgmt_cluster.sh
mikkosest: Done

03_launch_mgmt_cluster.sh
mikkosest: Changed completely url parsing

mikkosest:
> Both for CAPI and CAPM3, I think we should use the latest release, and fetch it automatically. It can be done by querying a url and using jq to get the tag name : https://developer.github.com/v3/repos/releases/#get-the-latest-release

Done

PR #299, opened 2020-04-28
Add missing privilege to run docker and kind commands

Xenwar:
Add missing privilege to run docker and kind commands

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -154,8 +154,8 @@ function launch_baremetal_operator() {
 }
 
 function launch_kind() {
-  reg_ip="$(docker inspect -f '{{.NetworkSettings.IPAddress}}' registry)"
-  cat <<EOF | kind create cluster --name kind --image=kindest/node:"${KUBERNETES_VERSION}" --config=-
+  reg_ip="$(sudo docker inspect -f '{{.NetworkSettings.IPAddress}}' registry)"
+  cat <<EOF | sudo su -l -c "kind create cluster --name kind --image=kindest/node:${KUBERNETES_VERSION} --config=- " "$USER"
   kind: Cluster
   apiVersion: kind.x-k8s.io/v1alpha4
   containerdConfigPatches:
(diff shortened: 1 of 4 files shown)


03_launch_mgmt_cluster.sh
```diff
@@ -154,7 +154,7 @@ function launch_baremetal_operator() {
 }
 
 function launch_kind() {
-  reg_ip="$(docker inspect -f '{{.NetworkSettings.IPAddress}}' registry)"
+  reg_ip="$(sudo docker inspect -f '{{.NetworkSettings.IPAddress}}' registry)"
```
maelk, line 157: maybe we should make a deeper fix, as this will only result in hitting further problems. We probably should add the current user to the docker group and use `sudo su -l -c "command to run" "$USER"` wherever we are running kind commands to make sure that our user can use the docker socket.

Xenwar: Thanks. Working on it .

Xenwar: Done

Xenwar:
> I think there is one more place we forgot, that needs the same `sudo su -l ..`
> 
> https://github.com/metal3-io/metal3-dev-env/blob/b9425bdc9b13268956601a4689e78d65eb99374d/cluster_cleanup.sh#L9

Done

ubuntu_install_requirements.sh
```diff
@@ -69,6 +69,7 @@ EOF
   sudo mkdir -p /etc/docker
   sudo mv daemon.json /etc/docker/daemon.json
   sudo systemctl restart docker
+  sudo usermod -aG docker "${USER}"
```
maelk, line 72: there is also a docker install in the centos requirements. Can you add it there too ?

fmuyassarov: but we don't use docker in Centos, why would we need it?

Xenwar: Ok. Working on it.

Xenwar: Done

PR #301, opened 2020-04-30
:bug: CAPM3BRANCH type wrongly mixed with CAPM3 release version

mikkosest:


03_launch_mgmt_cluster.sh
```diff
@@ -41,7 +41,8 @@ export CAPIBRANCH="v0.3.3"
 CAPIPATH="${CAPIPATH:-${M3PATH}/cluster-api}"
 CAPM3RELEASEPATH="${CAPM3RELEASEPATH:-https://api.github.com/repos/${CAPM3_BASE_URL}/releases/latest}"
 CAPIRELEASEPATH="${CAPIRELEASEPATH:-https://api.github.com/repos/${CAPI_BASE_URL}/releases/latest}"
-CAPM3BRANCH=${CAPM3BRANCH:-$(get_latest_release "${CAPM3RELEASEPATH}")}
+CAPM3RELEASE=${CAPM3RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}")}
+CAPM3BRANCH=${CAPM3BRANCH:-${CAPM3RELEASE}}
 CAPIBRANCH=${CAPIBRANCH:-$(get_latest_release "${CAPIRELEASEPATH}")}
```
maelk, line 46: This is not consistent, it might need renaming. you introduced CAPM3RELEASE but you do not have CAPIRELEASE just CAPIBRANCH. since CAPIBRANCH contains the release, I would rather name it CAPIRELEASE

mikkosest: CAPIBRANCH is used for branching, now this branch is same as release but are you sure that we do not want to test with CAPI git hash ever i.e. taking CAPI version from master branch?

maelk: I would actually rather go with testing with master branch of CAPI as much as possible, not from released since that allows us to catch problems in CAPI early

mikkosest: I changed CAPIBRANCH to master by default

mikkosest: That will just cause that problem, that CAPI can be broken the way that it breaks our CI and it causes long CI stop-the-lines if CAPI does not react fast

maelk: in that case we can switch to a release branch if needed. it would allow us to find the bug and report it to CAPI, that we would not see it otherwise

PR #302, opened 2020-04-30
:bug: CAPM3 local image registry path faulty

mikkosest:


03_launch_mgmt_cluster.sh
```diff
@@ -104,8 +104,8 @@ function clone_repos() {
 function patch_clusterctl(){
   pushd "${CAPM3PATH}"
   if [ -n "${CAPM3_LOCAL_IMAGE}" ]; then
-    export MANIFEST_IMG="192.168.111.1:5000/localimages/cluster-api-provider-metal3"
-    export MANIFEST_TAG="${CAPM3RELEASE}"
+    export MANIFEST_IMG="192.168.111.1:5000/localimages/tested_repo"
```
maelk, line 107: Isn't that the CAPM3_LOCAL_IMAGE ?

maelk: Then you would be better off using something like
```
IMAGE_NAME="${CAPM3_LOCAL_IMAGE##*/}:latest"
LOCAL_IMAGE="192.168.111.1:5000/localimages/$IMAGE_NAME"
```

mikkosest: done

PR #305, opened 2020-05-07
Fix exporting the image variable after override

maelk:
they are currrently not properly exported, so the override is ineffective for run_local_ironic.sh script

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -131,7 +131,7 @@ function update_images(){
       kustomize edit set image $OLD_IMAGE=$LOCAL_IMAGE
     fi
     declare "$OLD_IMAGE_VAR"="$LOCAL_IMAGE"
-    export OLD_IMAGE_VAR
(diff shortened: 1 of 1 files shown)


maelk:
This cannot be tested as the CI is broken by ironic-image, and this is needed to run the tests for the ironic fix

03_launch_mgmt_cluster.sh
```diff
@@ -131,7 +131,7 @@ function update_images(){
       kustomize edit set image $OLD_IMAGE=$LOCAL_IMAGE
     fi
     declare "$OLD_IMAGE_VAR"="$LOCAL_IMAGE"
-    export OLD_IMAGE_VAR
+    export "$OLD_IMAGE_VAR"
```
fmuyassarov, line 134: I think you need {} to pass the shellcheck

maelk: it's actually the ? that matters to silence the shellcheck. I have modified it

PR #306, opened 2020-05-07
Fix the local image variable setting

maelk:
using eval instead of declare for the variable to be
set outside of the function


maelk:
This cannot go through testing as it is required to be able to test the fix for the ironic image that broke the CI. I'll merge it directly.

PR #310, opened 2020-05-08
Clean up ironic-endpoint-keepalived container during make clean

fmuyassarov:
Delete the `ironic-endpoint-keepalived` container during the host cleanup
Related to this [#511](https://github.com/metal3-io/baremetal-operator/pull/511)

### DIFF
--- a/cluster_cleanup.sh
+++ b/cluster_cleanup.sh
@@ -12,4 +12,3 @@ fi
 if [ "${EPHEMERAL_CLUSTER}" == "minikube" ]; then
   sudo su -l -c "minikube delete" "${USER}"
 fi
-
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -7,7 +7,7 @@ source lib/logging.sh
 source lib/common.sh
 
 # Kill and remove the running ironic containers
-for name in ipa-downloader ironic ironic-inspector dnsmasq httpd mariadb vbmc sushy-tools httpd-infra; do
+for name in ipa-downloader ironic ironic-inspector ironic-endpoint-keepalived dnsmasq httpd mariadb vbmc sushy-tools httpd-infra; do
     sudo "${CONTAINER_RUNTIME}" ps | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" kill $name
     sudo "${CONTAINER_RUNTIME}" ps --all | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" rm $name -f
 done
@@ -61,4 +61,3 @@ if [[ $OS == "centos" || $OS == "rhel" ]]; then
 fi
 
 rm -rf  "${HOME}"/.cluster-api
-
--- a/vm-setup/roles/libvirt/tasks/network_teardown_tasks.yml
+++ b/vm-setup/roles/libvirt/tasks/network_teardown_tasks.yml
@@ -16,7 +16,7 @@
 
 - name: Delete bridges on Ubuntu
   shell: |
-     sudo ip link set baremetal down 
+     sudo ip link set baremetal down
      sudo ip link set provisioning down
      brctl delbr baremetal | true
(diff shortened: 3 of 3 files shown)


fmuyassarov:
> This would be incorrect when ironic goes back into the cluster. I think you need to add some conditions to set the IP properly depending on the scenario. And if something needs to be changed / reverted after the ironic fix, please mention it explicitely in a comment

Updated.

fmuyassarov:
Related to this [#511](https://github.com/metal3-io/baremetal-operator/pull/511)

03_launch_mgmt_cluster.sh
```diff
@@ -293,6 +293,10 @@ create_clouds_yaml
 
 if [ "${EPHEMERAL_CLUSTER}" == "kind" ]; then
   launch_kind
+  sudo brctl addbr "${CLUSTER_PROVISIONING_INTERFACE}"
+  sudo ip link set "${CLUSTER_PROVISIONING_INTERFACE}" up
+  sudo ip addr add "${INITIAL_IRONICBRIDGE_IP/$PROVISIONING_CIDR}" dev "${CLUSTER_PROVISIONING_INTERFACE}"
```
maelk, line 298: that bridge is not connected to anything ?

fmuyassarov: Right now no, but it will be used to serve Ironic endpoint

maelk: if it is used to serve Ironic endpoint but is not connected to anything, how will clients talk to ironic ? (like BMO etc.)

fmuyassarov: Sorry if I wasn't clear. What I mean is that, after when run the ironic-endpoint-keepalived as a docker container, then the Ironic will be connected to this bridge based on [this ](https://github.com/metal3-io/baremetal-operator/blob/9c21a13425efcc475b220264c6f816a38b3d6d33/deploy/ironic-keepalived-config/ironic_bmo_configmap.env#L8)configuration. We have to create a bridge here and assign 172.22.0.9 IP before Ironic starts serving at 172.22.0.2 because if there is nothing on that bridge that Ironic can connect to it. We have the similar with [Minikube ](https://github.com/metal3-io/metal3-dev-env/blob/0f21decd48128ba1df52385ec2f4edae2462a593/03_launch_mgmt_cluster.sh#L307).

maelk: Sorry, I was not clear. How is the traffic from the VMs (arriving in the provisioning bridge) going to reach ironic (this bridge). Considering that the IP would be on another interface than provisioning bridge without a link (like a veth) between the two, I'm not sure the ARP queries would be answered, that would make it so that the nodes and ironic are not able to communicate. Also, if you apply .9 on that bridge, when you have .1 on the provisioning bridge, you have two addresses (in addition to the VIP) on that network on that node. You could just use the .1, you don't need the .9, if it is at the correct interface

fmuyassarov: @maelk thanks for clarification.
Made some updates as discussed. Now, this changes clean up the ironic-endpoint-keepalived container after `make clean` and replaces `CLUSTER_URL_HOST` ( == 172.22.0.2) with `PROVISIONING_URL_HOST` ( == 172.22.0.1), since `IRONIC_ENDPOINT` == 172.22.0.1 in [ironic_ci.env](https://github.com/metal3-io/baremetal-operator/blob/71ebf501f631927842cfa8d3248754acf9ad8d5c/deploy/ironic_ci.env#L6).

fmuyassarov:
> This uses the .1 address everywhere, while we want to use the .2 (keepalived) instead. you need to use the CLUSTER_URL_HOST rather

fixed

fmuyassarov:
> This PR does not do what its title says. It will still not fix the clouds.yaml generation (still with .1) as far as I can say

Updated, thanks.

Issue #298, opened 2020-04-28
Target cluster endpoint is unreachable when we do make clean and make again due to firewall rules


kashifest:
When we do a `make clean` and `make` again and try to do the provisioning of the control-plane again we see the following error in logs of CAPI and CAKCP:
```
"failed to create remote cluster client: failed to create client for workload cluster metal3/test1: Get https://192.168.111.249:6443/api?timeout=30s: dial tcp 192.168.111.249:6443: connect: connection refused" "cluster"="test1" "kubeadmControlPlane"="test1-controlplane" "namespace"="metal3" 
```
This particularly happens when we have Kind as the ephemeral cluster and we have done a `make clean` and `make` again. The iptable rules are as follows:

```
Chain FORWARD (policy DROP 14 packets, 560 bytes)
 pkts bytes target     prot opt in     out     source               destination         
 272K 3441M ACCEPT     all  --  *      baremetal  0.0.0.0/0            192.168.111.0/24     ctstate RELATED,ESTABLISHED
 106K 6508K ACCEPT     all  --  baremetal *       192.168.111.0/24     0.0.0.0/0           
   16  5236 ACCEPT     all  --  baremetal baremetal  0.0.0.0/0            0.0.0.0/0           
 9177  551K REJECT     all  --  *      baremetal  0.0.0.0/0            0.0.0.0/0            reject-with icmp-port-unreachable
    0     0 REJECT     all  --  baremetal *       0.0.0.0/0            0.0.0.0/0            reject-with icmp-port-unreachable
    0     0 ACCEPT     all  --  *      virbr0  0.0.0.0/0            192.168.122.0/24     ctstate RELATED,ESTABLISHED
    0     0 ACCEPT     all  --  virbr0 *       192.168.122.0/24     0.0.0.0/0           
    0     0 ACCEPT     all  --  virbr0 virbr0  0.0.0.0/0            0.0.0.0/0           
    0     0 REJECT     all  --  *      virbr0  0.0.0.0/0            0.0.0.0/0            reject-with icmp-port-unreachable
    0     0 REJECT     all  --  virbr0 *       0.0.0.0/0            0.0.0.0/0            reject-with icmp-port-unreachable
 108K 1068M DOCKER-USER  all  --  *      *       0.0.0.0/0            0.0.0.0/0           
 108K 1068M DOCKER-ISOLATION-STAGE-1  all  --  *      *       0.0.0.0/0            0.0.0.0/0           
[...]

PR #311, opened 2020-05-11
Add firewall rules state "absent" during make clean in Metal3-dev-env 

smoshiur1237:
#298
This issue is related to the iptables rules in Metal3-dev-env. After we do `make clean`, the iptables rules are still active in the firewall. So here we keep the state absent for each iptables rule during the `make clean` process.

### DIFF
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -46,6 +46,14 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -i vm-setup/inventory.ini \
     -b -vvv vm-setup/teardown-playbook.yml
 
+if [ "$USE_FIREWALLD" == "False" ]; then
+ ANSIBLE_FORCE_COLOR=true ansible-playbook \
+    -e "{use_firewalld: $USE_FIREWALLD}" \
+    -e "firewall_rule_state=absent" \
+    -i vm-setup/inventory.ini \
+    -b -vvv vm-setup/firewall.yml
+fi
+
 # There was a bug in this file, it may need to be recreated.
 if [[ $OS == "centos" || $OS == "rhel" ]]; then
   sudo rm -rf /etc/NetworkManager/conf.d/dnsmasq.conf
--- a/vm-setup/roles/firewall/tasks/iptables.yaml
+++ b/vm-setup/roles/firewall/tasks/iptables.yaml
@@ -13,6 +13,7 @@
     match: udp
     destination_port: "{{ vbmc_port_range }}"
     jump: ACCEPT
+    state: "{{ firewall_rule_state }}"
 
 - name: "iptables: Established and related"
   iptables:
@@ -21,6 +22,7 @@
     match: conntrack
     ctstate: ESTABLISHED,RELATED
     jump: ACCEPT
+    state: "{{ firewall_rule_state }}"
 
 - name: "iptables: Ironic Ports"
   iptables:
@@ -30,6 +32,7 @@
     match: tcp
     destination_port: "{{ item }}"
     jump: ACCEPT
+    state: "{{ firewall_rule_state }}"
   loop: "{{ ironic_ports }}"
 
 - name: "iptables: Provisioning host ports"
@@ -40,6 +43,7 @@
     match: tcp
     destination_port: "{{ item }}"
     jump: ACCEPT
+    state: "{{ firewall_rule_state }}"
   loop: "{{ provisioning_host_ports }}"
 
 - name: "iptables: PXE Ports"
@@ -50,6 +54,7 @@
     match: udp
     destination_port: "{{ item }}"
     jump: ACCEPT
+    state: "{{ firewall_rule_state }}"
   loop: "{{ pxe_udp_ports }}"
 
 - name: "iptables: Ironic Endpoint Keepalived"
@@ -58,6 +63,7 @@
     in_interface: "{{ provisioning_interface }}"
     protocol: "{{ item }}"
     jump: ACCEPT
+    state: "{{ firewall_rule_state }}"
   loop: "{{ ironic_keepalived_proto }}"
 
 - name: "iptables: Allow access to baremetal network from kind"
@@ -67,4 +73,5 @@
     out_interface: "{{ baremetal_interface }}"
     destination: "192.168.111.1/24"
     jump: ACCEPT
+    state: "{{ firewall_rule_state }}"
   when: (EPHEMERAL_CLUSTER == "kind")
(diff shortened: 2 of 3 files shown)


vm-setup/roles/libvirt/tasks/network_teardown_tasks.yml
```diff
@@ -23,6 +23,12 @@
 
   when:
     - ansible_distribution == 'Ubuntu'
-
   become: yes
 
+- name: Disable the firewall forward rule
+  iptables:
+    chain: FORWARD
```
maelk, line 30: This removes all rules on the forward chain. It is too strong in the way that you would be removing rules that are not added by Metal3-dev-env. You should selectively remove the rules. An alternative approach could be to parameterize the firewall role, so that the state is present when you do make, and the state is absent when you do make clean.

vm-setup/roles/libvirt/tasks/network_teardown_tasks.yml
```diff
@@ -23,6 +23,12 @@
...
+    state: absent
+  when:
+    - EPHEMERAL_CLUSTER == 'kind'
```
maelk, line 33: I don't think we should restrict the solution to Kind only

kashifest:
I agree with @maelk's comments

vm-setup/roles/libvirt/tasks/network_teardown_tasks.yml
smoshiur1237: As this reject rule is appearing when you have kind cluster, I used this condition.  do you think that this condition doesn't need at all.

kashifest: The problem was kind specific but we can make the solution to work both for kind and minikube.  So you can just remove the lines 32-33

smoshiur1237: In that case I don't need to use any condition.

kashifest: Yes. We can remove it.

vm-setup/roles/libvirt/tasks/network_teardown_tasks.yml
smoshiur1237: Done

vm-setup/roles/libvirt/tasks/network_teardown_tasks.yml
smoshiur1237: Done

vm-setup/roles/libvirt/tasks/network_teardown_tasks.yml
smoshiur1237: I tried to find a connection with firewall rules creation during we call the setup playbook and firewall deletion with tear down playbook file. Here during the setup process the firewall rules are generated with separate call in 02_configure_host.sh, it doesn't have have any link with the setup playbook ansiable file.  So if I take this path, need to modify different setting and eventually will be complicated to follow.  That is the reason I take the 2nd path with tear down process where I generate a new file named firewall_teardown_tasks and from here we can delete each firewalls rules generated by metal3-dev-env. The testing works properly when I do `make clean` and `make`command.  After make command all the rules are back in the iptables output..

vm-setup/roles/libvirt/defaults/main.yml
```diff
@@ -16,6 +16,42 @@ libvirt_arch: x86_64
...
+  - "68"
+  # DHCPv6
+  - "546"
+  - "547"
+  # TFTP
+  - "69"
+ironic_keepalived_proto:
+  - "112"
+  - "icmp"
+
```
kashifest, line 54: I am not so in-favor of keeping the same info in two places. That would require back-porting each and every change and its difficult to manage. Is it possible to define this in one place and then let firewall and libvirt tasks source it?

smoshiur1237: I do agree.  I can take a look on this issue..

maelk:
This approach creates a lot of duplication when we could do only with an additional variable in an existing playbook. You could set the state to be present by default so that you do not need to modify the 02_* script but then call it with absent in host_cleanup.sh script

vm-setup/roles/libvirt/tasks/network_teardown_tasks.yml
smoshiur1237: updated

host_cleanup.sh
```diff
@@ -46,6 +46,12 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -i vm-setup/inventory.ini \
     -b -vvv vm-setup/teardown-playbook.yml
 
+ANSIBLE_FORCE_COLOR=true ansible-playbook \
+    -e "{use_firewalld: $USE_FIREWALLD}" \
```
fmuyassarov, line 50: Is the intention here to lookup for` USE_FIREWALLD` variable that is set on the current shell session and inherit its value to `use_firewalld`? If yes, I think you don't need {}. See the example [here](https://github.com/metal3-io/metal3-dev-env/blob/0f21decd48128ba1df52385ec2f4edae2462a593/scripts/v1alphaX/provision_cluster.sh#L18)

vm-setup/roles/firewall/tasks/iptables.yaml
```diff
@@ -13,6 +13,7 @@
     match: udp
     destination_port: "{{ vbmc_port_range }}"
     jump: ACCEPT
+    state: "{{ firewall_rule_state }}"
```
fmuyassarov, line 16: Are you expecting `state: present` ?

fmuyassarov: I got it, so when disabling rules you pass the `absent` value and when in opposite which is enabling you just rely on the default (`present`) value.

smoshiur1237: Yes You got it

smoshiur1237:
/cc @maelk 
can you please approve.

kashifest:
> This change impacts the firewalld based deployments too. you should modify the firewalld task to achieve the same result (firewalld is by default on Centos8 deployments), it works differently though so maybe the exact same thing is not applicable, and you would need another task

Why not just use `use_firewalld: false` instead in the cleanup script? I am not sure if changin firewalld tasks falls into the scope of this bug.

maelk:
/approve
Now this does not impact the firewalld based setup. I am however wondering if there would be the same problem there. This can be dealt in another PR if it is the case.

PR #312, opened 2020-05-13
Add local registry as insecure for podman

maelk:
Running ironic locally is failing because podman is unable to pull the image from local registry. This PR sets the local registry as insecure globally

### DIFF
--- a/ubuntu_install_requirements.sh
+++ b/ubuntu_install_requirements.sh
@@ -47,6 +47,7 @@ sudo apt -y update
 
 if [[ "${CONTAINER_RUNTIME}" == "podman" ]]; then
   sudo apt -y install podman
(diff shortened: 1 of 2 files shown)


ubuntu_install_requirements.sh
```diff
@@ -47,6 +47,7 @@ sudo apt -y update
 
 if [[ "${CONTAINER_RUNTIME}" == "podman" ]]; then
   sudo apt -y install podman
+  sudo sed -i '/^\[registries\.insecure\]$/,/^\[/ s/^registries =.*/registries = ["192.168.111.1:5000"]/g' /etc/containers/registries.conf
```
kashifest, line 50: I think we have same or similar code block for docker also. Can we use the same content and parameterize the name of the file only? or is the formatting or content different?

maelk: It is different, the docker one is a couple of lines below. I have created an issue to parameterize the registry IP #316 but it needs to be dealt with separately as it touches more things

kashifest: ok

PR #314, opened 2020-05-13
Switch to pip installing virtualbmc

derekhiggins:
pyghmi 1.5.13 is needed for a cipher fix but not
yet available in RDO. Switch to pip while we wait.


derekhiggins:
/hold
not the exact patch I've been testing, will test this afternoon.

dtantsur:
I think it's a bit of an overkill, I'd rather do something like this: https://github.com/metal3-io/metal3-dev-env/pull/315

PR #315, opened 2020-05-13
vbmc: work around old pyghmi in RDO

dtantsur:
We need at least pyghmi 1.5.13 for VirtualBMC to work with newer
ipmitool in RHEL 8.2.

### DIFF
--- a/resources/vbmc/Dockerfile
+++ b/resources/vbmc/Dockerfile
@@ -1,10 +1,14 @@
 FROM docker.io/centos:centos8
 
-RUN dnf install -y python3 python3-requests && \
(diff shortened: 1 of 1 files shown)


hardys:
@maelk do we set VBMC_LOCAL_IMAGE in the integration test environment to test these kind of changes? 

I've built locally and lgtm though.

e-minguez:
Previously my VMs took ~2 minutes to be powered on, meaning provision issues. I've tested this locally and the VM is up almost inmediately.

derekhiggins:
/lgtm
works for me

PR #320, opened 2020-05-18
Use raw image streaming when deploying

maelk:
We are hitting memory limits when deploying qcow2 image. This PR attempts at using raw image to leverage the raw image streaming capability of IPA.


maelk:
https://cloud.centos.org seems to be down. that's why the centos job fails

hardys:
I don't think this will impact the openshift testing, but I wonder if it's necessary to change the IMAGE_NAME variable?  Not blocking on it, but that's going to break anyone that's specifying a custom image in their config, right?

maelk:
I had the reverse in mind. the users can keep giving their image as qcow2 images in IMAGE_NAME, they should not have to know that under the hood we convert it to raw before deploying. So keeping IMAGE_NAME with the original content preserves this behaviour. they can still use the IMAGE_NAME if they want to in their script, they will just not benefit from image streaming if it is not raw. How does that sound to you @hardys ?

maelk:
/test-integration
The first one did not start properly

maelk:
regex test
/test-integration

maelk:
/test-centos-integration
trigger testing

maelk:
/test-centos-integration
trigger testing

maelk:
trigger testing
/test-centos-integration
/test-integration

maelk:
/test-integration
trigger testing

maelk:
/test-integration
trigger testing

maelk:
/test-integration
abc

maelk:
/test-integration
trigger testing2

maelk:
/test-integration
trigger testing

maelk:
/test-centos-integration
We have a new ironic image with the latest code that should contain the fix

maelk:
This passes in v1a4. Once https://github.com/metal3-io/cluster-api-provider-metal3/pull/102 is merged, this should also pass the v1a3 tests

PR #321, opened 2020-05-18
Remove container images pre-pull

maelk:
Since we do not pre-bake the image anymore for CI, this pre-pull
is unnecessary and consumes a lot of time downloading twice
some images that are not even used if we use the one built locally

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -334,12 +334,6 @@ function init_minikube() {
     #If the vm exists, it has already been initialized
     if [[ "$(sudo virsh list --all)" != *"minikube"* ]]; then
       sudo su -l -c "minikube start --insecure-registry 192.168.111.1:5000" "$USER"
-      # Pre-pull the image to reduce pod initialization time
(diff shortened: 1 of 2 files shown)


lib/common.sh
```diff
@@ -334,12 +334,6 @@ function init_minikube() {
     #If the vm exists, it has already been initialized
     if [[ "$(sudo virsh list --all)" != *"minikube"* ]]; then
       sudo su -l -c "minikube start --insecure-registry 192.168.111.1:5000" "$USER"
-      # Pre-pull the image to reduce pod initialization time
-      for IMAGE_VAR in IRONIC_IMAGE IPA_DOWNLOADER_IMAGE IRONIC_INSPECTOR_IMAGE BAREMETAL_OPERATOR_IMAGE; do
-        IMAGE=${!IMAGE_VAR}
-        sudo su -l -c "minikube ssh sudo docker pull $IMAGE" "${USER}"
-      done
-      sudo su -l -c "minikube ssh sudo docker image ls" "${USER}"
```
kashifest, line 342: This is for minikube, was this also done twice? or minikube will fetch the local images?

maelk: This was done both in metal3-dev-env and in minikube, both fetching the image from quay.io

PR #329, opened 2020-05-28
Add Ansible script for remediation test

fmuyassarov:
This patch adds Ansible script for testing Reboot API operations. Currently, we try to
1. Reboot a worker node
2. Power off a worker node
3. Power on a worker node

Once re-introspection and remediation controller are implemented, we will add more tests for those use cases.


vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
```diff
@@ -0,0 +1,105 @@
...
+        retries: 150
+        delay: 3
+        register: powered_off_node
+        failed_when: >
+          (powered_off_node.stderr != "") or
+          (powered_off_node.rc != 0) or
+          (powered_off_node.stdout != "1")
+        until: powered_off_node.stdout == "1"
+
+  - name: Power om "{{ WORKER_NODE }}"
```
furkatgofurov7, line 81: `on` maybe?

vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
```diff
@@ -0,0 +1,105 @@
...
+          KUBECONFIG: "/home/{{ IMAGE_USERNAME }}/.kube/config"
+        retries: 150
+        delay: 3
+        register: powered_off_node
+        failed_when: >
+          (powered_off_node.stderr != "") or
+          (powered_off_node.rc != 0) or
+          (powered_off_node.stdout != "1")
+        until: powered_off_node.stdout == "1"
+
```
maelk, line 80: you should add something to verify that the vm is properly off IMO

maelk: like checking virsh list

vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
```diff
@@ -0,0 +1,105 @@
...
+        environment:
+          KUBECONFIG: "/home/{{ IMAGE_USERNAME }}/.kube/config"
+        retries: 150
+        delay: 3
+        register: powered_on_node
+        failed_when: >
+          (powered_on_node.stderr != "") or
+          (powered_on_node.rc != 0) or
+          (powered_on_node.stdout != "2")
+        until: powered_on_node.stdout == "2"
```
maelk, line 105: Same here, even though this condition would not pass if the node is not back online, I would still verify the virsh output.

vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
fmuyassarov: fixed

vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
```diff
@@ -0,0 +1,105 @@
...
+       | .metadata.name'
+    register: worker_node_name
+
+  - name: Define WORKER_NODE variable with worker node name
+    set_fact:
+      WORKER_NODE: "{{ worker_node_name.stdout }}"
+
+  - name: Reboot "{{ WORKER_NODE }}"
+    shell: |
+       kubectl annotate bmh "{{ WORKER_NODE }}" -n metal3 reboot.metal3.io=
```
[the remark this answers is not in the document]
fmuyassarov: fixed :)

vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
fmuyassarov: Virsh checking added

vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
fmuyassarov: added virsh checking

fmuyassarov:
> I am a bit concerned with the hardcoded numbers when we consider that the number of nodes will probably be 4. Would that have any impact ?

No, because in the remediation test we provision only two BMH. As such, the other two libvirt VMs are not going to have any impact on kubectl checking. 

> Also, for the future, should the test also be done with a HA cluster, and turning off a master ? to test the resilience of the setup. This is not a direct test of the feature, but might be good to have. The same as turning off fully the cluster (with virsh command) and bringing it up again.

Yes, I completely agree that we need to test HA scenarios as well, but for now maybe we can try to get a single master scenario working and then add HA based tests on top of it. 

Regarding `virsh ` comments, I added a `virsh `command to check only shutted off VMs, after we reboot/power-off a BMH. I couldn't figure out a simple way with `virsh ` to check that rebooted libvirt VM is again running but anyway we have `kubectl `command to verify that the node is Ready again.

vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
```diff
@@ -0,0 +1,127 @@
...
+
+  - name: Get the name of the worker libvirt VM
+    set_fact:
+      WORKER_NODE_VM: "node_2"
+    when: WORKER_NODE == "node-2"
+
+  - name: Get the name of the worker libvirt VM
+    set_fact:
+      WORKER_NODE_VM: "node_3"
+    when: WORKER_NODE == "node-3"
```
maelk, line 31: why not using the replace filter, it would far more elegant

vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
```diff
@@ -0,0 +1,127 @@
...
+       kubectl annotate bmh "{{ WORKER_NODE }}" -n "{{ NAMESPACE }}" reboot.metal3.io=
+
+  - name: List only shutdown VMs
+    virt:
+      command: list_vms
+      state: shutdown
+    register: shutdown_vms
+    retries: 50
+    delay: 10
+    until: shutdown_vms.list_vms == ["{{ WORKER_NODE_VM }}"]
```
maelk, line 44: what if the other nodes are off too ? this might not match, because you would have three nodes shutdown

fmuyassarov: That's a good point. I had the same issue for checking running VMs and that's why I didn't include it.

fmuyassarov: Should we then exclude libvirt checking? or add a small shell script for checking node name in the array of strings ?

vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
fmuyassarov: The problem with replace is that you are required to give the file to be modified. I don't think that you can replace variable itself.

maelk: I'm talking about the variable filter : "{{ WORKER_NODE | replace('-','_') }}"

vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
maelk: what about until: 'WORKER_NODE_VM in shutdown_vms.list_vms' or something of the same idea. you can use the in operator. you could also use the select filter, to match that vm name only. There are quite some other alternatives

fmuyassarov: Thanks, done.

vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
fmuyassarov: Done.

PR #333, opened 2020-06-03
Remove hardcoded IPs for external subnet

### LINKED ISSUE #316: Hardcoded registry IP in local images variables
The external subnet is parameterized, with a default to 192.168.111.0/24 . However, the registry IP is hardcoded, when setting up docker/podman in ubuntu/centos_install_requirements.sh and in the image when using local images (02_* and 03_*) as 192.168.111.1:5000 . This should be parameterized, setting the IP address in a variable in lib/network.sh and using it in all those files.

felixest:
The external subnet is configurable, but it was still hardcoded in some
places. Especially the registry running on the host was hardcoded.

Now the IP of the host can be set as a parameter (defaults to the first
host in the EXTERNAL_SUBNET_V4).

Closes #316

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -106,6 +106,7 @@ fi
 
 ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -e "{use_firewalld: $USE_FIREWALLD}" \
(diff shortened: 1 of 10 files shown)


Jaakko-Os:
Good work!

PR #336, opened 2020-06-08
🌱 Export missing variables in BMO as part of CAPM3 deployment

furkatgofurov7:
Exporting missing variables in CI failure for BMO as part of CAPM3 deployment in cluster init

Related to: [PR68](https://github.com/metal3-io/cluster-api-provider-metal3/pull/68)


lib/network.sh
```diff
@@ -118,3 +118,8 @@ fi
 export REGISTRY=${REGISTRY:-"${EXTERNAL_SUBNET_V4_HOST}:5000"}
 
 network_address INITIAL_IRONICBRIDGE_IP "$PROVISIONING_NETWORK" 9
+
+export DEPLOY_KERNEL_URL=${DEPLOY_KERNEL_URL:-"http://172.22.0.1:6180/images/ironic-python-agent.kernel"}
+export DEPLOY_RAMDISK_URL=${DEPLOY_RAMDISK_URL:-"http://172.22.0.1:6180/images/ironic-python-agent.initramfs"}
+export IRONIC_URL=${IRONIC_URL:-"http://172.22.0.1:6385/v1/"}
+export IRONIC_INSPECTOR_URL=${IRONIC_INSPECTOR_URL:-"http://172.22.0.1:5050/v1/"}
```
kashifest, line 125: I do not think this is necessary since you are exporting the variables already in the manifests via [this](https://github.com/metal3-io/cluster-api-provider-metal3/pull/68/files#diff-6cbfde9715d028c7a46203cdc6acb878) ??

maelk: It's at different places. This is necessary for 03_* "clusterctl init" to pass. However, I think that what you point out should use the variables exported here instead of redefining them. In addition, you should not hardcode the IP, there is already a variable containing the ironic IP address.

furkatgofurov7: As I understood, for clusterctl init those variables needs to be exported in 03.script, @maelk is that correct?

kashifest: I thought the reason to put it in CAPM3 was to make usable in metal3-dev-env via  [this](https://github.com/metal3-io/metal3-dev-env/blob/master/03_launch_mgmt_cluster.sh#L125)  which in-turn does this (https://github.com/metal3-io/cluster-api-provider-metal3/blob/master/Makefile#L352). So I thought only a sourcing of this file was missing before the call of clusterctl config which would use the template

furkatgofurov7: @maelk I have updated both what @kashifest pointed out [here](https://github.com/metal3-io/cluster-api-provider-metal3/blob/64b0dc9e682bfe2cae3e91c76ad67399ab306306/examples/clusterctl-templates/example_variables.rc) and parameterized this patch as well

furkatgofurov7: @kashifest that is what I thought as well, but `make release-manifests` is passing in CI, however [this](https://github.com/metal3-io/metal3-dev-env/blob/fd89ae0eabfb34c6d15d9719952ac13db7aeeaae/03_launch_mgmt_cluster.sh#L268) is not, and it needs those vars to be exported, IMO.

maelk: @kashifest we are talking about clusterctl init, not config!

kashifest: @maelk shouldn't the variable substitution take place in config so that the templates are rightly generated to be used by init?

kashifest: Sorry, my order was wrong, ignore my comments please

PR #337, opened 2020-06-08
Take CAPM3 Metadata and IPAM feature into use for target cluster

maelk:
Since IPAM and Metadata were implemented in CAPM3, they could be used in metal3-dev-env to set the IP address statically to test the feature.


lib/network.sh
```diff
@@ -76,7 +77,11 @@ fi
 # Calculate DHCP range
 network_address dhcp_range_start "$PROVISIONING_NETWORK" 10
 network_address dhcp_range_end "$PROVISIONING_NETWORK" 100
+network_address PROVISIONING_POOL_RANGE_START "$PROVISIONING_NETWORK" 100
+network_address PROVISIONING_POOL_RANGE_END "$PROVISIONING_NETWORK" 200
```
jan-est, line 81: Are these ranges only for the cases where user does not define them?

vm-setup/roles/v1aX_integration_test/vars/main.yml
```diff
@@ -17,6 +17,18 @@ KUBERNETES_VERSION: "{{ lookup('env', 'KUBERNETES_VERSION') | default('v1.18.0',
 KUBERNETES_BINARIES_VERSION: "{{ lookup('env', 'KUBERNETES_BINARIES_VERSION') | default(lookup('env', 'KUBERNETES_VERSION') | default('v1.18.0', true), true) }}"
 KUBERNETES_BINARIES_CONFIG_VERSION: "{{ lookup('env', 'KUBERNETES_BINARIES_CONFIG_VERSION') | default('v0.2.7', true) }}"
 
+IP_STACK: "{{ lookup('env', 'IP_STACK') | default('v4', true) }}"
+PROVISIONING_POOL_RANGE_START: "{{ lookup('env', 'PROVISIONING_POOL_RANGE_START') | default('172.22.0.100', true) }}"
+PROVISIONING_POOL_RANGE_END: "{{ lookup('env', 'PROVISIONING_POOL_RANGE_END') | default('172.22.0.200', true) }}"
+PROVISIONING_CIDR: "{{ lookup('env', 'PROVISIONING_CIDR') | default('24', true) }}"
+BAREMETALV4_POOL_RANGE_START: "{{ lookup('env', 'BAREMETALV4_POOL_RANGE_START') | default('192.168.111.100', true) }}"
+BAREMETALV4_POOL_RANGE_END: "{{ lookup('env', 'BAREMETALV4_POOL_RANGE_END') | default('192.168.111.100', true) }}"
```
furkatgofurov7, line 25: should `default` be '192.168.111.200' for `BAREMETALV4_POOL_RANGE_END`?

jan-est, line 25: Should BAREMETALV4_POOL_RANGE_START and BAREMETALV4_POOL_RANGE_END be both defaulted default('192.168.111.100', true)?

maelk: good catch

lib/network.sh
maelk: no, they are always defined

vm-setup/roles/v1aX_integration_test/vars/main.yml
maelk: fixing it

maelk:
The v1a4 tests passed previously and the change made since then is not impacting (defaulting a variable to a different value, since the variable is anyways defined)

lib/network.sh
```diff
@@ -115,7 +110,50 @@ if [[ "${EPHEMERAL_CLUSTER}" == "minikube" ]] && [[ -n "${EXTERNAL_SUBNET_V6}" ]
...
+  export EXTERNAL_SUBNET_V6_PREFIX
+  if [[ -z "${EXTERNAL_SUBNET_V6_HOST}" ]]; then
+    network_address EXTERNAL_SUBNET_V6_HOST "$EXTERNAL_SUBNET_V6" 1
+  fi
+
+  # Calculate DHCP range for baremetal network (20 to 60 is the libvirt dhcp) IPv6
+  network_address BAREMETALV6_POOL_RANGE_START "$EXTERNAL_SUBNET_V6" 100
+  network_address BAREMETALV6_POOL_RANGE_END "$EXTERNAL_SUBNET_V6" 200
+  export BAREMETALV6_POOL_RANGE_START
+  export BAREMETALV6_POOL_RANGE_END
```
felixest, line 144: After network_address or prefixlen are called, the variables are already exported. Some exports may be unnecessary here

PR #339, opened 2020-06-09
Upgrade test Makefile update

Jaakko-Os:


furkatgofurov7:
lgtm

fmuyassarov:
Can we also add upgrade target in the root Makefile with this PR? or should it go in a separate PR?

maelk:
maybe you should run / test-features too ?

fmuyassarov:
> maybe you should run / test-features too ?

It will not run upgrade tests anyway because upgrade Makefile target is currently excluded from root Makefile.

kashifest:
> > maybe you should run / test-features too ?
> 
> It will not run upgrade tests anyway because upgrade Makefile target is currently excluded from root Makefile.

True. Once we are sure that upgrade tests are running properly with the framework, we can push the root Makefile changes also in the same PR. @Jaakko-Os  whats the status of upgrade tests? Are they running OK with the framework?

Jaakko-Os:
> > > maybe you should run / test-features too ?
> > 
> > 
> > It will not run upgrade tests anyway because upgrade Makefile target is currently excluded from root Makefile.
> 
> True. Once we are sure that upgrade tests are running properly with the framework, we can push the root Makefile changes also in the same PR. @Jaakko-Os whats the status of upgrade tests? Are they running OK with the framework?

Lets hold the root Makefile change. Upgrade tests are being run but some issues need to be fixed.

PR #341, opened 2020-06-11
Fix exporting variables for provisioning

fmuyassarov:
NUM_OF_MASTER_REPLICAS  and NUM_OF_WORKER_REPLICAS variables value isn't passed to  clusterctl command in [here](https://github.com/metal3-io/metal3-dev-env/blob/ad8b9f630ee515a82b1d9cc0628f41db83fe5054/vm-setup/roles/v1aX_integration_test/tasks/generate_templates.yml#L18), and clustertl is using default values ( i.e. 1 for both of the variables) from [/vars/main.yml](https://github.com/metal3-io/metal3-dev-env/blob/ad8b9f630ee515a82b1d9cc0628f41db83fe5054/vm-setup/roles/v1aX_integration_test/vars/main.yml#L12). Exporting those variables in Makefile of each feature will allow to provision correct number of BMHs as given in Makefile.

Also increase timeout to wait for BMHs to be provisioned so that the task doesn't fail because of the timeout when provisioning 4 BMHs.

### DIFF
--- a/scripts/feature_tests/upgrade/Makefile
+++ b/scripts/feature_tests/upgrade/Makefile
@@ -1,8 +1,8 @@
 M3PATH := "$(dirname "$(readlink -f "${0}")")/../../"
 # This needs to be moved to upgrade.sh and different value should be set before
 # each use case.
-# NUM_OF_MASTER_REPLICAS := 1
-# NUM_OF_WORKER_REPLICAS := 1
+# export NUM_OF_MASTER_REPLICAS := 1
+# export NUM_OF_WORKER_REPLICAS := 1
 
 all: upgrade
 
--- a/scripts/feature_tests/README.md
+++ b/scripts/feature_tests/README.md
@@ -2,7 +2,7 @@
 
 ## Build Status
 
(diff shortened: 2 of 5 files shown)


scripts/feature_tests/upgrade/Makefile
```diff
@@ -1,8 +1,8 @@
 M3PATH := "$(dirname "$(readlink -f "${0}")")/../../"
 # This needs to be moved to upgrade.sh and different value should be set before
 # each use case.
-# NUM_OF_MASTER_REPLICAS := 1
-# NUM_OF_WORKER_REPLICAS := 1
+# export NUM_OF_MASTER_REPLICAS := 1
+# export NUM_OF_WORKER_REPLICAS := 1
```
Jaakko-Os, line 6: upgrade tests modify these variables between and during(not all cases) test runs so the exports are not actually needed here

fmuyassarov: OK, I added them so that you also don't run to the same issue in case if you use those vars. Those lines will be deleted anyway when upgrade is ready.

PR #346, opened 2020-06-18
Add export of CAPM3RELEASE for use in ansible

maelk:
without this, ansible defaults CAPM3RELEASE to v0.3.1


kashifest:
I was doing this as part of my pivoting work, but its good that you are doing it separately

PR #347, opened 2020-06-18
Update BMO deployment & Add Pivot Back

kashifest:
This PR updates the pivot test ansible script. Specifically

-  It takes care of the updated BMO and Ironic deployment Paths 
-  Adds Pivot back to source cluster capabilities


vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -33,30 +54,17 @@
...
-          line: "PROVISIONING_INTERFACE={{ IRONIC_ENDPOINT_BRIDGE }}"
-
-        # Install ironic
-      - name: Install ironic
-        shell: "bash /home/{{ IMAGE_USERNAME }}/baremetal-operator/tools/run_local_ironic.sh"
-        environment:
-          CONTAINER_RUNTIME: "{{ CONTAINER_RUNTIME }}"
-        become: yes
-        become_user: root
+        when: CAPI_VERSION == "v1alpha3"
```
fmuyassarov, line 57: You could use this condition once for the entire block so that you don't have to duplicate it in the next task under the same block.

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -1,9 +1,24 @@
 ---
+  - name: Remove ironic container from source cluster
+    vars:
+      ansible_python_interpreter: /usr/bin/python3
```
fmuyassarov, line 4: Is this needed here?

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -110,3 +118,24 @@
...
+      kind: Deployment
+      state: absent
+      namespace: "{{ NAMESPACE }}"
+    environment:
+      KUBECONFIG: "/tmp/target.yaml"
+
+  - name: Install Ironic in Source cluster
+    shell: "bash /home/$USER/go/src/github.com/metal3-io/baremetal-operator/tools/run_local_ironic.sh"
+    environment:
+      CONTAINER_RUNTIME: "{{ CONTAINER_RUNTIME }}"
```
maelk, line 137: this is specific to Kind, if someone would attempt to run this with minikube it would fail

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -1,9 +1,24 @@
...
+    docker_container:
+      name: "{{ item }}"
+      state: absent
+    with_items:
+       - ironic
+       - ironic-inspector
+       - dnsmasq
+       - httpd
+       - mariadb
+       - ironic-endpoint-keepalived
```
maelk, line 14: this is specific to Kind, if someone would attempt to run this with minikube would it fail ? If not failing, at least it would not undeploy ironic on Minikube. Until now everything was pretty much agnostic of the source cluster, but with this, you either need to tackle both or make it really clear that this is not supposed to be run with Minikube. I would much prefer the first option tbh.

vm-setup/roles/v1aX_integration_test/tasks/move.yml
kashifest: Actually we have explicitly mentioned it here https://github.com/metal3-io/metal3-dev-env/tree/master/scripts/feature_tests to use kind

vm-setup/roles/v1aX_integration_test/tasks/move.yml
kashifest: Please see the above comment.

vm-setup/roles/v1aX_integration_test/tasks/move.yml
kashifest: this docker_container module is installed using pip3 so that I would explicitly use python3 interpreter. Is it wrong?

vm-setup/roles/v1aX_integration_test/tasks/move.yml
maelk: well it is not mentioned that it should not be run with Minikube, it just says that it defaults to Kind

maelk: and until now there wasn't really anything preventing from running it with Minikube, was there ?

kashifest: TBH, feature test is not tested at all with minikube, so I cannot specifically say that it would work or not, since the entity which is calling this move.yml is not supposed to call it with minikube.

maelk: Then please make it clear that it shall not be run with Minikube. Saying that it defaults to Kind is not clear enough.

vm-setup/roles/v1aX_integration_test/tasks/move.yml
fmuyassarov: I think it is not needed here, because we are anyway using python 3 on the source cluster VM.

vm-setup/roles/v1aX_integration_test/tasks/move.yml
kashifest: I have added additional code and check for minikube, pleas review

vm-setup/roles/v1aX_integration_test/tasks/move.yml
kashifest: I have added additional code and check for minikube, pleas review

vm-setup/roles/v1aX_integration_test/tasks/move.yml
kashifest: Removed

vm-setup/roles/v1aX_integration_test/tasks/move.yml
kashifest: Done. Thanks

PR #349, opened 2020-06-22
Reduce target vm ram to 4096 by default

maelk:
With Raw image streaming, we do not need that much memory anymore


kashifest:
> Would you mind to update the README as well? otherwise lgtm

Ahh you are right. The README shouldbe updated as well. 

/lgtm cancel

maelk:
updated, thanks for the reminder

PR #350, opened 2020-06-24
Fix ansible version for centos to 2.9.8

kashifest:
Latest Ansible version in Centos is not working properly in metal3-dev-env.  This PR fixes the ansible version to 2.9.8 for Centos to fix the issue for now and get the CI going. This needs further investigation.  An issue has been created to track this. 

Issue #353

### DIFF
--- a/centos_install_requirements.sh
+++ b/centos_install_requirements.sh
@@ -37,10 +37,11 @@ fi
 
 # Install required packages
 sudo dnf -y install \
-  ansible \
   redhat-lsb-core \
   python3-pip \
   wget
+
+sudo pip3 install ansible==2.9.8
 
 if [[ $DISTRO == "centos7" ]]; then
   # Install pip since it is used by the k8s Ansible module


vm-setup/roles/v1aX_integration_test/tasks/verify.yml
```diff
@@ -85,6 +85,8 @@
           url: https://docs.projectcalico.org/v3.9/manifests/calico.yaml
           dest: /tmp/
           mode: '664'
```
fmuyassarov, line 87: Perhaps we can try with `pipelining = True ` in inventory.ini.

fmuyassarov: Or set it as an  environmental variable only for that specific task.

fmuyassarov:
For now we can merge it so that the Centos CI job is not failing and we don't block any PR out there, and the issue with the latest version of ansible can be investigated in parallel.

maelk:
/approve
Linking the issue here : #353

PR #352, opened 2020-06-24
Add new "mixed" vm_driver

derekhiggins:
Add a new mode for the power driver that uses both
redfish and ipmi. Adding node with both types of BMC
will allow us to CI both in a single CI job.

### DIFF
--- a/config_example.sh
+++ b/config_example.sh
@@ -91,7 +91,8 @@
 #export TEST_MAX_TIME=120
 
 #
-# Set the driver. The default value is 'ipmi'
+# Set the driver. The default value is 'mixed' (alternate nodes between ipmi
+# and redfish).
 #
 #export BMC_DRIVER="redfish"
 
--- a/vm-setup/roles/firewall/defaults/main.yml
+++ b/vm-setup/roles/firewall/defaults/main.yml
@@ -3,6 +3,7 @@ baremetal_interface: baremetal
 provisioning_interface: provisioning
 external_subnet_v4: "192.168.111.0/24"
 vbmc_port_range: "6230:6235"
+sushy_port: 8000
 firewall_rule_state: present
 ironic_ports:
   - "6180"
--- a/vm-setup/roles/firewall/tasks/firewalld.yaml
+++ b/vm-setup/roles/firewall/tasks/firewalld.yaml
@@ -48,3 +48,11 @@
     permanent: yes
     state: enabled
     immediate: yes
+
+- name: "firewalld: sushy Port"
+  firewalld:
+    zone: libvirt
+    port: "{{ sushy_port }}/tcp"
+    permanent: yes
+    state: enabled
+    immediate: yes
--- a/vm-setup/roles/firewall/tasks/iptables.yaml
+++ b/vm-setup/roles/firewall/tasks/iptables.yaml
@@ -15,6 +15,16 @@
     jump: ACCEPT
     state: "{{ firewall_rule_state }}"
 
+- name: "iptables: sushy Port"
+  iptables:
+    chain: INPUT
+    in_interface: "{{ baremetal_interface }}"
+    protocol: tcp
+    match: tcp
+    destination_port: "{{ sushy_port }}"
+    jump: ACCEPT
+    state: "{{ firewall_rule_state }}"
+
 - name: "iptables: Established and related"
   iptables:
     chain: FORWARD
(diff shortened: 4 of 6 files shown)


maelk:
/test-integration
( centos tests are broken right now due to an ansible package upgrade)

fmuyassarov:
> /test-integration
> ( centos tests are broken right now due to an ansible package upgrade)

The issue with Ansible version is fixed now in [#350](https://github.com/metal3-io/metal3-dev-env/pull/350).

maelk:
Could you maybe change the default driver to mixed ? so that we test this change in the CI and keep running with mixed by default in CI ? thank you

derekhiggins:
> Could you maybe change the default driver to mixed ? so that we test this change in the CI and keep running with mixed by default in CI ? thank you

My intention was to switch our dev-scripts default (where possible) to "mixed". I was reluctant to change the default here as it may break ironic images that don't support redfish. Or maybe I could change it for metal3-dev-env CI only?

maelk:
The metal3-dev-env setup is very opinionated. the ironic image deployed by default supports redfish so IMO we could change the default driver. If someone deploys with an image that does not support redfish in metal3-dev-env, then they can update this parameter. We should definitely change it in the CI if we do not change it in metal3-dev-env.
Could you also update the config_example.sh to add the mixed as an accepted value ? Maybe the try-it should also be updated in the website.

derekhiggins:
> The metal3-dev-env setup is very opinionated. the ironic image deployed by default supports redfish so IMO we could change the default driver. If someone deploys with an image that does not support redfish in metal3-dev-env, then they can update this parameter. We should definitely change it in the CI if we do not change it in metal3-dev-env.
> Could you also update the config_example.sh to add the mixed as an accepted value ? Maybe the try-it should also be updated in the website.

Ack, will change the default, it was dev-scripts using openshift 4.3 that I was worried about breaking but it uses a pinned version of metal3-dev-env so it wont be effected by this merging.

I'll also update the website.

maelk:
/test-integration
/test-centos-integration
lgtm, waiting for the tests

maelk:
could that be a firewall issue ?
`For node 80c8fd20-3a6d-4094-834a-b397a59dc5a6, got a connection error from Redfish at address "http://192.168.111.1:8000" using auth type "auto" when fetching System "/redfish/v1/Systems/596b8c21-7713-4e4f-b500-0612c9900e3c". Error: Unable to connect to http://192.168.111.1:8000/redfish/v1/. Error: HTTPConnectionPool(host='192.168.111.1', port=8000): Max retries exceeded with url: /redfish/v1/ (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7eff48450ba8>: Failed to establish a new connection: [Errno 111] ECONNREFUSED',)): sushy.exceptions.ConnectionError: Unable to connect to http://192.168.111.1:8000/redfish/v1/`

derekhiggins:
> could that be a firewall issue ?

Looks like it, push a patch to add firewall rules for sushy, it it doesn't work I'll try and reproduce.

vm-setup/roles/common/defaults/main.yml
```diff
@@ -14,6 +14,7 @@ default_vcpu: 4
 num_nodes: 2
 extradisks: false
 virtualbmc_base_port: 6230
+sushy_port: 8000
```
maelk, line 17: this should be defined in vm-setup/roles/firewall/defaults/main.yaml , if in common, ansible does not find it :
`The task includes an option with an undefined variable. The error was: 'sushy_port' is undefined\n`

derekhiggins: > this should be defined in vm-setup/roles/firewall/defaults/main.yaml , if in common, ansible does not find it :
> `The task includes an option with an undefined variable. The error was: 'sushy_port' is undefined\n`

thanks, I was hoping I'd be able to use it from ironic_nodes.json.j2 by putting in common/defaults/main.yml but will put it back into the firewall role

derekhiggins:
hmm, seeing an error in the ironic logs, trying to find the problem
`
2020-07-07 14:49:44.140 27 ERROR ironic.conductor.manager [req-0e52cda8-79ad-4cce-a193-89cb501d6180 - - - - -] Suspicious activity detected for node b4b19ef7-179f-4ad5-809f-c31ae48239eb when attempting to heartbeat. Heartbeat request has been rejected as the version of ironic-python-agent indicated in the heartbeat operation should support agent token functionality.
`

derekhiggins:
/test-centos-integration
/test-integration

should be fixed now by https://github.com/metal3-io/ironic-image/pull/173

PR #358, opened 2020-07-01
Add veth interface to keep provisioning bridge up

fmuyassarov:
During cleanup process (deprovisioning of BMHs) in feature tests, libvirt VMs are turned off which results in loss of provisioning bridge. Since the bridge is down, next provisioning is failing because 172.22.0.1 ironic endpoint is not there. Creating veth will keep the provisioning bridge up and running and make sure that the ironic endpoint will be there.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -178,10 +178,17 @@ VBMC_IMAGE=${VBMC_LOCAL_IMAGE:-$VBMC_IMAGE}
 SUSHY_TOOLS_IMAGE=${SUSHY_TOOLS_LOCAL_IMAGE:-$SUSHY_TOOLS_IMAGE}
 
 # Start httpd container
-#shellcheck disable=SC2086
-sudo "${CONTAINER_RUNTIME}" run -d --net host --privileged --name httpd-infra \
-     ${POD_NAME_INFRA} -v "$IRONIC_DATA_DIR":/shared --entrypoint /bin/runhttpd\
-     "${IRONIC_IMAGE}"
+if [[ $OS == ubuntu ]]; then
+  #shellcheck disable=SC2086
+  sudo "${CONTAINER_RUNTIME}" run -d --net host --privileged --name httpd-infra ${POD_NAME_INFRA} \
+      -v "$IRONIC_DATA_DIR":/shared --entrypoint /bin/runhttpd \
+      --env "PROVISIONING_INTERFACE=ironicendpoint" "${IRONIC_IMAGE}"
+else
+  #shellcheck disable=SC2086
+  sudo "${CONTAINER_RUNTIME}" run -d --net host --privileged --name httpd-infra ${POD_NAME_INFRA} \
+      -v "$IRONIC_DATA_DIR":/shared --entrypoint /bin/runhttpd \
+      "${IRONIC_IMAGE}"
+fi
 
 # Start vbmc and sushy containers
 #shellcheck disable=SC2086
--- a/ubuntu_bridge_network_configuration.sh
+++ b/ubuntu_bridge_network_configuration.sh
@@ -6,16 +6,28 @@ set -xe
 source lib/logging.sh
 # shellcheck disable=SC1091
 source lib/common.sh
+# shellcheck disable=SC1091
+source lib/network.sh
 
 if [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
      # Adding an IP address in the libvirt definition for this network results in
      # dnsmasq being run, we don't want that as we have our own dnsmasq, so set
-     # the IP address here
+     # the IP address here.
+     # Create a veth iterface peer.
+     sudo ip link add ironicendpoint type veth peer name ironic-peer
+     # Create provisioning bridge.
      sudo brctl addbr provisioning
      # sudo ifconfig provisioning 172.22.0.1 netmask 255.255.255.0 up
      # Use ip command. ifconfig commands are deprecated now.
-     sudo ip addr add dev provisioning "$PROVISIONING_IP"/"$PROVISIONING_CIDR"
      sudo ip link set provisioning up
+     if [[ "${PROVISIONING_IPV6}" == "true" ]]; then
+        sudo ip -6 addr add "$PROVISIONING_IP"/"$PROVISIONING_CIDR" dev ironicendpoint
+      else
+        sudo ip addr add dev ironicendpoint "$PROVISIONING_IP"/"$PROVISIONING_CIDR"
+     fi
+     sudo brctl addif provisioning ironic-peer
+     sudo ip link set ironicendpoint up
+     sudo ip link set ironic-peer up
 
      # Need to pass the provision interface for bare metal
      if [ "$PRO_IF" ]; then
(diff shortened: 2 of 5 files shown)


ubuntu_bridge_network_configuration.sh
```diff
@@ -10,16 +10,23 @@ source lib/common.sh
...
+     # Create provisioning bridge.
      sudo brctl addbr provisioning
      # sudo ifconfig provisioning 172.22.0.1 netmask 255.255.255.0 up
      # Use ip command. ifconfig commands are deprecated now.
      sudo ip addr add dev provisioning "$PROVISIONING_IP"/"$PROVISIONING_CIDR"
      sudo ip link set provisioning up
 
      # Need to pass the provision interface for bare metal
-     if [ "$PRO_IF" ]; then
+     if [ "$PRO_IF1" && "$PRO_IF2" ]; then
```
furkatgofurov7, line 24: this should be either $PRO_IF1 **OR** $PRO_IF2 ?

fmuyassarov:
> by doing it this way you are removing the option to use PRO_IF to add a single interface to the bridge as you require PRO_IF2 now

Fixed.

ubuntu_bridge_network_configuration.sh
fmuyassarov: Fixed.

ubuntu_bridge_network_configuration.sh
```diff
@@ -10,16 +10,23 @@ source lib/common.sh
 if [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
      # Adding an IP address in the libvirt definition for this network results in
      # dnsmasq being run, we don't want that as we have our own dnsmasq, so set
-     # the IP address here
+     # the IP address here.
+     # Create a veth iterface peer.
+     sudo ip link add "$PRO_IF1" type veth peer name "$PRO_IF2"
```
maelk, line 15: what if the user sets PRO_IF1 to be a physical interface of his machine ? and defines PRO_IF2 to "" ?

fmuyassarov: Fixed based on our discussion.

ubuntu_bridge_network_configuration.sh
```diff
@@ -10,12 +10,18 @@ source lib/common.sh
 if [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
      # Adding an IP address in the libvirt definition for this network results in
      # dnsmasq being run, we don't want that as we have our own dnsmasq, so set
-     # the IP address here
+     # the IP address here.
+     # Create a veth iterface peer.
+     sudo ip link add ironicendpoint type veth peer name veth2
```
maelk, line 15: Can you please rename veth2 to something else ? like ironicendpoint-veth2 or anything that would not be as predictible for collisions. thank you

vm-setup/roles/libvirt/tasks/network_teardown_tasks.yml
```diff
@@ -14,12 +14,14 @@
...
 
-- name: Delete bridges on Ubuntu
+- name: Delete bridges and veth interfaces on Ubuntu
   shell: |
      sudo ip link set baremetal down
      sudo ip link set provisioning down
+     sudo ip link set ironicendpoint down
      brctl delbr baremetal | true
      brctl delbr provisioning | true
+     sudo ip link del ironicendpoint down | true
```
maelk, line 24: should the "down" be here ?

ubuntu_bridge_network_configuration.sh
fmuyassarov: Renamed. It is not really explicit or not ironicendpoint-veth2 because maximum length of characters should be 16 for the interface name.

vm-setup/roles/libvirt/tasks/network_teardown_tasks.yml
fmuyassarov: Good catch. Fixed, thank you.

fmuyassarov:
I don't know what is the issue in the CI, because my local tests are passing. 
/test-features

fmuyassarov:
/hold 
Testing with locally built ironic-inspector image

fmuyassarov:
/test-centos-integration
/test-integration
@maelk I have added a condition for adding provisioning interface var, because seems there is no veth interface on centos8 but bond interfaces. If we go for bond interface then that is suppose to be attached to an existing/new interface. So, I think it would be easier to rely on condition for now.

fmuyassarov:
> /test-centos-integration
> /test-integration
> @maelk I have added a condition for adding provisioning interface var, because seems there is no veth interface on centos8 but bond interfaces. If we go for bond interface then that is suppose to be attached to an existing/new interface. So, I think it would be easier to rely on condition for now.

Or if you strongly suggest to avoid having condition, then I can try to test it with bond interfaces.

02_configure_host.sh
```diff
@@ -51,57 +51,8 @@ if [[ $OS == ubuntu ]]; then
...
-  # restart the libvirt network so it applies an ip to the bridge
-  if [ "$MANAGE_BR_BRIDGE" == "y" ] ; then
-      sudo virsh net-destroy baremetal
-      sudo virsh net-start baremetal
-      if [ "$INT_IF" ]; then #Need to bring UP the NIC after destroying the libvirt network
-          sudo ifup "$INT_IF"
-      fi
-  fi
+  # shellcheck disable=SC1091
+  source centos_bridge_network_configuration.sh
```
furkatgofurov7, line 55: This should be `centos_bridge_network_configurations.sh`

maelk:
> > /test-centos-integration
> > /test-integration
> > @maelk I have added a condition for adding provisioning interface var, because seems there is no veth interface on centos8 but bond interfaces. If we go for bond interface then that is suppose to be attached to an existing/new interface. So, I think it would be easier to rely on condition for now.
> 
> Or if you strongly suggest to avoid having condition, then I can try to test it with bond interfaces.

this last patch looks good. let's check the tests.
/test-centos-integration
/test-integration

centos_bridge_network_configuration.sh
```diff
@@ -0,0 +1,69 @@
...
+
+if [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
+    # Create a veth iterface peer.
+    sudo ip link add ironicendpoint type veth peer name ironic-peer
+    # Adding an IP address in the libvirt definition for this network results in
+    # dnsmasq being run, we don't want that as we have our own dnsmasq, so set
+    # the IP address here
+    if [ ! -e /etc/sysconfig/network-scripts/ifcfg-provisioning ] ; then
+        if [[ "${PROVISIONING_IPV6}" == "true" ]]; then
+          echo -e "DEVICE=provisioning\nTYPE=Bridge\nONBOOT=yes\nNM_CONTROLLED=no\nIPV6_AUTOCONF=no\nIPV6INIT=yes" | sudo dd of=/etc/sysconfig/network-scripts/ifcfg-provisioning
```
maelk, line 20: I'm wondering if IPV6INIT should be there. SInce this is now only used as a bridge, we probably don't care about layer 3 configuration, so it might need to be no ?

maelk:
@stbenjam I know you have been working on IPv6 in metal3-dev-env recently. Can you please check that this change does not create any impact wrt IPv6 ? thank you!

centos_bridge_network_configuration.sh
fmuyassarov: As discussed, I have added a condition to run httpd-infra container with PROVISIONING_INTERFACE var.

PR #361, opened 2020-07-02
Make CAPI-controllers's version configurable

Xenwar:
Currently, the version of capi-controllers is the latest.
For upgrade uses cases, we would like is to be frozen in earlier version by default.

This PR fixes the version the latest stable release (v0.3.6 as of now).
It is possible to use any other version by setting the environmental variable ```CAPIRELEASE```


lib/common.sh
```diff
@@ -103,6 +107,8 @@ CAPM3REPO="${CAPM3REPO:-https://github.com/${CAPM3_BASE_URL}}"
 CAPIPATH="${CAPIPATH:-${M3PATH}/cluster-api}"
 CAPM3RELEASEPATH="${CAPM3RELEASEPATH:-https://api.github.com/repos/${CAPM3_BASE_URL}/releases/latest}"
 export CAPM3RELEASE="${CAPM3RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}")}"
+export CAPIRELEASE="${CAPIRELEASE:-v0.3.0}"
```
maelk, line 110: you should do this the same way as it is done for CAPM3, defaulting to latest release, and being able to override to an older release. It is not a good approach to block everything with an older release (including the integration tests) just for one use-case. That use-case should be setting the environment variables to get the setup it needs. So default to latest release and override the version when doing the upgrade tests

lib/common.sh
```diff
@@ -80,6 +81,8 @@ function get_latest_release() {
     echo "$(curl -sL "${1}"  | jq -r '.tag_name')"
 }
 
+export CLUSTERCTL_PATH=${CLUSTERCTL_PATH:-"${HOME}/go/src/github.com/metal3-io/cluster-api/bin"}
```
maelk, line 84: where is this used ?

lib/common.sh
```diff
@@ -93,6 +96,7 @@ BMOPATH="${BMOPATH:-${M3PATH}/baremetal-operator}"
 # shellcheck disable=SC2034
 RUN_LOCAL_IRONIC_SCRIPT="${BMOPATH}/tools/run_local_ironic.sh"
 CAPM3PATH="${CAPM3PATH:-${M3PATH}/cluster-api-provider-metal3}"
+CAPIPATH="${CAPIPATH:-${M3PATH}/cluster-api}"
```
maelk, line 99: this duplicates line 107

lib/common.sh
Xenwar: I decided to use v0.3.0 based on https://github.com/kubernetes-sigs/cluster-api/blob/2ac3728d26593f7c54520999477aad45934e1c59/cmd/clusterctl/hack/local-overrides.py#L51-L68.

For the local override, the version is hard coded in the code itself. I was trying to imitate that. 
But, I can do the change as you suggested as well.

lib/common.sh
Xenwar: removed

lib/common.sh
Xenwar: removed

lib/common.sh
Xenwar: > you should do this the same way as it is done for CAPM3, defaulting to latest release, and being able to override to an older release. It is not a good approach to block everything with an older release (including the integration tests) just for one use-case. That use-case should be setting the environment variables to get the setup it needs. So default to latest release and override the version when doing the upgrade tests

Changed. 
Tested with and without ```export CAPIRELEASE=v0.3.0``
if not export, then ```v0.3.6``` is installed. v0.3.0 otherwise.

PR #365, opened 2020-07-06
Add user check to avoid running as root

### LINKED ISSUE #366: Running make as root is not forbidden but minikube fails the init
If you run the make command as root, it is not prohibited, but when it comes to initialize minikube, it fails if provider is KVM/libvirt

I created the PR https://github.com/metal3-io/metal3-dev-env/pull/365 to add an initial check.

kubealex:
Force using non-root user if EPHEMERAL_CLUSTER is minikube, to avoid the initialization of minikube fail after ALL the execution.

Fixes #366

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -6,6 +6,11 @@ source lib/logging.sh
 # shellcheck disable=SC1091
 source lib/common.sh
 
+if [[ $(id -u) == 0 ]]; then
+  echo "Please run 'make' as a non-root user"
+  exit 1
+fi
+
 if [[ $OS == ubuntu ]]; then
   # shellcheck disable=SC1091
   source ubuntu_install_requirements.sh


01_prepare_host.sh
```diff
@@ -6,6 +6,11 @@ source lib/logging.sh
 # shellcheck disable=SC1091
 source lib/common.sh
 
+if [[ "${EPHEMERAL_CLUSTER}" == "minikube" ]] && [[ $(id -u) == 0 ]]; then
```
maelk, line 9: maybe the test could be independent of the EPHEMERAL_CLUSTER, so that we always make sure not to run as root. wtr, there should not be differences between the two setups IMO.

kubealex:
> Thank you for the PR. Could you maybe please add a "Fixes #366" in your PR comment so that when this gets merged it automatically closes the issue ? thank you.

Done, thank you!

maelk:
thank you. What do you think about making the test independent of the ephemeral cluster we are using ?

kubealex:
> thank you. What do you think about making the test independent of the ephemeral cluster we are using ?

I am a fan of non-root stuff, so from my point of view is more than ok. I just limited it to the particular case to avoid impacting other stuff around and focus on the only real blocking point.

maelk:
I think you can go ahead for Kind too, that's not a problem, you can just remove the check

kubealex:
Done!

01_prepare_host.sh
```diff
@@ -6,6 +6,11 @@ source lib/logging.sh
 # shellcheck disable=SC1091
 source lib/common.sh
 
+if [[ $(id -u) == 0 ]]; then
+  echo "Please run as a non-root user when using minikube as EPHEMERAL_CLUSTER"
```
maelk, line 10: this might need updating then

kubealex: done, sorry!

maelk:
Please squash the commits, then good to go for me

kubealex:
> Please squash the commits, then good to go for me

done!

kubealex:
@maelk seems like tests fail for reasons not linked to the modification, aren't them?

maelk:
yes, Metal3-dev-env is broken currently. we're working on it

kubealex:
@maelk do you see any chance to have this merged?

maelk:
I think this has been opened long enough. I'll merge it.
/approve

PR #367, opened 2020-07-07
Avoid printing skipped Ansible tasks

fmuyassarov:
Avoid printing skipped Ansible tasks output, since we now have too many tasks that run depending on certain condition but still print out some output that's not relevant. This will help to have a bit clearer output both on the CI & local run.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -187,6 +187,9 @@ TEST_MAX_TIME="${TEST_MAX_TIME:-240}"
 FAILS=0
 RESULT_STR=""
 
+# Avoid printing skipped Ansible tasks
+export ANSIBLE_DISPLAY_SKIPPED_HOSTS=no
+
 # Verify requisites/permissions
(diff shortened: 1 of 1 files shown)


lib/common.sh
```diff
@@ -187,6 +187,9 @@ TEST_MAX_TIME="${TEST_MAX_TIME:-240}"
 FAILS=0
 RESULT_STR=""
 
+# Avoid printing skipped Ansible tasks
+export ANSIBLE_DISPLAY_SKIPPED_HOSTS=no
```
furkatgofurov7, line 191: I am not sure if it should be set to `no` but `false` maybe?

fmuyassarov: You can check [here](https://docs.ansible.com/ansible/latest/plugins/callback/skippy.html) that it is no. And also if you compare the CI output of this PR and master you will notice that skipped tasks aren't printed anymore.

PR #370, opened 2020-07-09
Removes loop where there is only a single item

Xenwar:


vm-setup/roles/v1aX_integration_test/tasks/main.yml
```diff
@@ -20,8 +20,6 @@
 
 - name: "Generate templates for v1alpha3"
   include: "{{ item }}.yml"
```
fmuyassarov, line 22:
```suggestion
  include: generate_templates.yml
```

PR #373, opened 2020-07-10
Bug fix: Assign 172.22.0.2 address to ironic BMO configmap. 

smoshiur1237:
During the metal3-dev-env building using v1alpha4 the capm3-ironic-bmo-configmap is using 172.22.0.1 instead of 172.22.0.2. So when we are doing the pivoting with v1alpha4 the BMO is giving error and it's looking for ironicendpoint on 172.22.0.1.  Here is the fix for this error so that the BMO is always looking for .02 address instead of .01 address.

### DIFF
--- a/lib/network.sh
+++ b/lib/network.sh
@@ -157,14 +157,7 @@ fi
 
 network_address INITIAL_IRONICBRIDGE_IP "$PROVISIONING_NETWORK" 9
 
-if [ "${EPHEMERAL_CLUSTER}" == "minikube" ]; then
-  export DEPLOY_KERNEL_URL=${DEPLOY_KERNEL_URL:-"http://${CLUSTER_URL_HOST}:6180/images/ironic-python-agent.kernel"}
-  export DEPLOY_RAMDISK_URL=${DEPLOY_RAMDISK_URL:-"http://${CLUSTER_URL_HOST}:6180/images/ironic-python-agent.initramfs"}
-  export IRONIC_URL=${IRONIC_URL:-"http://${CLUSTER_URL_HOST}:6385/v1/"}
-  export IRONIC_INSPECTOR_URL=${IRONIC_INSPECTOR_URL:-"http://${CLUSTER_URL_HOST}:5050/v1/"}
-else
-  export DEPLOY_KERNEL_URL=${DEPLOY_KERNEL_URL:-"http://${PROVISIONING_URL_HOST}:6180/images/ironic-python-agent.kernel"}
-  export DEPLOY_RAMDISK_URL=${DEPLOY_RAMDISK_URL:-"http://${PROVISIONING_URL_HOST}:6180/images/ironic-python-agent.initramfs"}
-  export IRONIC_URL=${IRONIC_URL:-"http://${PROVISIONING_URL_HOST}:6385/v1/"}
-  export IRONIC_INSPECTOR_URL=${IRONIC_INSPECTOR_URL:-"http://${PROVISIONING_URL_HOST}:5050/v1/"}
-fi
+export DEPLOY_KERNEL_URL=${DEPLOY_KERNEL_URL:-"http://${CLUSTER_URL_HOST}:6180/images/ironic-python-agent.kernel"}
+export DEPLOY_RAMDISK_URL=${DEPLOY_RAMDISK_URL:-"http://${CLUSTER_URL_HOST}:6180/images/ironic-python-agent.initramfs"}
+export IRONIC_URL=${IRONIC_URL:-"http://${CLUSTER_URL_HOST}:6385/v1/"}
+export IRONIC_INSPECTOR_URL=${IRONIC_INSPECTOR_URL:-"http://${CLUSTER_URL_HOST}:5050/v1/"}


03_launch_mgmt_cluster.sh
```diff
@@ -10,7 +10,7 @@ source lib/network.sh
 
 if [ "${EPHEMERAL_CLUSTER}" == "minikube" ]; then
   IRONIC_HOST="${CLUSTER_URL_HOST}"
-  export IRONIC_HOST_IP="${CLUSTER_PROVISIONING_IP}"
+  export IRONIC_HOST_IP="${PROVISIONING_IP}"
```
furkatgofurov7, line 13: Why do we need duplicated IRONIC_HOST_IP exported in the condition, maybe we can just export it before the condition?

lib/network.sh
```diff
@@ -52,19 +52,19 @@ prefixlen PROVISIONING_CIDR "$PROVISIONING_NETWORK"
 export PROVISIONING_CIDR
 export PROVISIONING_NETMASK=${PROVISIONING_NETMASK:-$(python -c "import ipaddress; print(ipaddress.ip_network(u\"$PROVISIONING_NETWORK\").netmask)")}
 
-network_address PROVISIONING_IP "$PROVISIONING_NETWORK" 1
-network_address CLUSTER_PROVISIONING_IP "$PROVISIONING_NETWORK" 2
+network_address PROVISIONING_IP "$PROVISIONING_NETWORK" 2
+#network_address CLUSTER_PROVISIONING_IP "$PROVISIONING_NETWORK" 3
```
furkatgofurov7, line 56: why do we need this?

lib/network.sh
```diff
@@ -52,19 +52,19 @@ prefixlen PROVISIONING_CIDR "$PROVISIONING_NETWORK"
...
 
 export PROVISIONING_IP
-export CLUSTER_PROVISIONING_IP
+#export CLUSTER_PROVISIONING_IP
```
furkatgofurov7, line 59: this one as well?

03_launch_mgmt_cluster.sh
fmuyassarov: This is for testing only for now. The idea is that we want to have a single IP,  that's why there is a duplication.

lib/network.sh
smoshiur1237: We want to use single IP .02 address.

lib/network.sh
smoshiur1237: We want to use single IP .02 address.

03_launch_mgmt_cluster.sh
maelk: I don't get the reason @fmuyassarov  now line 13 and 16 are the same, you can take it out of the condition, as pointed out by @furkatgofurov7

lib/network.sh
maelk: remove the commented line please

lib/network.sh
maelk: same here

maelk:
Maybe some variable renaming could also be beneficial here, to make it clear which does what

03_launch_mgmt_cluster.sh
smoshiur1237: I am reverting the changes and do the variable change only for the ironic-configmap.

lib/network.sh
smoshiur1237: I am reverting the changes and do the variable change only for the ironic-configmap.

lib/network.sh
smoshiur1237: Agreed

lib/network.sh
```diff
@@ -157,14 +157,7 @@ fi
 
 network_address INITIAL_IRONICBRIDGE_IP "$PROVISIONING_NETWORK" 9
 
-if [ "${EPHEMERAL_CLUSTER}" == "minikube" ]; then
   export DEPLOY_KERNEL_URL=${DEPLOY_KERNEL_URL:-"http://${CLUSTER_URL_HOST}:6180/images/ironic-python-agent.kernel"}
   export DEPLOY_RAMDISK_URL=${DEPLOY_RAMDISK_URL:-"http://${CLUSTER_URL_HOST}:6180/images/ironic-python-agent.initramfs"}
   export IRONIC_URL=${IRONIC_URL:-"http://${CLUSTER_URL_HOST}:6385/v1/"}
   export IRONIC_INSPECTOR_URL=${IRONIC_INSPECTOR_URL:-"http://${CLUSTER_URL_HOST}:5050/v1/"}
```
maelk, line 163: please fix indentation

smoshiur1237: Fixed done

smoshiur1237:
> I am still not getting the idea why we are setting `.2` IP everywhere @smoshiur1237 , because BMO configmap should fail if we will not have conditions for `EPHEMERAL_CLUSTER`?

@furkatgofurov7  I have added the .2 address for the bmo-ironic-configmap instead of .01 address. Because before this change, the bmo-ironic-configmap was having . 01 address and cannot complete the pivoting as it was looking for .02 ironicendpoint address in configmap and all the provisioned node was in registering status. For minikube it was configured already in .02 address. In both case, its needed to have .02 as ironic endpoint. That is the reason I removed the condition.

PR #378, opened 2020-07-13
Remove PROVISIONING_INTERFACE customization

fmuyassarov:
Currently we are replacing PROVISIONING_INTERFACE with sed in Metal3-dev-env. [#572](https://github.com/metal3-io/baremetal-operator/pull/572) does the same in ironic_ci.env. This patch removes that modification from metal3-dev-env since it will be taken care by BMO repo itself.


fmuyassarov:
/hold 
until [#572](https://github.com/metal3-io/baremetal-operator/pull/572) gets merged.
/assign @maelk

PR #394, opened 2020-07-27
Add support for Ubuntu 20.04

maelk:
This PR modifies the Ubuntu setup to add support for Ubuntu 20.04 (Focal Fossa)

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -208,7 +208,9 @@ fi
 OS=$(awk -F= '/^ID=/ { print $2 }' /etc/os-release | tr -d '"')
 export OS
 OS_VERSION=$(awk -F= '/^VERSION_ID=/ { print $2 }' /etc/os-release | tr -d '"' | cut -f1 -d'.')
+OS_VERSION_ID=$(awk -F= '/^VERSION_ID=/ { print $2 }' /etc/os-release | tr -d '"')
 export OS_VERSION
+export OS_VERSION_ID
 if [[ $OS == centos ]]; then
   if [[ ${OS_VERSION} != 7 && ${OS_VERSION} != 8 ]]; then
     echo "Required CentOS 7/8 or RHEL 8 or Ubuntu 18.04"
@@ -220,8 +222,9 @@ elif [[ $OS == rhel ]]; then
     exit 1
   fi
 elif [[ $OS == ubuntu ]]; then
-  if [[ ${OS_VERSION} -ne 18 ]]; then
-    echo "Required CentOS 7/8 or RHEL 8 or Ubuntu 18.04"
+  if [[ ${OS_VERSION_ID} != "18.04" && ${OS_VERSION_ID} != "20.04" ]]; then
+    # Ubuntu 18.04 is supported, but the target VM cannot use UEFI, hence we are deprecating it
+    echo "Required CentOS 7/8 or RHEL 8 or Ubuntu 20.04"
     exit 1
   fi
 else
--- a/ubuntu_install_requirements.sh
+++ b/ubuntu_install_requirements.sh
@@ -21,9 +21,14 @@ sudo apt -y install \
   python3-pip \
   python3-setuptools \
   zlib1g-dev \
-  libssl1.0-dev \
   openssh-server \
   wget
+
+if [[ ${OS_VERSION_ID} == "18.04" ]]; then
+  sudo apt -y install libssl1.0-dev
+else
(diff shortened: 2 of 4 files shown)


maelk:
Sure, I forgot about those, thank you

maelk:
actually, the readme is for the target hosts, those are still 18.04. I did not upgrade those

lib/common.sh
```diff
@@ -208,7 +208,9 @@ fi
 OS=$(awk -F= '/^ID=/ { print $2 }' /etc/os-release | tr -d '"')
 export OS
 OS_VERSION=$(awk -F= '/^VERSION_ID=/ { print $2 }' /etc/os-release | tr -d '"' | cut -f1 -d'.')
+OS_VERSION_ID=$(awk -F= '/^VERSION_ID=/ { print $2 }' /etc/os-release | tr -d '"')
```
Xenwar, line 211: OS_VERSION=$(awk -F= '/^VERSION_ID=/ { print $2 }' /etc/os-release | tr -d '"')  and remove OS_VERSION_ID

I do not see the relevance of having both 18 and 18.04 as the comparison can be made string based.

ubuntu_install_requirements.sh
```diff
@@ -21,10 +21,15 @@ sudo apt -y install \
   python3-pip \
   python3-setuptools \
   zlib1g-dev \
-  libssl1.0-dev \
   openssh-server \
   wget
 
+if [[ ${OS_VERSION} -eq 18 ]]; then
```
Xenwar, line 27: See the above comment, 
The comparison can also be made against 18.04, avoiding having a second variable.

ubuntu_install_requirements.sh
```diff
@@ -49,7 +53,11 @@ sudo pip3 install \
   pyYAML
 
 # Set update-alternatives to python3
-sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.6 1
+if [[ ${OS_VERSION} -eq 18 ]]; then
```
Xenwar, line 56: same here

lib/common.sh
```diff
@@ -220,8 +222,8 @@ elif [[ $OS == rhel ]]; then
     exit 1
   fi
 elif [[ $OS == ubuntu ]]; then
-  if [[ ${OS_VERSION} -ne 18 ]]; then
-    echo "Required CentOS 7/8 or RHEL 8 or Ubuntu 18.04"
+  if [[ ${OS_VERSION_ID} != "18.04" && ${OS_VERSION_ID} != "20.04" ]]; then
+    echo "Required CentOS 7/8 or RHEL 8 or Ubuntu 18.04 or 20.04"
```
Xenwar, line 226: or Ubuntu 18.04/20.04

lib/common.sh
maelk: OS_VERSION is used for CentOS, OS_VERSION_ID is used for Ubuntu, because we need a specific minor version, but in CentOS we have to stick with the major

ubuntu_install_requirements.sh
maelk: see my comment above

lib/common.sh
Xenwar: Got it.

fmuyassarov:
@maelk can we run centos job as well?

fmuyassarov:
> @maelk can we run centos job as well?
Ups, a bit late :)

fmuyassarov:
Are you also considering to have changes in project-infra to actually use Ubuntu 20.04?

maelk:
there should not be any need to change project-infra. I'll just switch the base image in the image building jobs

vm-setup/roles/packages_installation/defaults/main.yml
```diff
@@ -1,26 +1,51 @@
...
+        - bridge-utils
+        - jq
+        - unzip
+        - yarn
+        - genisoimage
+        - qemu-kvm
+        - libguestfs-tools
+        - gir1.2-polkit-1.0
+        - libpolkit-agent-1-0
+        - libpolkit-gobject-1-0
```
fmuyassarov, line 48: If the difference between these two blocks is a single package (`libpolkit-backend-1-0`), can't we default packages that are the same for both 18.04 & 20.04?

maelk: done, thanks

PR #398, opened 2020-07-28
Default to docker for ubuntu based dev-env

fmuyassarov:
Set docker as default container runtime for Ubuntu based metal3-dev-env. Currently this default value (podman) is overwritten to docker in project-infra when we trigger the CI. However, when running local metal3-dev-env a user has to export the `CONTAINER_RUNTIME` with docker value if he/she wants use kind as an ephemeral cluster. This patch would reduce the number of vars that the user has to export before running `make`.


fmuyassarov:
> good work!
> there's a rather extensive list of vars to export by the user, maybe most of those could be defaulted too?

Yes I agree. However, I'm not wrong all the vars now should have default values, isn't it so?

PR #403, opened 2020-07-29
Refactor deployment and test scripts

maelk:
This commit reduces the number of times the Github API is called
to fetch the releases, and reorganizes the tests to remove duplication


maelk:
@fmuyassarov it looks like it is the wrong node that gets notReady. Is that something you've seen already ?

fmuyassarov:
> @fmuyassarov it looks like it is the wrong node that gets notReady. Is that something you've seen already ?

Yes, sorry for that. We missed to add `sort` in some places. Basically,` kubectl get nodes` doesn't print nodes in the right order, that's why it is failing. Here is the fix https://github.com/metal3-io/metal3-dev-env/pull/406

fmuyassarov:
@maelk would you mind to rebase your branch please?

maelk:
The CI should do it normally :/

fmuyassarov:
> The CI should do it normally :/

Hum, yes you are right.

fmuyassarov:
> The CI should do it normally :/

That's really weird that we are still seeing the same issue even after the fix.

maelk:
We found a bug in the features tests, @fmuyassarov will fix it. merging this one meanwhile

PR #405, opened 2020-07-30
Add small update feature test remediation task

smoshiur1237:
Use source cluster to fetch kubernetes nodes(masters and worker) name instead of accessing target cluster.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
@@ -29,11 +29,11 @@
     shell: "kubectl get secrets {{ CLUSTER_NAME }}-kubeconfig -n {{ NAMESPACE }} -o json | jq -r '.data.value'| base64 -d > /tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
 
   - name: Get the name of the worker node
-    shell: "kubectl get nodes --kubeconfig /tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml | grep none | awk '{print $1}' | sort"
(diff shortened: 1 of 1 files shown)


vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
```diff
@@ -29,11 +29,11 @@
     shell: "kubectl get secrets {{ CLUSTER_NAME }}-kubeconfig -n {{ NAMESPACE }} -o json | jq -r '.data.value'| base64 -d > /tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
 
   - name: Get the name of the worker node
-    shell: "kubectl get nodes --kubeconfig /tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml | grep none | awk '{print $1}'"
+    shell: "kubectl get m3m -n metal3 -o json | jq -r '.items[] | select (.metadata.labels."cluster.x-k8s.io/control-plane" == null) | .metadata.ownerReferences[] | select(.apiVersion=="cluster.x-k8s.io/v1alpha3").name'"
```
maelk, line 32: is this just to get the machine name, and not the metal3machine ? because you could directly get the machines, there should be the same annotation

smoshiur1237: I could get the machine name easily  but wasn't able to separate them as master and worker.. But using metal3machine, I was able to separate them as master and worker.

maelk: you should have the same annotation or label there, so you should be able to separate them the same way.

smoshiur1237: Done and updated.  thanks

PR #420, opened 2020-08-10
Update ironic images to locally built for v1alpha4 deployments

maelk:
There is a bug currently that prevents the ironic images to
be updated with the locally built image when deploying v1a4.
This fixes it.


maelk:
The Jenkins CI is currently broken, so we'll forcefully merge this PR to be able to properly test the ironic image fix

PR #421, opened 2020-08-10
Install clusterctl binary with curl

furkatgofurov7:
This patch moves clusterctl to 01_* and installs it using binary


01_prepare_host.sh
```diff
@@ -79,6 +79,12 @@ if ! command -v kustomize 2>/dev/null ; then
     rm "kustomize_${KUSTOMIZE_VERSION}_linux_amd64.tar.gz"
 fi
 
+if ! [ -x "$(command -v clusterctl)" ]; then
+  curl -L -O "https://github.com/kubernetes-sigs/cluster-api/releases/download/${CLUSTERCTL_VERSION}/clusterctl-linux-amd64 -o clusterctl"
```
fmuyassarov, line 83:
```suggestion
  curl -L -O "https://github.com/kubernetes-sigs/cluster-api/releases/download/${CLUSTERCTL_VERSION}/clusterctl-linux-amd64" -o clusterctl
```

furkatgofurov7: Fixed, thanks!

01_prepare_host.sh
```diff
@@ -79,6 +79,12 @@ if ! command -v kustomize 2>/dev/null ; then
     rm "kustomize_${KUSTOMIZE_VERSION}_linux_amd64.tar.gz"
 fi
 
+if ! [ -x "$(command -v clusterctl)" ]; then
+  curl -L -O "https://github.com/kubernetes-sigs/cluster-api/releases/download/${CLUSTERCTL_VERSION}/clusterctl-linux-amd64" -o clusterctl
```
fmuyassarov, line 83:
```suggestion
  curl -L -O https://github.com/kubernetes-sigs/cluster-api/releases/download/"${CLUSTERCTL_VERSION}"/clusterctl-linux-amd64 -o clusterctl
```
Maybe this will work?

lib/common.sh
```diff
@@ -178,9 +178,13 @@ else
   echo "Management cluster forced to be minikube when container runtime is not docker"
   export EPHEMERAL_CLUSTER="minikube"
 fi
+
 #Kustomize version
 export KUSTOMIZE_VERSION=${KUSTOMIZE_VERSION:-"v3.6.1"}
 
+#clusterctl version
+export CLUSTERCTL_VERSION=${CLUSTERCTL_VERSION:-"v0.3.8"}
```
maelk, line 186: You should probably use the CAPIRELEASE variable to get the latest CAPI release (you need to import lib/releases.sh)

01_prepare_host.sh
maelk: it might be slightly more readable with the quotes outside ?

lib/releases.sh
```diff
@@ -7,6 +7,9 @@ function get_latest_release() {
   else
     release="$(curl -H "Authorization: token ${GITHUB_TOKEN}" -sL "${1}")" || ( set -x && exit 1 )
   fi
+  if [ "${EPHEMERAL_CLUSTER}" == "minikube" ]; then
+    sudo yum install jq -y
+  fi
```
maelk, line 12: why is this here ? This file is only fetching the releases. if jq is not here, it means that it is sourced at the wrong place in 01_ . In addition, minikube != centos, you can run on Ubuntu with minikube

01_prepare_host.sh
```diff
@@ -79,6 +81,12 @@ if ! command -v kustomize 2>/dev/null ; then
     rm "kustomize_${KUSTOMIZE_VERSION}_linux_amd64.tar.gz"
 fi
 
+if ! [ -x "$(command -v clusterctl)" ]; then
+  curl -L https://github.com/kubernetes-sigs/cluster-api/releases/download/"${CAPIRELEASE}"/clusterctl-linux-amd64 -o clusterctl
+  chmod +x ./clusterctl
+  sudo mv ./clusterctl /usr/local/bin/clusterctl
+fi
```
maelk, line 88: what if CAPIRELEASE changes in between two runs ? you should probably check the version and if it is not in sync with the release version, upgrade.

fmuyassarov: Fixed, thanks.

lib/releases.sh
fmuyassarov: Fixed.

fmuyassarov:
As @furkatgofurov7 on leave right now, I will try to finalize the PR.

PR #422, opened 2020-08-11
Preload kind container image

fmuyassarov:
This patch downloads kind image in advance to minimize the time that CI takes to build Metal3-dev-env.
Currently we are attaching a volume to the VM on each CI run, which includes all the necessary packages that we normally install during 01_* script. Downloading kind image in 01_* script will make sure that the container image is already in the volume that we attach to the CI VM. As such, there will not be a need to download it again when executing `kind create cluster .... `command. Of course this path doesn't improve the time when running Metal3-dev-env locally and only meant for the CI improvement.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -57,6 +57,7 @@ elif [ "${EPHEMERAL_CLUSTER}" == "kind" ]; then
         curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/"${KIND_VERSION}"/kind-"$(uname)"-amd64
         chmod +x ./kind
         sudo mv kind /usr/local/bin/.
+        sudo "${CONTAINER_RUNTIME}" pull kindest/node:"${KUBERNETES_VERSION}"
     fi
 fi
 
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -199,7 +199,7 @@ else
   export CAPM3_IMAGE=${CAPM3_IMAGE:-"quay.io/metal3-io/cluster-api-provider-metal3:master"}
 fi
 
-#default hosts memory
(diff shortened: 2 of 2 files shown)


01_prepare_host.sh
```diff
@@ -57,6 +57,7 @@ elif [ "${EPHEMERAL_CLUSTER}" == "kind" ]; then
         curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/"${KIND_VERSION}"/kind-"$(uname)"-amd64
         chmod +x ./kind
         sudo mv kind /usr/local/bin/.
+        sudo "${CONTAINER_RUNTIME}" pull kindest/node:"${KUBERNETES_VERSION}"
```
maelk, line 60: This is the target cluster kubernetes version. I don't think we want to tie them together. depending on the kind version installed, it will probably use another image and hence download it again. you can either download the correct image based on the kind version, or give the --image parameter to take this specific image into use

fmuyassarov: I was thinking to create a new var `KIND_IMAGE_VERSION` but then I saw that we are already using `KUBERNETES_VERSION` var when running [kind create](https://github.com/metal3-io/metal3-dev-env/blob/b156fcafcf47b982c4cc5ad8c08d55d60d592b6e/03_launch_mgmt_cluster.sh#L191). Also, I have tested locally, and even if the version of the Kind image is not the one that is downloaded by default depending on the Kind version you still can run that image.
I'm fine with both option. I can create a new var or rely on ` KUBERNETES_VERSION`

maelk: oh, I didn't see that we are already forcing the image in the create, sorry. Then it is fine as is

PR #423, opened 2020-08-11
Fix incorrect ci tests and add verification

maelk:
Due to a typo, the tests were not properly running, this fixes it, and
adds a separate test verifying that the target cluster was properly
deployed before deprovisioning.


05_test.sh
```diff
@@ -3,6 +3,25 @@ set -xe
...
+
+"${METAL3_DIR}"/scripts/run.sh
+
+"${METAL3_DIR}"/scripts/fetch_manifests.sh
+
+kubectl get secrets "${CLUSTER_NAME}-kubeconfig" -n "${NAMESPACE}" -o json | jq -r '.data.value'| base64 -d > "/tmp/kubeconfig-${CLUSTER_NAME}.yaml"
+NUM_DEPLOYED_NODES="$(kubectl get nodes --kubeconfig "/tmp/kubeconfig-${CLUSTER_NAME}.yaml" | grep -c -w Ready)"
+process_status $? "Fetch number of deployed nodes"
+
+if [ "${NUM_DEPLOYED_NODES}" -ne "$((NUM_OF_MASTER_REPLICAS + NUM_OF_WORKER_REPLICAS))" ]; then
```
kashifest, line 20: why not use [`NUM_NODES`](https://github.com/metal3-io/metal3-dev-env/blob/master/lib/common.sh#L134) ?

maelk: Because it is not equivalent to the number of provisioned nodes. NUM_NODES is the number of vms created. That can be 4 for the feature tests, but we provision only 2 BMHs

fmuyassarov: Because it is not the actual var that would have any impact on the later scripts, but rather to make sure that we have correct number of nodes. Second. `NUM_NODES` var in not reachable from here since it is declared within an ansible script

maelk: there is a NUM_NODES defined in common.sh that is the number of VM that will be created as BMHs

fmuyassarov: Oh true. Didn't notice that.

maelk:
This is currently blocked by the CentOS tests not passing due to an issue with the CI ssh keys, but is required to fix the CI tests for Ubuntu at least. So we'll merge this manually to fix the tests.
/override test-centos-integration

PR #437, opened 2020-08-20
Fix symlink issue in monitor.keepalived.service in cloud-init

### LINKED ISSUE #436: monitor.keepalived.service fails to start during cloud-init in ubuntu target cluster
When the cloud-init runs for target cluster control plane in Ubuntu with focal version the following error is seen: 

```
[  158.848161] cloud-init[1944]: Failed to enable unit: File /etc/systemd/system/multi-user.target.wants/monitor.keepalived.service already exists.
[  158.863025] cloud-init[1944]: Failed to start monitor.keepalived.service: Unit monitor.keepalived.service not found.
```

Further investigation suggests that although the following line is in the output:
```
[  158.462347] cloud-init[1944]: Created symlink /etc/systemd/system/monitor.keepalived.service → /etc/systemd/system/multi-user.target.wants/monitor.keepalived.service.
```
But there is no actual symlink on the file

```
-rw-r--r--  1 root root  243 Aug 20 10:25 monitor.keepalived.service
```
Which is why the service not found error is seen.

kashifest:
Fixes #436

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_ubuntu.rc
+++ b/vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_ubuntu.rc
@@ -57,9 +57,8 @@ export CTLPLANE_KUBEADM_EXTRA_CONFIG="
       - systemctl enable --now docker kubelet
       - if (curl -sk --max-time 10 https://{{ CLUSTER_APIENDPOINT_HOST }}:6443/healthz); then echo \"keepalived already running\";else systemctl start keepalived; fi
       - usermod -aG docker {{ IMAGE_USERNAME }}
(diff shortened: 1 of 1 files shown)


Jaakko-Os:
Needs to be fixed on centos too, if needed

jan-est:
Looks good to me

jan-est:
> Needs to be fixed on centos too, if needed

Do we need to do something with Centos? I think this problem was caused by the upgrade of the Ubuntu version.

smoshiur1237 (commented):
Looks good to me

fmuyassarov (approved):

kashifest:
> Needs to be fixed on centos too, if needed

In centos, the qemu logs are not visible(#438 ), so not sure if the bug is there or not, if it is we can fix it later.

PR #439, opened 2020-08-21
Export CONTAINER_RUNTIME so subshells can use it

namnx228:
The container runtime used in the dev-env is decided by the environment variable CONTAINER_RUNTIME. By default, its value is Docker when the host is Ubuntu, and it is Podman when the host is Centos. This variable is set in lib/common.sh, and every script that sources this file has the right value of this variable. However, the dev-env also calls some scripts from other repos (e.g. Baremetal operator) that don't source the file, and they use the wrong container runtime. This leads to that when we run the dev-env on Ubuntu where Podman is not installed, but it tries to use Podman, resulting in an error.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -43,9 +43,9 @@ NODE_HOSTNAME_FORMAT=${NODE_HOSTNAME_FORMAT:-"node-%d"}
 OS=$(awk -F= '/^ID=/ { print $2 }' /etc/os-release | tr -d '"')
 # Container runtime
 if [[ "${OS}" == ubuntu ]]; then
-  CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-"docker"}
+  export CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-"docker"}
 else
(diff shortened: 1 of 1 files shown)


kashifest:
I think the CI explicitly exports `CONTAINER_RUNTIME` , so is this change gonna affect anything? https://github.com/metal3-io/project-infra/blob/0a813784186d1ea1812c2e83175da34c3e88e97f/jenkins/scripts/files/run_integration_tests.sh#L46

fmuyassarov:
> I think the CI explicitly exports `CONTAINER_RUNTIME` atleast for Ubuntu , so is this change gonna affect anything? https://github.com/metal3-io/project-infra/blob/0a813784186d1ea1812c2e83175da34c3e88e97f/jenkins/scripts/files/run_integration_tests.sh#L46

I think it won't have any affect. Because here we are first checking if the var is exported, and then if not we export it here.

namnx228:
@kashifest It doesn't affect the CI. This small fix may help new users to easier get the metal3-dev-env run.

Issue #440, opened 2020-08-24
Local registry cannot be reachable during metal3-dev-env deployment


smoshiur1237:
We are facing this issue more often related to local registry, when deploying metal3-dev-env. Here is the error code:
```
++ sudo docker inspect registry --format '{{.State.Status}}'
+ reg_state=exited
+ [[ exited == \e\x\i\t\e\d ]]
+ sudo docker start registry
registry
++ env
++ grep -v _LOCAL_IMAGE=
++ grep _IMAGE=
++ grep -o '^[^=]*'
+ for IMAGE_VAR in $(env | grep -v "_LOCAL_IMAGE=" | grep "_IMAGE=" | grep -o "^[^=]*")
+ IMAGE=docker.io/registry:latest
+ IMAGE_NAME=registry:latest
+ LOCAL_IMAGE=192.168.111.1:5000/localimages/registry:latest
+ sudo docker tag docker.io/registry:latest 192.168.111.1:5000/localimages/registry:latest
+ [[ docker == \p\o\d\m\a\n ]]
+ sudo docker push 192.168.111.1:5000/localimages/registry:latest
The push refers to repository [192.168.111.1:5000/localimages/registry]
Get http://192.168.111.1:5000/v2/: dial tcp 192.168.111.1:5000: connect: connection refused
make: *** [Makefile:7: configure_host] Error 1

```
If the local registry already exist in your local environment, during the Make process it goes to exited state. After that 02 script will start local registry, then begin to push images to local registry and cannot connect with the local registry.  The possible solution can be to have a delay in the script after starting the local registry. We haven't face this issue in CI run, as each CI run creates new VM, always have new deployment and the script creates new registry each time.

PR #441, opened 2020-08-24
Bug Fix: Local registry connection error during Metal3 deployment.

smoshiur1237:
This PR will fix the issue #440 . Here we have added a delay for 5s, when the local registry is in exited state and the script starts local registry. It solved the issue regarding the local registry connection error in local environment building. We have tested the script with different delays such as 30s, 20s, 10s and 5s. In each and every delays the deployment was running without any error and choose the minimum delay 5s for Metal3-dev-env.  The CI run will have also 5s delay because of this change.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -132,6 +132,7 @@ elif [[ "$reg_state" != "running" ]]; then
   sudo "${CONTAINER_RUNTIME}" rm registry -f || true
   sudo "${CONTAINER_RUNTIME}" run -d -p 5000:5000 --name registry "$DOCKER_REGISTRY_IMAGE"
 fi
+sleep 5
 
 # Pushing images to local registry
 for IMAGE_VAR in $(env | grep -v "_LOCAL_IMAGE=" | grep "_IMAGE=" | grep -o "^[^=]*") ; do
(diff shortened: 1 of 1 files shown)


fmuyassarov (approved):
Can you please add a small note why this is needed?
And wouldn't 10 or 15 or 20 secs be enough for the registry to be ready to accept images? I know it is still very small difference compared to 30 sec, but Metal3-dev-env already takes a lot of time to be ready, so we should try not to increase the time as much as we can.

smoshiur1237:
> Can you please add a small note why this is needed?
> And wouldn't 10 or 15 or 20 secs be enough for the registry to be ready to accept images? I know it is still very small difference compared to 30 sec, but Metal3-dev-env already takes a lot of time to be ready, so we should try not to increase the time as much as we can.

I will test tomorrow with different time interval to confirm the delays. 
/hold

Jaakko-Os:
Is this issue seen in both CI and local testing?
I've seen it several times in local testing. The delay could be skipped in the CI if the problem doesn't appear there?

smoshiur1237:
> Is this issue seen in both CI and local testing?
> I've seen it several times in local testing. The delay could be skipped in the CI if the problem doesn't appear there?

We are having this issue in local testing and when you do multiple time Make and Make clean . CI is not failing because it’s always have new environment and first time deployment.  I am going to test it with minimal delay and hope it will work with very short delay.

Jaakko-Os (commented):
Good work!

02_configure_host.sh
```diff
@@ -127,7 +127,8 @@ reg_state=$(sudo "$CONTAINER_RUNTIME" inspect registry --format  "{{.State.Statu
 
 # ubuntu_install_requirements.sh script restarts docker daemon which causes local registry container to be in exited state.
 if [[ "$reg_state" == "exited" ]]; then
-  sudo "${CONTAINER_RUNTIME}" start registry  
+  sudo "${CONTAINER_RUNTIME}" start registry
+  sleep 5
```
fmuyassarov, line 131: Should not it be outside of the condition as you did earlier? Because we might have the similar issue not only when we restart the registry but also when we start it initially?

smoshiur1237: This issue occurs only when we restart the registry. In new environment/VM it always create new registry and in this case will not have this issue. For example if you see all the CI run, we never face this issue. Also I have tried with new VM and I didn't face any error. So in CI we will not see any delays.

fmuyassarov:
/hold 
Needs to be checked if we can start registry container earlier so that there is enough time for the registry to be ready to accept images before we start pushing.

PR #445, opened 2020-09-01
Update Centos to use Metal3DataTemplate

furkatgofurov7:
Currently Centos setup does not use Metal3DataTemplate and that's why the node name is the same as BMH name. This patch will update a setup so that Centos will also use Metal3DataTemplate to override the node name.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml
@@ -48,7 +48,7 @@ spec:
         checksum: ${ IMAGE_CHECKSUM }
         checksumType: ${ IMAGE_CHECKSUM_TYPE }
         format: ${ IMAGE_FORMAT }
-{% if CAPM3_VERSION != "v1alpha3" and IMAGE_OS != "Centos" %}
+{% if CAPM3_VERSION != "v1alpha3" %}
       dataTemplate:
         name: ${ CLUSTER_NAME }-controlplane-template
 ---
--- a/vm-setup/roles/v1aX_integration_test/templates/cluster-template-workers.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/cluster-template-workers.yaml
@@ -52,7 +52,7 @@ spec:
         checksum: ${ IMAGE_CHECKSUM }
         checksumType: ${ IMAGE_CHECKSUM_TYPE }
         format: ${ IMAGE_FORMAT }
-{% if CAPM3_VERSION != "v1alpha3" and IMAGE_OS != "Centos" %}
+{% if CAPM3_VERSION != "v1alpha3" %}
       dataTemplate:
         name: ${ CLUSTER_NAME }-workers-template
 ---
--- a/vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
+++ b/vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
@@ -29,7 +29,9 @@ export CTLPLANE_KUBEADM_EXTRA_CONFIG="
       sshAuthorizedKeys:
       - {{ SSH_PUB_KEY_CONTENT }}
     preKubeadmCommands:
+{% if CAPM3_VERSION == "v1alpha3" %}
       - ifup eth1
+{% endif %}
       - dnf update -y
       - dnf install -y ebtables socat conntrack-tools
       - dnf install python3 -y
@@ -129,6 +131,7 @@ export CTLPLANE_KUBEADM_EXTRA_CONFIG="
                   {{ CLUSTER_APIENDPOINT_HOST }}
               }
           }
+{% if CAPM3_VERSION == "v1alpha3" %}
       - path: /etc/sysconfig/network-scripts/ifcfg-eth1
         owner: root:root
         permissions: '0644'
@@ -138,6 +141,7 @@ export CTLPLANE_KUBEADM_EXTRA_CONFIG="
           ONBOOT=yes
           TYPE=Ethernet
           USERCTL=no
+{% endif %}
       - path: /etc/yum.repos.d/kubernetes.repo
         owner: root:root
         permissions: '0644'
@@ -164,7 +168,9 @@ export WORKERS_KUBEADM_EXTRA_CONFIG="
         sshAuthorizedKeys:
         - {{ SSH_PUB_KEY_CONTENT }}
       preKubeadmCommands:
+{% if CAPM3_VERSION == "v1alpha3" %}
         - ifup eth1
+{% endif %}
         - dnf update -y
         - dnf install -y ebtables conntrack-tools socat
         - dnf install python3 -y
@@ -200,6 +206,7 @@ export WORKERS_KUBEADM_EXTRA_CONFIG="
               else
                 cat \"\${tmpfile}\"| sed '\$d' > \"\${dst}\";
               fi
+{% if CAPM3_VERSION == "v1alpha3" %}
         - path: /etc/sysconfig/network-scripts/ifcfg-eth1
           owner: root:root
           permissions: '0644'
@@ -209,6 +216,7 @@ export WORKERS_KUBEADM_EXTRA_CONFIG="
             ONBOOT=yes
             TYPE=Ethernet
             USERCTL=no
+{% endif %}
         - path: /etc/yum.repos.d/kubernetes.repo
           owner: root:root
           permissions: '0644'
--- a/vm-setup/roles/v1aX_integration_test/templates/cluster-template-cluster.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/cluster-template-cluster.yaml
@@ -37,7 +37,7 @@ spec:
     port: ${ API_ENDPOINT_PORT }
   noCloudProvider: true
 ---
-{% if CAPM3_VERSION != "v1alpha3" and IMAGE_OS != "Centos" %}
+{% if CAPM3_VERSION != "v1alpha3" %}
 apiVersion: ipam.metal3.io/v1alpha1
 kind: IPPool
 metadata:


maelk (changes requested):
you missed something in cluster-template-cluster.yaml

kashifest:
would it also need to change more things like https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml#L79 ????

furkatgofurov7:
> you missed something in cluster-template-cluster.yaml

Missed that, was updating it, thanks anyways.

furkatgofurov7:
> would it also need to change more things like https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml#L79 ????

@kashifest are you referring to `ethernets_id`, how is it related?

maelk (changes requested):
That will conflict with the network configuration that is done as part of the kubeadmconfig and KCP files. you need to remove those.

kashifest:
> > would it also need to change more things like https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml#L79 ????
> 
> @kashifest are you referring to `ethernets_id`, how is it related?

AFAIK, `enp1s0` is ubuntu specific

furkatgofurov7:
> > > would it also need to change more things like https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml#L79 ????
> > 
> > 
> > @kashifest are you referring to `ethernets_id`, how is it related?
> 
> AFAIK, `enp1s0` is ubuntu specific

`enp2s0` seems to be only ubuntu related too

kashifest:
> > > > would it also need to change more things like https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml#L79 ????
> > > 
> > > 
> > > @kashifest are you referring to `ethernets_id`, how is it related?
> > 
> > 
> > AFAIK, `enp1s0` is ubuntu specific
> 
> `enp2s0` seems to be only ubuntu related too

Yes, thats what I meant in my initial comment, you need to parameterize those variables also which are image specific IMO

maelk:
Actually not that one. https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml#L79 is the name of the interface set by cloud-init, it can be arbitrary. However, https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml#L81 is what is in the BMH status nics, so that might need to be adapted. You can of course change both if you wish, but the only required one is the second one.

vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml
```diff
@@ -78,11 +78,19 @@ spec:
         - type: phy
           id: enp1s0
           macAddress:
+{% if IMAGE_OS != "Centos" %}
             fromHostInterface: enp1s0
+{% else %}
+            fromHostInterface: eth1
```
maelk, line 84: that is surprising. I would expect it to be eth0 if it was using the eth* naming. However, you should consider the following: that interface name is coming from the introspection data. and introspection, whether you deploy Ubuntu or CentOS, is always done with IPA, so the naming will remain the same, so there should not be a need for this condition. The bottom line is that this should work without modifications, both for Ubuntu and CentOS.

maelk (changes requested):
You still need to modify the centos environment variables file, to remove the network configuration done there.

vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml
furkatgofurov7: Updated

vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml
```diff
@@ -78,7 +78,9 @@ spec:
         - type: phy
           id: enp1s0
           macAddress:
+{% if IMAGE_OS != "Centos" %}
             fromHostInterface: enp1s0
+{% endif %}
```
maelk, line 83: There should not be a condition here. you need the fromHostInterface in both cases

vm-setup/roles/v1aX_integration_test/templates/cluster-template-workers.yaml
```diff
@@ -82,7 +82,9 @@ spec:
         - type: phy
           id: enp1s0
           macAddress:
+{% if IMAGE_OS != "Centos" %}
             fromHostInterface: enp1s0
+{% endif %}
```
maelk, line 87: same here

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
```diff
@@ -101,25 +100,6 @@ export CTLPLANE_KUBEADM_EXTRA_CONFIG="
...
-          vrrp_instance VI_1 {
-              state MASTER
-              interface eth1
-              virtual_router_id 1
-              priority 101
-              advert_int 1
-              virtual_ipaddress {
-                  {{ CLUSTER_APIENDPOINT_HOST }}
-              }
-          }
```
maelk, line 113: Why is part of the keepalived config removed ?

maelk: also, be careful with the interface name here

maelk (changes requested):

vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml
furkatgofurov7: I am sorry, but it is a bit confusing now, is not `enp1s0` specific to Ubuntu as Kashif was mentioning earlier in the comments. So I was conditioning it to make it work only for specific OS.

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
furkatgofurov7: > also, be careful with the interface name here

What does it mean, should not it be `eth1`?

vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml
kashifest: what Mael is suggesting is `enp1s0` is not specific to Ubuntu but its coming from IPA, so ignore my comments

vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml
furkatgofurov7: > that is surprising. I would expect it to be eth0 if it was using the eth* naming. However, you should consider the following: that interface name is coming from the introspection data. and introspection, whether you deploy Ubuntu or CentOS, is always done with IPA, so the naming will remain the same, so there should not be a need for this condition. The bottom line is that this should work without modifications, both for Ubuntu and CentOS.

It means, there is no need for modification in [https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml#L81](https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml#L81) or any other `fromHostInterface` ?

vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml
furkatgofurov7: Thanks for clarification sir!

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
furkatgofurov7: Done, thanks.

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
```diff
@@ -103,23 +102,13 @@ export CTLPLANE_KUBEADM_EXTRA_CONFIG="
           }
           vrrp_instance VI_1 {
               state MASTER
-              interface eth1
```
kashifest, line 106: You should not remove it , rename it perhaps

furkatgofurov7: rename to `enp1s0` | `enp2s0` | `eth0` ? Most probably `enp1s0`..

maelk: You also need to keep the backwards compatibility for v1a3 without Metal3datatemplate

furkatgofurov7: Updated

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
```diff
@@ -121,14 +123,19 @@ export CTLPLANE_KUBEADM_EXTRA_CONFIG="
           }
           vrrp_instance VI_1 {
               state MASTER
+{% if CAPM3_VERSION == "v1alpha3" %}
               interface eth1
+{% else %}
+              interface enp1s0
```
furkatgofurov7, line 129: @maelk is current state of PR meets the requirement of backwards compatibility and also is interface name `enp1s0` should be used here? I was not able to find the root cause of failure except this interface naming might be a problem..

vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml
kashifest: @maelk the interface naming is not working properly, here are few outputs from cloud-init and the node itself: 
```
[   36.940155] virtio_net virtio1 enp2s0: renamed from eth1
[   36.982242] virtio_net virtio0 enp1s0: renamed from eth0
```

```
ci-info: ++++++++++++++++++++++++++++++++++++++Net device info+++++++++++++++++++++++++++++++++++++++
ci-info: +--------+------+-----------------------------+---------------+--------+-------------------+
ci-info: | Device |  Up  |           Address           |      Mask     | Scope  |     Hw-Address    |
ci-info: +--------+------+-----------------------------+---------------+--------+-------------------+
ci-info: |  eth0  | True | fe80::22c:a3ff:fef7:5384/64 |       .       |  link  | 00:2c:a3:f7:53:84 |
ci-info: |  eth1  | True |       192.168.111.100       | 255.255.255.0 | global | 00:2c:a3:f7:53:86 |
ci-info: |  eth1  | True | fe80::22c:a3ff:fef7:5386/64 |       .       |  link  | 00:2c:a3:f7:53:86 |
ci-info: |   lo   | True |          127.0.0.1          |   255.0.0.0   |  host  |         .         |
ci-info: |   lo   | True |           ::1/128           |       .       |  host  |         .         |
ci-info: +--------+------+-----------------------------+---------------+--------+-------------------+
ci-info: +++++++++++++++++++++++++++++++Route IPv4 info+++++++++++++++++++++++++++++++
ci-info: +-------+---------------+---------------+---------------+-----------+-------+
ci-info: | Route |  Destination  |    Gateway    |    Genmask    | Interface | Flags |
ci-info: +-------+---------------+---------------+---------------+-----------+-------+
ci-info: |   0   |    0.0.0.0    | 192.168.111.1 |    0.0.0.0    |    eth1   |   UG  |
ci-info: |   1   | 192.168.111.0 |    0.0.0.0    | 255.255.255.0 |    eth1   |   U   |
ci-info: +-------+---------------+---------------+---------------+-----------+-------+
ci-info: +++++++++++++++++++Route IPv6 info+++++++++++++++++++
ci-info: +-------+-------------+---------+-----------+-------+
ci-info: | Route | Destination | Gateway | Interface | Flags |
ci-info: +-------+-------------+---------+-----------+-------+
ci-info: |   1   |  fe80::/64  |    ::   |    eth0   |   U   |
ci-info: |   2   |  fe80::/64  |    ::   |    eth1   |   U   |
ci-info: |   4   |   ff00::/8  |    ::   |    eth0   |   U   |
ci-info: |   5   |   ff00::/8  |    ::   |    eth1   |   U   |
ci-info: +-------+-------------+---------+-----------+-------+
Cloud-init v. 18.5 running 'modules:config' at Fri, 04 Sep 2020 08:23:22 +0000. Up 58.79 seconds.
Error: Connection activation failed: No suitable device found for this connection (device eth0 not available because profile is not compatible with device (mismatching interface name)).
```

```
metal3@test1-lbtbj ~]$ ip link show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 00:2c:a3:f7:53:84 brd ff:ff:ff:ff:ff:ff
3: eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 00:2c:a3:f7:53:86 brd ff:ff:ff:ff:ff:ff
4: docker0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN mode DEFAULT group default 
    link/ether 02:42:8f:ce:8d:75 brd ff:ff:ff:ff:ff:ff
[metal3@test1-lbtbj ~]$ nmcli device status
DEVICE   TYPE      STATE      CONNECTION  
eth1     ethernet  connected  System eth1 
docker0  bridge    connected  docker0     
eth0     ethernet  connected  System eth0 
lo       loopback  unmanaged  --          
[metal3@test1-lbtbj ~]$ netstat -i
Kernel Interface table
Iface             MTU    RX-OK RX-ERR RX-DRP RX-OVR    TX-OK TX-ERR TX-DRP TX-OVR Flg
docker0          1500        0      0      0 0             0      0      0      0 BMU
eth0             1500     2856      0      0 0            23      0      0      0 BMRU
eth1             1500    44725      0    451 0         22800      0      0      0 BMRU
lo              65536    61764      0      0 0         61764      0      0      0 LRU

```
Are you sure interface names coming from introspection would override that coming from the image? Or then what is done wrong here because we can see that cloud-init says interface is renamed but that is not reflected in the node itself.

furkatgofurov7: Yes, it is not overriding it I guess or the conditioning is not in the right place

furkatgofurov7:
@maelk @kashifest can you please review it?

Issue #353, opened 2020-06-24
Latest ansible versions should be used in metal3-dev-env 


kashifest:
Currentlly, in ubuntu based deployments ansible version is fixed to [2.9.1](https://github.com/metal3-io/metal3-dev-env/blob/master/ubuntu_install_requirements.sh#L46) . In case of Centos, it was latest, but since version 2.9.9, the following issue is faced 

```
TASK [v1aX_integration_test : Download Calico manifest] ************************
[1;30mtask path: /home/****/tested_repo/vm-setup/roles/v1aX_integration_test/tasks/verify.yml:81[0m
[0;34m<192.168.111.249> ESTABLISH SSH CONNECTION FOR USER: metal3[0m
[0;34m<192.168.111.249> SSH: EXEC ssh -C -o ControlMaster=auto -o ControlPersist=60s -o StrictHostKeyChecking=no -o 'IdentityFile="/home/****/.ssh/id_rsa"' -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o 'User="metal3"' -o ConnectTimeout=10 -o ControlPath=/home/****/.ansible/cp/da21a37a76 192.168.111.249 '/bin/sh -c '"'"'echo ~metal3 && sleep 0'"'"''[0m
[0;34m<192.168.111.249> (0, b'/home/metal3\n', b'')[0m
[0;34m<192.168.111.249> ESTABLISH SSH CONNECTION FOR USER: metal3[0m
[...]

PR #447, opened 2020-09-06
Bump ansible to latest

fmuyassarov:
Fixes https://github.com/metal3-io/metal3-dev-env/issues/353

### DIFF
--- a/centos_install_requirements.sh
+++ b/centos_install_requirements.sh
@@ -41,7 +41,7 @@ sudo dnf -y install \
   python3-pip \
   wget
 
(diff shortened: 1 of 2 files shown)


fmuyassarov:
/test-integration
test-centos-integration

maelk (approved):
so the issue with the user was a glitch in a specific version ?

fmuyassarov:
> so the issue with the user was a glitch in a specific version ?

Yes, it seems so. Latest version is 2.9.13 and apparently we don't have that problem anymore.

Issue #446, opened 2020-09-03
Upgrade fails due to duplicate control plane machine 


Jaakko-Os:
Upgrade test case [1cp_3w_bootDiskImage_scaleInWorkers_upgrade_both.sh](https://github.com/Nordix/metal3-dev-env/blob/master/scripts/feature_tests/upgrade/workers_upgrade/1cp_3w_bootDiskImage_scaleInWorkers_upgrade_both.sh) fails.

Reason:
- during control plane upgrade, an extra machine (test1-g5ncz) is encountered, only one is expected

```
$ kubectl get bmh -A
NAMESPACE   NAME     STATUS   PROVISIONING STATUS   CONSUMER                             BMC                                                                                         HARDWARE PROFILE   ONLINE   ERROR
metal3      node-0   OK       provisioned           test1-new-controlplane-image-gf8rj   ipmi://192.168.111.1:6230                                                                   unknown            true     
metal3      node-1   OK       provisioned           test1-controlplane-94r8d             redfish+http://192.168.111.1:8000/redfish/v1/Systems/5b8f7840-eeb6-4e26-8896-607e363fa89b   unknown            true     
metal3      node-2   OK       provisioned           test1-new-workers-image-9j9sr        ipmi://192.168.111.1:6232                                                                   unknown            true     
metal3      node-3   OK       provisioned           test1-new-workers-image-5jsgn        redfish+http://192.168.111.1:8000/redfish/v1/Systems/8e20cc8a-7b7e-483b-aa33-efbf8ddaf357   unknown            true
$ kubectl get m3m -A
NAMESPACE   NAME                                 PROVIDERID                                      READY   CLUSTER   PHASE
metal3      test1-controlplane-94r8d             metal3://ac6ef08b-6401-492a-8fd1-eba303f7b91b   true    test1     
metal3      test1-new-controlplane-image-gf8rj   metal3://49fa9a13-d05a-4d35-b7fb-80024da0b219   true    test1     
metal3      test1-new-controlplane-image-qq2jp                                                           test1     
metal3      test1-new-workers-image-5jsgn        metal3://1d75e3fb-2517-4d51-97d1-8b3065496f73   true    test1     
metal3      test1-new-workers-image-9j9sr        metal3://f2350717-a6b6-476e-84fc-6422ed4f24b3   true    test1     
metal3      test1-new-workers-image-nnrbj                                                                test1
$ kubectl get machines -A -owide
[...]

PR #448, opened 2020-09-07
CI: improve upgrade test coverage

Jaakko-Os:
Fixes: #446

### DIFF
--- a/scripts/feature_tests/upgrade/clean_setup_env.sh
+++ b/scripts/feature_tests/upgrade/clean_setup_env.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+METAL3_DEV_ENV_DIR="$(dirname "$(readlink -f "${0}")")/../../../"
+
+# Deploy a fresh metal3-dev-env after each test case
+# to overcome environmental flakiness
+pushd "${METAL3_DEV_ENV_DIR}" || exit
+make clean
+make setup_env
+rc=$?
+if [ "${rc}" -ne 0 ]; then
+    echo "Metal3-dev-env setup failed, try again"
+    make clean
+    make setup_env
+fi
+popd || exit
--- a/scripts/feature_tests/upgrade/upgrade.sh
+++ b/scripts/feature_tests/upgrade/upgrade.sh
@@ -34,25 +34,21 @@ pushd "${METAL3_DEV_ENV_DIR}/scripts/feature_tests/upgrade" || exit
 source 1cp_1w_bootDiskImage_cluster_upgrade.sh
 popd || exit
 
-# Deploy a fresh metal3-dev-env after each test case
-# to overcome environmental flakiness
-pushd "${METAL3_DEV_ENV_DIR}" || exit
-make clean
-make setup_env
+pushd "${METAL3_DEV_ENV_DIR}/scripts/feature_tests/upgrade" || exit
+# shellcheck disable=SC1091
+source clean_setup_env.sh
 popd || exit
 
 # Run worker upgrade cases
-#pushd "${METAL3_DEV_ENV_DIR}/scripts/feature_tests/upgrade/workers_upgrade" || exit
+pushd "${METAL3_DEV_ENV_DIR}/scripts/feature_tests/upgrade/workers_upgrade" || exit
 # shellcheck disable=SC1091
-#source 1cp_3w_bootDiskImage_scaleInWorkers_upgrade_both.sh
-#popd || exit
+source 1cp_3w_bootDiskImage_scaleInWorkers_upgrade_both.sh
+popd || exit
 
-# Deploy a fresh metal3-dev-env after each test case
-# to overcome environmental flakiness
-#pushd "${METAL3_DEV_ENV_DIR}" || exit
-#make clean
-#make setup_env
-#popd || exit
+pushd "${METAL3_DEV_ENV_DIR}/scripts/feature_tests/upgrade" || exit
+# shellcheck disable=SC1091
+source clean_setup_env.sh
+popd || exit
 
 # Run controlplane upgrade tests
 pushd "${METAL3_DEV_ENV_DIR}/scripts/feature_tests/upgrade/controlplane_upgrade" || exit
--- a/scripts/feature_tests/upgrade/upgrade_common.sh
+++ b/scripts/feature_tests/upgrade/upgrade_common.sh
@@ -214,7 +214,7 @@ function controlplane_has_correct_replicas() {
         cp_replicas=$(kubectl --kubeconfig=/tmp/kubeconfig-"${CLUSTER_NAME}".yaml get nodes |
             awk 'NR>1' | grep -c master)
         if [[ "${cp_replicas}" == "${replicas}" ]]; then
-            echo "Successfully provisioned controlplane replica nodes: ${CP_NODE_NAME}"
+            echo "Successfully provisioned controlplane replica nodes"
             break
         else
             echo -n "+"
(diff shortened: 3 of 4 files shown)


scripts/feature_tests/upgrade/clean_setup_env.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+METAL3_DEV_ENV_DIR="$(dirname "$(readlink -f "${0}")")/../../../"
+
+# Deploy a fresh metal3-dev-env after each test case
+# to overcome environmental flakiness
+pushd "${METAL3_DEV_ENV_DIR}" || exit
+make clean
```
fmuyassarov, line 8: Maybe you could put this in a single loop?

scripts/feature_tests/upgrade/upgrade.sh
```diff
@@ -34,25 +34,21 @@ pushd "${METAL3_DEV_ENV_DIR}/scripts/feature_tests/upgrade" || exit
 source 1cp_1w_bootDiskImage_cluster_upgrade.sh
 popd || exit
 
-# Deploy a fresh metal3-dev-env after each test case
-# to overcome environmental flakiness
-pushd "${METAL3_DEV_ENV_DIR}" || exit
-make clean
-make setup_env
+pushd "${METAL3_DEV_ENV_DIR}/scripts/feature_tests/upgrade" || exit
```
fmuyassarov, line 37: Would it be possible to source it form by specifying absolute path to the `clean_setup_env.sh`?

scripts/feature_tests/upgrade/clean_setup_env.sh
Jaakko-Os: True. This was just a workaround to re-run 'make' once in case of failure.

I think it should be discussed separately how to handle 'make' problems since those impact the whole CI. The logic should be put to some common directory then.

Those are the reasons why I would not change this workaround.

scripts/feature_tests/upgrade/upgrade.sh
Jaakko-Os: can be ubuntu or centos so the relative path is more flexible

fmuyassarov: My point was that we could do  `export "${METAL3_DEV_ENV_DIR}/scripts/feature_tests/upgrade/clean_setup_env.sh"`. Os is there something that `clean_setup_env.sh` will need from /upgrade folder?

scripts/feature_tests/upgrade/clean_setup_env.sh
fmuyassarov: > True. This was just a workaround to re-run 'make' once in case of failure.
> 
> I think it should be discussed separately how to handle 'make' problems since those impact the whole CI. The logic should be put to some common directory then.

But whole CI, do you mean upgrade CI jobs?
> Those are the reasons why I would not change this workaround.

Jaakko-Os: No, I mean the airship CI. It's rather often that a CI job needs re-triggering manually due to an environment deployment failure.

scripts/feature_tests/upgrade/upgrade.sh
Jaakko-Os: Ok. Referring to https://github.com/metal3-io/metal3-dev-env/pull/448/files/80adc0d4f47056f987b17f5994d6a8f1f1e174d1#r485412338.
clean_setup_env.sh doesn't need anything from upgrade folder. The clean_setup_env.sh script should be moved to some common directory if it is seen to be beneficial for airship CI jobs.

fmuyassarov (commented):
/lgtm

scripts/feature_tests/upgrade/upgrade_common.sh
```diff
@@ -209,12 +209,9 @@ function controlplane_has_correct_replicas() {
     replicas="${1}"
     echo "Waiting for all replicas of controlplane node"
     for i in {1..3600}; do
-        kubectl get secrets "${CLUSTER_NAME}"-kubeconfig -n "${NAMESPACE}" -o json | \
-          jq -r '.data.value'| base64 -d > /tmp/kubeconfig-"${CLUSTER_NAME}".yaml
-        cp_replicas=$(kubectl --kubeconfig=/tmp/kubeconfig-"${CLUSTER_NAME}".yaml get nodes |
-            awk 'NR>1' | grep -c master)
+        cp_replicas=$(kubectl get bmh -n metal3 | grep -i provisioned | grep -c controlplane)
```
maelk, line 212: this means a change in the functionality of the test. bmh provisioned != node joined in target cluster. as it was before, it was checking that the node actually joined the cluster. by doing this change, you would miss all the errors that might happen once the node is booted but not joined yet (package installation error, kubeadm error etc.) I would not change this condition.

scripts/feature_tests/upgrade/upgrade_common.sh
```diff
@@ -237,23 +234,26 @@ function worker_has_correct_replicas() {
...
-    for i in {1..3600}; do
+    for i in {1..1800}; do
+        wr_replicas=$(kubectl get bmh -n metal3 | grep -i provisioned | grep -c worker)
         if [[ "${replicas}" -eq 0 ]]; then
-            wr_replicas=$(kubectl get bmh -n metal3 | grep -c worker)
-        else
-            kubectl get secrets "${CLUSTER_NAME}"-kubeconfig -n "${NAMESPACE}" -o json | \
-            jq -r '.data.value'| base64 -d > /tmp/kubeconfig-"${CLUSTER_NAME}".yaml
-            wr_replicas=$(kubectl --kubeconfig=/tmp/kubeconfig-"${CLUSTER_NAME}".yaml get nodes |
-                awk 'NR>1' | grep -vc master)
```
maelk, line 247: same thing here

scripts/feature_tests/upgrade/upgrade_common.sh
```diff
@@ -237,23 +234,26 @@ function worker_has_correct_replicas() {
...
+                echo "Expected worker replicas have left the cluster"
+                break
         fi
-        if [[ "${wr_replicas}" == "${replicas}" ]]; then
-            echo "Expected worker replicas have joined the cluster"
-            break
+        if [[ "${wr_replicas}" -eq "${replicas}" ]]; then
+          for ind in {1..1800}; do
+                running_machines=$(kubectl get m3m -A  | awk 'NR>1' | grep worker | grep -c "metal3://")
```
maelk, line 245: this is problematic. the machine having a provider ID does not mean that it joined the cluster. It means that it was only successfully provisioned (if nocloudprovider=false). However, if you want to do this check in the management cluster, you could do it by verifying that the machine (not the m3m) is in running state. That would mean that the node joined properly.

scripts/feature_tests/upgrade/upgrade_common.sh
Jaakko-Os: change reverted

scripts/feature_tests/upgrade/upgrade_common.sh
Jaakko-Os: When scaling to zero, the target cluster worker node deletion has failed very rarely. That's why the scale to zero condition is kept separate.
When scaling to >=1, the management cluster status is checked first to avoid 100+ times writing the kubeconfig to file. Target cluster check is done after the management cluster check is met.

scripts/feature_tests/upgrade/upgrade_common.sh
Jaakko-Os: When scaling to zero, the target cluster worker node deletion has failed very rarely. That's why the scale to zero condition is kept separate.
When scaling to >=1, the management cluster status is checked first to avoid 100+ times writing the kubeconfig to file. Target cluster check is done after the management cluster check is met.

maelk (approved):

PR #451, opened 2020-09-08
Create clusterctl local folder with overrides

maelk:
independently of the ephemeral cluster, to permit deploying
the workload cluster using the scripts even in a tilt setup.

Without this, the deployment scripts fail due to a missing folder

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -63,11 +63,15 @@ function clone_repos() {
 
 function patch_clusterctl(){
   pushd "${CAPM3PATH}"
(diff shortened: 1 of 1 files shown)


fmuyassarov (commented):
Does tilt based environment need whatever is in `path_clustterctl()`? If not, maybe we should just take L66 & L67 outside of the `path_clusterctl()` and keep `calling patch_clusterctl()` as it is done currently inside the `launch_cluster_api_provider_metal3()`?

maelk:
I think we need what is in release_manifests, that's where the templates are generated. I'd maybe keep running the whole function to make sure we do not break anything with clusterctl.
/test-integration

PR #452, opened 2020-09-11
Fetch pre-Baked Ubuntu node image 

kashifest:
This PR fetches pre-baked ubuntu node image with required packages from nordix artifactory.
As such, this PR also removes the redundant prekubeadm commands.

### DIFF
--- a/lib/images.sh
+++ b/lib/images.sh
@@ -3,8 +3,8 @@
 # Image url and checksum
 IMAGE_OS=${IMAGE_OS:-Centos}
 if [[ "${IMAGE_OS}" == "Ubuntu" ]]; then
-  export IMAGE_NAME=${IMAGE_NAME:-focal-server-cloudimg-amd64.img}
-  export IMAGE_LOCATION=${IMAGE_LOCATION:-https://cloud-images.ubuntu.com/focal/current}
+  export IMAGE_NAME=${IMAGE_NAME:-UBUNTU_NODE_IMAGE_K8S_v1.18.8.qcow2}
+  export IMAGE_LOCATION=${IMAGE_LOCATION:-https://artifactory.nordix.org/artifactory/airship/images/k8s_v1.18.8/}
 elif [[ "${IMAGE_OS}" == "FCOS" ]]; then
   export IMAGE_NAME=${IMAGE_NAME:-fedora-coreos-30.20191014.0-openstack.x86_64.qcow2}
   export IMAGE_LOCATION=${IMAGE_LOCATION:-https://builds.coreos.fedoraproject.org/prod/streams/testing/builds/30.20191014.0/x86_64/}
--- a/vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_ubuntu.rc
+++ b/vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_ubuntu.rc
@@ -33,27 +33,7 @@ export CTLPLANE_KUBEADM_EXTRA_CONFIG="
       - ip link set dev enp2s0 up
       - dhclient enp2s0
 {% endif %}
-      - apt update -y
-      - apt-get install -y conntrack socat
       - netplan apply
-      - >-
-        apt install net-tools gcc linux-headers-$(uname -r) bridge-utils
-        apt-transport-https ca-certificates curl gnupg-agent
-        software-properties-common -y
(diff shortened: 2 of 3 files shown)


sumitjadhav1:
@kashifest good approach indeed. +1

maelk:
@kashifest did you change the building method of the image or is it still with a 50GB disk ?

kashifest:
> @kashifest did you change the building method of the image or is it still with a 50GB disk ?

Its 20GB now

lib/images.sh
```diff
@@ -3,8 +3,8 @@
 # Image url and checksum
 IMAGE_OS=${IMAGE_OS:-Centos}
 if [[ "${IMAGE_OS}" == "Ubuntu" ]]; then
-  export IMAGE_NAME=${IMAGE_NAME:-focal-server-cloudimg-amd64.img}
-  export IMAGE_LOCATION=${IMAGE_LOCATION:-https://cloud-images.ubuntu.com/focal/current}
+  export IMAGE_NAME=${IMAGE_NAME:-UBUNTU_NODE_IMAGE_K8S_v1.18.8.qcow2}
```
fmuyassarov, line 6: These need to be changed in https://github.com/metal3-io/metal3-dev-env/blob/ae23c93cc05163e38c95e99580fdf7d35e2386d8/vm-setup/roles/v1aX_integration_test/vars/main.yml#L57 & https://github.com/metal3-io/metal3-dev-env/blob/ae23c93cc05163e38c95e99580fdf7d35e2386d8/vm-setup/roles/v1aX_integration_test/vars/main.yml#L59as well I guess?

kashifest: Hmm.Thats interesting, I dont need to change https://github.com/metal3-io/metal3-dev-env/blob/ae23c93cc05163e38c95e99580fdf7d35e2386d8/vm-setup/roles/v1aX_integration_test/vars/main.yml#L57 since it taking from the ennvironment variable first, but why is the image location hard coded in https://github.com/metal3-io/metal3-dev-env/blob/ae23c93cc05163e38c95e99580fdf7d35e2386d8/vm-setup/roles/v1aX_integration_test/vars/main.yml#L59 and not defaulted?

fmuyassarov: TBH, I'm not sure how did it happen not to have a lookup. 

It is true that lookup in IMAGE_NAME will make sure to download the image from Nordix, but still I think it is better to have the same default values everywhere for the same var, wdyt?

Jaakko-Os:
please make sure that 'test-upgrade-features' tests pass before merging

kashifest:
> please make sure that 'test-upgrade-features' tests pass before merging

Why do you think that is necessary? How is this related to upgrade?

lib/images.sh
kashifest: Changed, thanks

Jaakko-Os:
> > please make sure that 'test-upgrade-features' tests pass before merging
> 
> Why do you think that is necessary? How is this related to upgrade?

since the PR is modifying IMAGE_NAME. If the raw image naming and generation is not impacted, then no need for upgrade test.

kashifest:
> > > please make sure that 'test-upgrade-features' tests pass before merging
> > 
> > 
> > Why do you think that is necessary? How is this related to upgrade?
> 
> since the PR is modifying IMAGE_NAME. If the raw image naming and generation is not impacted, then no need for upgrade test.

Well the changes are visible here, raw images are not touched. Since this is a metal3-dev-env enhancement, so you would be good as long as your variable declarations are in sync with metal3-dev-env.

fmuyassarov (commented):
/lgtm

maelk:
/test-integration
/test-centos-integration
lgtm, waiting for the tests

lib/images.sh
maelk, line 6: Could you maybe update the image to change thename ? We know that it is k8s v1.18.8 since the folder is named that way. but which Ubuntu version is it ? that might also be important to know.

maelk (approved):

maelk:
I removed the lgtm to prevent the merging, but feel free to put it again if the naming change would not be happening now.

lib/images.sh
kashifest: Yeah that can be added. I ll do that in a follow up PR if thats ok

fmuyassarov (commented):
/lgtm

PR #455, opened 2020-09-15
Move upgrade tests to the target cluster

Xenwar:
Moving upgrade tests so that they are done after pivoting. This requires changing the tests in two ways.
- Remove provisioning steps as the CP and worker nodes already provisioned before pivoting.
- Use `kubeconfig` of the target cluster for running `kubectl `and `clusterctl` commands.

### DIFF
--- a/scripts/feature_tests/upgrade/1cp_1w_bootDiskImage_cluster_upgrade.sh
+++ b/scripts/feature_tests/upgrade/1cp_1w_bootDiskImage_cluster_upgrade.sh
@@ -12,9 +12,14 @@ echo '' >~/.ssh/known_hosts
 start_logging "${1}"
 # Provision original nodes
 set_number_of_master_node_replicas 1
-set_number_of_worker_node_replicas 1
+set_number_of_worker_node_replicas 3 # Temporary solution for moving all BMHs
 
 provision_controlplane_node
+
+# Get kubeconfig before pivoting | will be overriden when pivoting
+# Relevant when re-using the cluster for successive tests
+kubectl get secrets "${CLUSTER_NAME}"-kubeconfig -n "${NAMESPACE}" -o json | jq -r '.data.value'| base64 -d > /tmp/kubeconfig-"${CLUSTER_NAME}".yaml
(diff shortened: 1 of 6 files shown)


maelk (commented):
It looks like this rely on the tests running in a specific order (i.e. after the pivoting tests), is it correct ? And all the cases are run one ofter another, so there could be interferences between tests. Maybe that should be documented ?

Xenwar:
> It looks like this rely on the tests running in a specific order (i.e. after the pivoting tests), is it correct ? And all the cases are run one ofter another, so there could be interferences between tests. Maybe that should be documented ?

Yes, it does. 
According to @kanwar-saad All cases are to be ignored except after pivoting. 
Which is the reason why we are moving them. In fact, the tests are not usable on their own any longer. 
If the requirements has changed, please do let me know.

maelk:
Sorry I was unclear. I meant that previously the tests were isolated, i.e. you provision a cluster for each of them. Now, those tests use the same cluster that has been previously set up and pivoted to. My question was whether this approach should be used, or the old one, where we would provision a cluster and pivot for each test cases individually, to prevent interference, as you had it before.

Xenwar:
> Sorry I was unclear. I meant that previously the tests were isolated, i.e. you provision a cluster for each of them. Now, those tests use the same cluster that has been previously set up and pivoted to. My question was whether this approach should be used, or the old one, where we would provision a cluster and pivot for each test cases individually, to prevent interference, as you had it before.

I agree. Failed to see that part. 
Should we do provision -> pivot for each test case ?

maelk:
That's time consumed vs. resiliency of the tests kind of question. Do you have an idea of how long tests would take in both cases ?

Xenwar:
I am guessing, 4 cases would take 12 hours max.
And, I agree it would make sense to do use a separate cluster for each test case. That way, it would be easy to reason out failures.

maelk:
TBH, here I don't really have a strong opinion. Considering that the upgrade cases are just validation, we could prioritize speed, but then if we have to debug an interference one day, we'll be angry at ourselves for that decision.

Xenwar:
This PR is being broken into smaller PRs and no longer needed.

PR #456, opened 2020-09-16
Update images before running ironic_run_local.sh

maelk:
Currently, since we skip the deployment of BMO in v1alpha4,
the images are not updated, so ironic and inspector images
are never actually tested. This fixes the issue.


namnx228 (approved):

fmuyassarov (commented):
/lgtm

PR #460, opened 2020-09-17
Update var doc with CAPI_VERSION var

fmuyassarov:
Add var description in the doc.


fmuyassarov:
@Jaakko-Os How would upgrade tests be affected by this change? Because I can see that there are couple of places where you using `CAPI_VERSION` var but I wasn't really sure who is the consumer of that in upgrade scripts.

Jaakko-Os:
> @Jaakko-Os How would upgrade tests be affected by this change? Because I can see that there are couple of places where you are using `CAPI_VERSION` var but I wasn't really sure who is the consumer of that in upgrade scripts.

@fmuyassarov The upgrade disk image tests use CAPI_VERSION when the 'metal3MachineTemplate' is generated. I suppose new CAPI_VERSIONs are introduced at some point of time?
Hard coding items would also decrease the maintainability.

kashifest:
I am thinking since CAPM3 has already diverged from CAPI with regards to api version, in near future we will have CAPI v1alpha4 but CAPM3 v1alpha5 or X, and then would we need this variable again? or would we always use CAPI's latest version. If the case is the later, I am ok with this change. If the case is the former, then we should rather keep this variable and adapt the use of CAPI_VERSION more appropriately like here https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml#L2

maelk:
I think we should keep this variable, for the reason Kashif listed. CAPI will come with v1alpha4, and we will at that point need to maintain support for both versions. Setting CAPM3 to be v1a4 by default is good thing though

fmuyassarov:
@kashifest , @maelk  I was thinking that we could change that (have a CAPI_VERSION) once CAPI releases new version which would be the end of the year If I'm not wrong. But I agree that we will anyways need it. Now I have reverted it and updated the vars.md since it was missing the var in there and causing some assumptions that it could have v1alpha4 as well.

Jaakko-Os:
Please update the PR headline to reflect the change

fmuyassarov:
> Please update the PR headline to reflect the change

done.

kashifest:
You can keep CAPM3 to be v1a4 by default as pointed by Mael

fmuyassarov:
> You can keep CAPM3 to be v1a4 by default as pointed by Mael

Ops, missed. Sorry.

maelk (approved):

PR #463, opened 2020-09-18
Parameterize external_subnet_v4

fmuyassarov:
Fixes https://github.com/metal3-io/metal3-dev-env/issues/462


fmuyassarov:
It seems we have a common issue in the Centos CI. I'll try to debug it.

PR #464, opened 2020-09-20
Disable Basic authentication and CA certs in BMO for Ironic

maelk:
This is to maintain compatibility for the TLS and basic auth
changes to come in BMO and CAPM3.

/cc @fmuyassarov 
/cc @kashifest


fmuyassarov (commented):
/lgtm

PR #465, opened 2020-09-21
Add usr/local/go/bin to the PATH env var

fmuyassarov:
Set `/usr/local/go/bin` to the PATH environment variable so that we don't have to explicitly export the PATH every time we use the golang in the Metal3-dev-env.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -33,6 +33,14 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
   -i vm-setup/inventory.ini \
   -b -vvv vm-setup/install-package-playbook.yml
 
+# Add usr/local/go/bin to the PATH environment variable
+GOBINARY="/usr/local/go/bin"
+if [[ ":$PATH:" != *":$GOBINARY:"* ]]; then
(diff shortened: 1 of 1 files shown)


smoshiur1237:
Looks good to me

01_prepare_host.sh
```diff
@@ -33,6 +33,11 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
   -i vm-setup/inventory.ini \
   -b -vvv vm-setup/install-package-playbook.yml
 
+# Add usr/local/go/bin to the PATH environment variable
+echo "export PATH=$PATH:/usr/local/go/bin" >> ~/.bashrc
+# shellcheck disable=SC1090
+source ~/.bashrc
+
```
maelk, line 40: you'll end up with as many lines like this as you run this script. Please first check that the line is not present before adding it. Also,there's no need to source the file here, considering that the variables are already properly set when running this script. So only line 37 (with a condition) would be needed

maelk (changes requested):

01_prepare_host.sh
fmuyassarov: Good catch, should be fixed now. Thanks

Jaakko-Os:
looks good!

PR #461, opened 2020-09-18
Decouple pivot and re-pivot actions

kashifest:
This PR separates pivoting and re-pivoting into separate actions.

PR #469, opened 2020-09-22
Introduce 'repivoting' action

kashifest:
Changes introduced in #461 was still performing pivoting and repivoting using the same actions `pivoting`. Upgrade feature test requires now to have a flexibility when to repivot and not do the repivot only when `pivoting` action is called. This PR introduces a new action and make target named `repivoting`.

### DIFF
--- a/scripts/feature_tests/pivoting/Makefile
+++ b/scripts/feature_tests/pivoting/Makefile
@@ -2,13 +2,16 @@ M3PATH := "$(dirname "$(readlink -f "${0}")")../../../"
 export NUM_OF_MASTER_REPLICAS := 1
 export NUM_OF_WORKER_REPLICAS := 1
 
-all: provision pivoting deprovision
+all: provision pivoting repivoting deprovision
 
 provision:
 	./../feature_test_provisioning.sh
 
 pivoting:
 	./pivot.sh
+
+repivoting:
+	./repivot.sh
 
 deprovision:
 	./../feature_test_deprovisioning.sh
--- a/scripts/feature_tests/pivoting/repivot.sh
+++ b/scripts/feature_tests/pivoting/repivot.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+METAL3_DIR="$(dirname "$(readlink -f "${0}")")/../../.."
+
+export ACTION="repivoting"
+
+"${METAL3_DIR}"/scripts/run.sh
--- a/vm-setup/roles/v1aX_integration_test/tasks/main.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/main.yml
@@ -51,11 +51,11 @@
 
 - name: pivot
   include: move.yml
-  when: v1aX_integration_test_action == "pivoting" or v1aX_integration_test_action == "upgrading"
+  when: v1aX_integration_test_action in pivot_actions
 
 - name: repivot
   include: move_back.yml
-  when: v1aX_integration_test_action == "pivoting"
+  when: v1aX_integration_test_action == "repivoting"
 
 - name: Deprovision worker nodes
   k8s:
--- a/vm-setup/roles/v1aX_integration_test/tasks/move.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/move.yml
@@ -1,4 +1,8 @@
 ---
+  - name: Define number of BMH's
+    set_fact:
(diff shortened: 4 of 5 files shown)


Jaakko-Os:
looks ok to me

smoshiur1237 (commented):
Good to have separate action for repivoting. We can use them in upgrade task. But here is one issue with the following variable export.  In our upgrade case we do use different value for them. ```
NUM_OF_MASTER_REPLICAS := 1
export NUM_OF_WORKER_REPLICAS := 1
```

kashifest:
> Good to have separate action for repivoting. We can use them in upgrade task. But here is one issue with the following variable export. In our upgrade case we do use different value for them.
> `export NUM_OF_MASTER_REPLICAS := 1`
> `export NUM_OF_WORKER_REPLICAS := 1`

those variables would only matter in case of provisioning and deprovisioning  called through pivoting Makefile, do you foresee any such occasion in upgrade when this might happen?

smoshiur1237:
> > Good to have separate action for repivoting. We can use them in upgrade task. But here is one issue with the following variable export. In our upgrade case we do use different value for them.
> > `export NUM_OF_MASTER_REPLICAS := 1`
> > `export NUM_OF_WORKER_REPLICAS := 1`
> 
> those variables would only matter in case of provisioning and deprovisioning called through pivoting Makefile, do you foresee any such occasion in upgrade when this might happen?

When we call the pivoting script , it will call the verify.yml file where It counts the NUM_OF_BMH from those two variables.

kashifest:
verify.yml is called during `pivoting` action as shown here https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/vars/main.yml#L94, you will not be calling `pivoting`, you should be calling `upgrading`

smoshiur1237:
For your pivoting case the scripts will work properly. Its looks good.  But if we want to use the pivoting and repivoting for upgrade case, need to use only move.yml and move_back.yml file. I think we need to find a way to use them separately from the upgrade script and not to use the `make pivoting` or `make repivoting`. Then we can solve the issue and don't need to change those variable in that file.

smoshiur1237:
> verify.yml is called during `pivoting` action as shown here https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/vars/main.yml#L94, you will not be calling `pivoting`, you should be calling `upgrading`

Got it thanks

kashifest:
Well those variables that you were pointing out, in the first comment is not the problem, but this will be https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/tasks/move.yml#L100

kashifest:
tell me one things, the `NUMBER_OF_BMH` is coming from verify.yml for pivoting, how will you define it in your case since you will not be using verify.yml? @Xenwar @smoshiur1237

smoshiur1237:
> Well those variables that you were pointing out, in the first comment is not the problem, but this will be https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/tasks/move.yml#L100

If we add the new varables with correct number of CP and worker in a new  upgrading.sh file, do you think it will solve the issue for upgrading?

kashifest:
> > Well those variables that you were pointing out, in the first comment is not the problem, but this will be https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/tasks/move.yml#L100
> 
> If we add the new varables with correct number of CP and worker in a new upgrading.sh file, do you think it will solve the issue for upgrading?

Yes I am thinking that way

smoshiur1237:
If we use number of bmh, we don't need those variables.

kashifest:
wait @smoshiur1237 we have dependency between move.yml and verify.yml, move uses variables like `NUMBER_OF_BMH` or `IMAGE_USERNAME`, and also deploys CNI on the target cluster, during upgrade are you deploying CNI with your script?

kashifest:
@smoshiur1237 I have redefined `NUMBER_OF_BMH` so atleast move.yml should not be dependent on verify.yml, but you guys have to figure out a way to redefine `NUM_OF_MASTER_REPLICAS` and `NUM_OF_WORKER_REPLICAS` before you call `upgrading`.

/test-centos-integration
/test-integration
/test-features

smoshiur1237:
Looks good to me

smoshiur1237:
Waiing for this PR to be merged. After that will open new PR which will call the pivoting with new variables.

fmuyassarov (approved):

PR #475, opened 2020-09-23
Separate pivoting action for priority  upgrade use case.

smoshiur1237:
This PR will add separate script for calling pivoting with correct variables for priority upgrade use case. With current pivoting configuration, the upgrade script was unable to use the pivoting action during our upgrade tests because of dependency on these (`NUM_OF_MASTER_REPLICAS=1` and `NUM_OF_WORKER_REPLICAS=1`) two variables. In case of upgrading, the priority use case use different values. The repivoting has no dependency and can use it in upgrading use case by calling `make repivoting`.  So with the help of this PR we can call the pivoting from our upgrade script using `make upgrading`.

### DIFF
--- a/scripts/feature_tests/pivoting/Makefile
+++ b/scripts/feature_tests/pivoting/Makefile
@@ -15,3 +15,6 @@ repivoting:
 
 deprovision:
 	./../feature_test_deprovisioning.sh
+
+upgrading:
+	./upgrade.sh
--- a/scripts/feature_tests/pivoting/upgrade.sh
+++ b/scripts/feature_tests/pivoting/upgrade.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+export NUM_OF_MASTER_REPLICAS=3
+export NUM_OF_WORKER_REPLICAS=1
+
+METAL3_DIR="$(dirname "$(readlink -f "${0}")")/../../.."
+
+export ACTION="upgrading"
+
+"${METAL3_DIR}"/scripts/run.sh


kashifest (approved):
lgtm

scripts/feature_tests/pivoting/Makefile
```diff
@@ -15,3 +15,6 @@ repivoting:
 
 deprovision:
 	./../feature_test_deprovisioning.sh
+
+upgrading:
```
fmuyassarov, line 19: Who is calling this Makefile target?

smoshiur1237: This will be called by the scrip that Anwar is working on a priority test case and he will pivot with make upgrading with 3cp +1 worker.

scripts/feature_tests/pivoting/Makefile
```diff
@@ -15,3 +15,6 @@ repivoting:
...
+	./upgrade.sh
```
Jaakko-Os, line 20: The upgrade tests are run by script:
https://github.com/metal3-io/metal3-dev-env/blob/master/scripts/feature_tests/upgrade/upgrade.sh
Why not use that directly after the pivoting is finished?

smoshiur1237: In pivoting it also runs the repivoting after the pivoting is done. In our case we need to run the upgrade for 3cp+1w priority use case after pivoting. To call the pivoting with the master and worker node variable we need separate script to call the pivoting.

scripts/feature_tests/pivoting/Makefile
fmuyassarov: As an alternative, I just realized that wouldn't it be possible to rely on `pivoting` target instead of creating a new one? If case of upgrade test you would need to export NUM_OF_MASTER_REPLICAS & NUM_OF_WORKER_REPLICAS with desired numbers before you call the `make pivoting`. Because, at the end of the day both pivoting and upgrading are calling pivot.yml

Also, I think it would be better use imperative verb for the new target, i.e. `upgrade`, for the sake of the consistency.

kashifest: @fmuyassarov pivoting Makefile exports those variables too, as discussed yesterday, it would need an script after this to override the variables I believe

Jaakko-Os: It might be that I don't see the whole flow here but regarding the 3cp+1w priority test case, the NUM_OF_ variables are set to correct values as a first thing when the test case is started. Control plane and worker replicas are provisioned after that.

fmuyassarov: > @fmuyassarov pivoting Makefile exports those variables too, as discussed yesterday, it would need an script after this to override the variables I believe

Ops, sorry I missed that part. I'm ok with the current state. But another option would be to not export those vars in pivoting Makefile since pivoting scripts run with 2 nodes (default value in Ansible ..../vars.yaml). That would avoid the need to override the vars in the Makefile. WDYT @kashifest ?

smoshiur1237: If we are calling the` make pivoting` ,  it calls verify.yml and in our upgrade case it not needed. Also make file for pivoting its declare those variable with 1 for CP and 1 for Worker but we need 3 for cp and 1 for worker in the move.yml file. That is the reason we use separate script for upgrade task.

fmuyassarov (approved):

scripts/feature_tests/pivoting/Makefile
kashifest: @fmuyassarov that can be an option, we need to test it

fmuyassarov: > upgrade has different use-cases and different number of CP and worker anyhow,

Can't they export the vars with different values every time they run different upgrade tests?

kashifest: yeah true but still for pivoting then we need to export it separately if not default value if we dont do it here.

kashifest: well actually we can take the export in the pivot.sh

smoshiur1237: Also to be noted to speed up our test after pivoting, we prioritize 3cp+1w use case.

Jaakko-Os:
> /test-centos-integration

How can this ever work because NUM_OF_NODES=2 and 3cp+1w requires 4 nodes?

Jaakko-Os:
lgtm

PR #479, opened 2020-09-24
Extend the waiting time for BMH deprovisioning

maelk:
We are sometimes close to timeout (or over) in some cases
even if the cleaning actually completes properly right after.
This extends the wait period to avoid false positive.

/cc @fmuyassarov 
/cc @kashifest

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
@@ -38,7 +38,7 @@
         | .metadata.name ] | length'
     register: deprovisioned_nodes
     retries: 150
-    delay: 3
+    delay: 10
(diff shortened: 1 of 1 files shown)


maelk:
@kashifest and @fmuyassarov are you okay to skip the tests (considering the nature of the change + CentOS failing). That would allow me to go forward with the tests on other PRs that are now randomly failing due to this.

fmuyassarov:
> @kashifest and @fmuyassarov are you okay to skip the tests (considering the nature of the change + CentOS failing). That would allow me to go forward with the tests on other PRs that are now randomly failing due to this.

Yes, that's ok from my side. I don't this change would break anything.

fmuyassarov (commented):
/lgtm

kashifest:
@maelk good that you changed it. Completely ok with it. Didn't notice it earlier

PR #483, opened 2020-09-28
Make common functions cluster agnostic

Xenwar:
As we have them now, the common functions for upgrade workflow run against specific cluster. With the introduction of pivoting, the hard coded kubeconfig need to change.

This PR removes mentions of kubecofig files and parameters the secret, cluster and file names involved. 
It also introduce two functions for setting the kubeconfig towards the target cluster or the ephemeral cluster.

Individual test cases should export the KUBECONFIG variable to the correct cluster as needed.

### DIFF
--- a/scripts/feature_tests/upgrade/upgrade_common.sh
+++ b/scripts/feature_tests/upgrade/upgrade_common.sh
@@ -20,6 +20,10 @@ export IMAGE_RAW_CHECKSUM="http://172.22.0.1/images/${IMAGE_RAW_NAME}.md5sum"
 
 export CAPM3PATH=${CAPM3PATH:-"/home/${USER}/go/src/github.com/metal3-io/cluster-api-provider-metal3"}
 
+export CLUSTER_NAME=${CLUSTER_NAME:-"test1"}
+export TARGET_KUBECONFIG_SECRET=${TARGET_SECRET:-"${CLUSTER_NAME}-kubeconfig"}
+export TARGET_KUBECONFIG_FILE=${TARGET_KUBECONFIG_FILE:-"/tmp/kubeconfig-${CLUSTER_NAME}.yaml"}
+
 function generate_metal3MachineTemplate() {
     NAME="${1}"
     CLUSTER_UID="${2}"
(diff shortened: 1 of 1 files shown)


scripts/feature_tests/upgrade/upgrade_common.sh
```diff
@@ -471,5 +463,14 @@ function verify_kubernetes_version_upgrade() {
...
     done
+}
+
+function point_to_target_cluster() {
+    kubectl get secrets "${TARGET_KUBECONFIG_SECRET}" -n "${NAMESPACE}" -o json | jq -r '.data.value' | base64 -d >"${TARGET_KUBECONFIG_FILE}"
+    export KUBECONFIG="${TARGET_KUBECONFIG_FILE}"
+
+}
 
+function point_to_ephemeral_cluster() {
```
Jaakko-Os, line 474: My experience is that 'management' cluster is more commonly used

scripts/feature_tests/upgrade/upgrade_common.sh
```diff
@@ -471,5 +463,14 @@ function verify_kubernetes_version_upgrade() {
             break
         fi
     done
+}
+
+function point_to_target_cluster() {
```
Jaakko-Os, line 468: The test script calls this function, right?
The test script(s) change(s) should be included in this PR.

Jaakko-Os (changes requested):

scripts/feature_tests/upgrade/upgrade_common.sh
Xenwar: changed.

scripts/feature_tests/upgrade/upgrade_common.sh
Xenwar: Indeed, the functions in here are called from other test scripts. 
The way you suggested, we have https://github.com/metal3-io/metal3-dev-env/pull/455
But, it got bigger and trying to create smaller PRs such this.

maelk:
Do we expect any case where the original kubeconfig would not be in place or the target cluster would not have that name. i.e. is your goal to only run in CI or allow the tests to be run against any management and deployed cluster pair ? If only in CI, this lgtm

Xenwar:
This is to be in CI

Xenwar:
If you see any future use cases that can be handled before hand, we can make changes to this PR accordingly as well.

scripts/feature_tests/upgrade/upgrade_common.sh
Jaakko-Os: Small PRs help the review process significantly.
Then it's good to refer to #455 as there is a dependency.

maelk (approved):

PR #485, opened 2020-09-28
Create missing file when installing clusterctl

### LINKED ISSUE #470: Clusterctl is always re-installed
Even if clusterctl has already been installed, it is downloaded again due to two consecutive bugs : 

```
++ clusterctl version                                                                                                                                                                                                                                                                     
++ grep -o -P '(?<=GitVersion:).*?(?=,)'                                                                                                                                                                                                                                                  
Error: unable to verify clusterctl version: unable to write version state file: open /home/airshipci/.cluster-api/version.yaml: no such file or directory                                                                                                                                 
+ '[' '"v0.3.9"' '!=' v0.3.9 ']'                                                                                                                                                                                                                                                          
+ sudo rm /usr/local/bin/clusterctl
```
There is an issue with a missing file, and the comparison has an issue with quotes.

/help-needed
/good-first-issue

furkatgofurov7:
Fixes #470

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -100,6 +100,8 @@ function install_clusterctl() {
   curl -L https://github.com/kubernetes-sigs/cluster-api/releases/download/"${CAPIRELEASE}"/clusterctl-linux-amd64 -o clusterctl
   chmod +x ./clusterctl
   sudo mv ./clusterctl /usr/local/bin/clusterctl
+  mkdir -p home/"${USER}"/.cluster-api
+  touch home/"${USER}"/.cluster-api/version.yaml
 }
 
 if ! [ -x "$(command -v clusterctl)" ]; then


01_prepare_host.sh
```diff
@@ -100,6 +100,8 @@ function install_clusterctl() {
   curl -L https://github.com/kubernetes-sigs/cluster-api/releases/download/"${CAPIRELEASE}"/clusterctl-linux-amd64 -o clusterctl
   chmod +x ./clusterctl
   sudo mv ./clusterctl /usr/local/bin/clusterctl
+  mkdir -p home/"${USER}"/.cluster-api
+  touch home/"${USER}"/.cluster-api/version.yaml
```
maelk, line 104: this file itself might not be needed, as long as the .cluster-api exists.

maelk (approved):

01_prepare_host.sh
furkatgofurov7: should we ignore it maybe, if it exists, i.e `mkdir -p home/"${USER}"/.cluster-api || true `, WDYT?

maelk: mkdir -p will not error if the folder exist. I just meant that the file "version.yaml" probably does not need to be created.

maelk (approved):

PR #488, opened 2020-09-29
Parameterize registry endpoint

fmuyassarov:
Parameterize registry endpoint (IP & port). Also, upgrade Go version.


furkatgofurov7 (commented):
Can't we use `EXTERNAL_SUBNET_V4_HOST` instead of introducing `REGISTRY_IP`?

furkatgofurov7:
Actually, var `REGISTRY` is the same thing IMO, correct me if I am wrong @fmuyassarov 
(REGISTRY=192.168.111.1:5000)

fmuyassarov:
> Actually, var `REGISTRY` is the same thing IMO, correct me if I am wrong @fmuyassarov
> (REGISTRY=192.168.111.1:5000)

Yes, I just realized that we already have a REGISTRY var for that. But this wasn't anyways parameterized in Ansible scripts. Will update the patch. Thanks

fmuyassarov:
> Can't we use `EXTERNAL_SUBNET_V4_HOST` instead of introducing `REGISTRY_IP`?

Done.

kashifest:
lgtm, but I am thinking the other way around, should we tie up the two entities just not to use one more variable?

fmuyassarov:
> lgtm, but I am thinking the other way around, should we tie up the two entities just not to use one more variable?

Do you mean a single var to contain both IP & Port?

fmuyassarov:
> > lgtm, but I am thinking the other way around, should we tie up the two entities just not to use one more variable?
> 
> Do you mean a single var to contain both IP & Port?
If yes,  I think both are fine.

lib/network.sh
```diff
@@ -149,15 +149,17 @@ else
   export BAREMETALV6_POOL_RANGE_END=""
 fi
 
+export REGISTRY_PORT=${REGISTRY_PORT:-"5000"}
```
maelk, line 152: I understand why you would want to have it parameterize, however, if you go that way, it should either not be configurable by the user, or you should also parameterize the docker command that starts the registry https://github.com/metal3-io/metal3-dev-env/blob/master/02_configure_host.sh#L133

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
```diff
@@ -183,7 +184,7 @@ export CTLPLANE_KUBEADM_EXTRA_CONFIG="
         permissions: '0644'
         content: |
           {
-            \"insecure-registries\" : [\"192.168.111.1:5000\"]
+            \"insecure-registries\" : ["{{ EXTERNAL_SUBNET_V4_HOST }}":"{{ REGISTRY_PORT }}"]
```
maelk, line 187: Maybe EXTERNAL_SUBNECT_V4_HOST is not the best variable to use there. you should probably use a variable that is IP version agnostic.

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
```diff
@@ -283,6 +284,6 @@ export WORKERS_KUBEADM_EXTRA_CONFIG="
           permissions: '0644'
           content: |
             {
-              \"insecure-registries\" : [\"192.168.111.1:5000\"]
+              \"insecure-registries\" : ["{{ EXTERNAL_SUBNET_V4_HOST }}":"{{ REGISTRY_PORT }}"]
```
maelk, line 287: same here

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_ubuntu.rc
```diff
@@ -136,7 +137,7 @@ export CTLPLANE_KUBEADM_EXTRA_CONFIG="
         permissions: '0644'
         content: |
           {
-            \"insecure-registries\" : [\"192.168.111.1:5000\"]
+            \"insecure-registries\" : ["{{ EXTERNAL_SUBNET_V4_HOST }}":"{{ REGISTRY_PORT }}"]
```
maelk, line 140: and here

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_ubuntu.rc
```diff
@@ -188,6 +189,6 @@ export WORKERS_KUBEADM_EXTRA_CONFIG="
           permissions: '0644'
           content: |
             {
-              \"insecure-registries\" : [\"192.168.111.1:5000\"]
+              \"insecure-registries\" : ["{{ EXTERNAL_SUBNET_V4_HOST }}":"{{ REGISTRY_PORT }}"]
```
maelk, line 192: and here

lib/network.sh
fmuyassarov: Fixed. Thanks

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
fmuyassarov: Fixed thanks

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_ubuntu.rc
fmuyassarov: Fixed, thanks

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_ubuntu.rc
fmuyassarov: Fixed

fmuyassarov:
@maelk  lgty?

02_configure_host.sh
```diff
@@ -130,7 +130,7 @@ if [[ "$reg_state" == "exited" ]]; then
   sudo "${CONTAINER_RUNTIME}" start registry
 elif [[ "$reg_state" != "running" ]]; then
   sudo "${CONTAINER_RUNTIME}" rm registry -f || true
-  sudo "${CONTAINER_RUNTIME}" run -d -p 5000:5000 --name registry "$DOCKER_REGISTRY_IMAGE"
+  sudo "${CONTAINER_RUNTIME}" run -d -p "${REGISTRY_PORT}:${REGISTRY_PORT}" --name registry "$DOCKER_REGISTRY_IMAGE"
```
maelk, line 133: One of them cannot be parameterized as it is the port where the executable inthe container listens. You can only configure the exposed one

maelk (changes requested):

02_configure_host.sh
fmuyassarov: Fixed, thanks

kashifest:
> > lgtm, but I am thinking the other way around, should we tie up the two entities just not to use one more variable?
> 
> Do you mean a single var to contain both IP & Port?
> If yes, I think both are fine.

My concerns are already addressed.

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
```diff
@@ -283,6 +284,6 @@ export WORKERS_KUBEADM_EXTRA_CONFIG="
           permissions: '0644'
           content: |
             {
-              \"insecure-registries\" : [\"192.168.111.1:5000\"]
+              \"insecure-registries\" : ["{{ REGISTRY }}"]
```
furkatgofurov7, line 287: Not sure, but maybe this might be the reason why CI is failing, WDYT?:
```suggestion
              \"insecure-registries\" : [\"{{ REGISTRY\ }}"]
```

fmuyassarov: Hum, it passed my local tests though.

fmuyassarov: @furkatgofurov7 Seems you were right. Added backslash now, let's see how it goes in the CI this time.

furkatgofurov7 (commented):
Looks good to me!
Leave it to others to review it once again.

PR #489, opened 2020-09-30
Add action for verify script before pivoting in upgrade task.

smoshiur1237:
We are facing an issue with the calico during the pivoting in upgrade task. Because the calico.yaml file that we apply in our upgrade script needs additional configuration for metal3-dev-env. If we run the verify.yml file before pivoting, it applies the correct configuration for CNI. This PR will solve this issue with the CNI.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/vars/main.yml
+++ b/vm-setup/roles/v1aX_integration_test/vars/main.yml
@@ -92,6 +92,7 @@ verify_actions:
     - "ci_test_provision"
     - "feature_test_provisioning"
     - "pivoting"
+    - "upgrading"  
 pivot_actions:
     - "pivoting"    
     - "upgrading"


Jaakko-Os (commented):
it would be important to run '/test-upgrade-features' for this PR

smoshiur1237:
> it would be important to run '/test-upgrade-features' for this PR

This change is not affecting any upgrade-features test yet. We are testing locally the upgrade script after pivoting. When we open that PR, this change will be used. Do you still think its necessary to run '/test-upgrade-features' for this PR.

Jaakko-Os:
> > it would be important to run '/test-upgrade-features' for this PR
> 
> This change is not affecting any upgrade-features test yet. We are testing locally the upgrade script after pivoting. When we open that PR, this change will be used. Do you still think its necessary to run '/test-upgrade-features' for this PR.

Ok. Then skip '/test-upgrade-features' part.

PR #498, opened 2020-10-02
Fetch the latest Cluster API release

furkatgofurov7:
Currently we are fetching Cluster API release `v0.3.9` which is not latest, this PR should fix the problem of fetching the latest release.

### DIFF
--- a/lib/releases.sh
+++ b/lib/releases.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# shellcheck disable=SC1091
+source lib/common.sh
+
 function get_latest_release() {
   set +x
   if [ -z "${GITHUB_TOKEN:-}" ]; then
@@ -8,7 +11,15 @@ function get_latest_release() {
     release="$(curl -H "Authorization: token ${GITHUB_TOKEN}" -sL "${1}")" || ( set -x && exit 1 )
   fi
   # This gets the latest release as vx.y.z , ignoring any version with a suffix starting with - , for example -rc0
-  release_tag="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\")) | select(contains(\"-\")==false)] | max ")"
+  release_tags_unsorted="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\")) | select(contains(\"-\")==false)]" \
+    | cut -sf2 -d\"  | tr '.' ' ' )"
+  if [[ $OS == ubuntu ]]; then
+    release_tags_sorted="$(echo "$release_tags_unsorted" | sort -n +1 +2 )"
+  else
+    release_tags_sorted="$(echo "$release_tags_unsorted" | sort -nk1 -nk2 -nk3 )"
+  fi
+  release_tag="$(echo "$release_tags_sorted" | tail -n 1 | tr ' ' '.')"
+
   if [[ "$release_tag" == "null" ]]; then
     set -x
     exit 1


namnx228 (approved):

lib/releases.sh
```diff
@@ -8,7 +8,14 @@ function get_latest_release() {
     release="$(curl -H "Authorization: token ${GITHUB_TOKEN}" -sL "${1}")" || ( set -x && exit 1 )
   fi
   # This gets the latest release as vx.y.z , ignoring any version with a suffix starting with - , for example -rc0
-  release_tag="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\")) | select(contains(\"-\")==false)] | max ")"
+  if [[ "${IMAGE_OS}" == "Ubuntu" ]]; then
```
maelk, line 11: I don't think this is the correct variable, you want to know if you are on an Ubuntu OS, not if you are deploying an Ubuntu target cluster.

maelk (changes requested):

lib/releases.sh
maelk: the fact that we deploy Ubuntu on Ubuntu and Centos target cluster on CentOS should not be assumed. it could be changed

furkatgofurov7: Yes, I want to know OS flavor, because `sort -n +1` did not work for Centos, it was Ubuntu OS specific.

furkatgofurov7: What do you suggest, any pointers ?

furkatgofurov7: Centos has failed the test, with sort, so I was thinking it is OS dependant and I could use a condition. What kind of drawback do you foresee with these changes? i.e different OS(fedora, cinder) willn't work with above changes, is that what you are concerned about?

furkatgofurov7: I got your point now. What variable I could have use in that case, so even if we are not deploying Ubuntu on Ubuntu, it will be getting the correct sorting. Any ideas welcome

maelk: check lib/common.sh . Everything you need is there. you could also have a look at 01_ that executes a script if it is Ubuntu and something else if CentOS

furkatgofurov7: Done

lib/releases.sh
```diff
@@ -8,7 +11,14 @@ function get_latest_release() {
...
   fi
   # This gets the latest release as vx.y.z , ignoring any version with a suffix starting with - , for example -rc0
-  release_tag="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\")) | select(contains(\"-\")==false)] | max ")"
+  if [[ $OS == ubuntu ]]; then
+    release_tag="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\")) | select(contains(\"-\")==false)]" \
+      | cut -sf2 -d\"  | tr '.' ' ' | sort -n +1 +2 | tail -n 1 | tr ' ' '.')"
+  else
+    release_tag="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\")) | select(contains(\"-\")==false)]" \
+      | cut -sf2 -d\"  | tr '.' ' ' | sort -nk1 -nk2 -nk3 | tail -n 1 | tr ' ' '.')"
+  fi
```
maelk, line 20: would the CentOS one work on Ubuntu ? If it is different version of sort there might be some compatibility in one way. You could also do something like this to reduce code repetition: 
```suggestion
  release_tags_unsorted="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\")) | select(contains(\"-\")==false)]" \
      | cut -sf2 -d\"  | tr '.' ' ' )"
  if [[ $OS == ubuntu ]]; then
    release_tags_sorted="$(echo "$release_tags_unsorted"  | sort -n +1 +2 )"
  else
    release_tags_sorted="$(echo "$release_tags_unsorted" | sort -nk1 -nk2 -nk3 )"
  fi
  release_tag="$(echo "$release_tags_sorted" | tail -n 1 | tr ' ' '.')"
```

maelk (approved):

lib/releases.sh
furkatgofurov7: I liked the proposed way of sorting, it is cleaner compared to what I have. Updated it. Thanks @maelk !

fmuyassarov (commented):
/lgtm

PR #499, opened 2020-10-05
CI: Remove sourcing of common.sh script

furkatgofurov7:
Master feature tests are [not passing](https://jenkins.nordix.org/view/Airship/job/airship_master_feature_tests_ubuntu/173/consoleFull) due to the unnecessary sourcing of `lib/common.sh` in `lib/releases.sh` script with logs:

`+ source /home/****/tested_repo/scripts/../lib/releases.sh
++ source lib/common.sh
/home/****/tested_repo/scripts/../lib/releases.sh: line 4: lib/common.sh: No such file or directory
make[1]: Leaving directory '/home/****/tested_repo/scripts/feature_tests/remediation'
make[1]: *** [Makefile:7: provision] Error 1
make: *** [Makefile:42: remediation_test] Error 2`

This PR removes sourcing of `lib/common.sh` script introduced in #498 . The reason for removal of `lib/common.sh` sourcing is because it is already sourced before running `lib/releases.sh` in all its execution places.

### DIFF
--- a/lib/releases.sh
+++ b/lib/releases.sh
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-# shellcheck disable=SC1091
-source lib/common.sh
-
 function get_latest_release() {
   set +x
   if [ -z "${GITHUB_TOKEN:-}" ]; then


fmuyassarov:
There are some vars in lib/common.sh that this script relies on. How are you planning to export them?

fmuyassarov:
> There are some vars in lib/common.sh that this script relies on. How are you planning to export them?

Will you rely on the default values within the script?

lib/releases.sh
```diff
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-# shellcheck disable=SC1091
-source lib/common.sh
```
maelk, line 4: no, this is required! some variables are defined below, and should come from lib/common.sh, like CAPM3_VERSION.

maelk (changes requested):
The problem with your error is not the sourcing per say, but the path depending on its execution place. you need to make it relative to the releases.sh script.

lib/releases.sh
furkatgofurov7: Maybe a dump question, but how CAPM3_VERSION variable was sourced from lib/common.sh even when it was not there before #498 merged?

fmuyassarov: I think the problem is here : https://github.com/metal3-io/metal3-dev-env/blob/8857751d750e33f210b114f05f929ca24775ad64/scripts/run.sh#L17
So, it is sourcing releases.sh from "${METAL3_DIR}/lib/" and but when it comes to sourcing common.sh from within the releases.sh it tries to find it from lib/ not "${METAL3_DIR}/lib/"

fmuyassarov: And @maelk actually removing it should be fine, because wherever we call releases.sh we always first source the common.sh. Of course, this is not the best option, because we always have to remember to execute common.sh before we execute releases.sh, but this answers your question above @furkatgofurov7

furkatgofurov7: @maelk wdyt?

maelk: yes, you can remove it

furkatgofurov7:
> /test-features

tests features is failing due to the another issue and this PR has nothing to do with the error.

fmuyassarov (commented):
/lgtm

furkatgofurov7:
CI instability, re-triggering.
/test-centos-integration

maelk (approved):

PR #453, opened 2020-09-11
CI: add combined upgrade test case

Jaakko-Os:
The combined upgrade consists of
-controlplane component upgrade using clusterctl
-k8s version and disk image upgrade for
 master and worker
-ironic image upgrade

Fixes: #427
Fixes: #418, #419

PR #500, opened 2020-10-06
[WIP] Re-introduces changes reverted by #453

Xenwar:
Re-introduces changes reverted by #453
Content was reviewed and approved. Therefore, does not need a review again.

This PR by itself will break the ci upgrade tests. Therefore, needs to be merged with other would be created PRs.

### DIFF
--- a/scripts/feature_tests/upgrade/upgrade_common.sh
+++ b/scripts/feature_tests/upgrade/upgrade_common.sh
@@ -20,6 +20,10 @@ export IMAGE_RAW_CHECKSUM="http://172.22.0.1/images/${IMAGE_RAW_NAME}.md5sum"
 
 export CAPM3PATH=${CAPM3PATH:-"/home/${USER}/go/src/github.com/metal3-io/cluster-api-provider-metal3"}
 
+export CLUSTER_NAME=${CLUSTER_NAME:-"test1"}
+export TARGET_KUBECONFIG_SECRET=${TARGET_SECRET:-"${CLUSTER_NAME}-kubeconfig"}
+export TARGET_KUBECONFIG_FILE=${TARGET_KUBECONFIG_FILE:-"/tmp/kubeconfig-${CLUSTER_NAME}.yaml"}
+
 function generate_metal3MachineTemplate() {
     NAME="${1}"
     CLUSTER_UID="${2}"
@@ -101,8 +105,8 @@ function wait_for_cluster_deprovisioned() {
 }
 
 function deploy_workload_on_workers() {
-    echo "Deploy workloads on workers"
-    # Deploy workloads
+echo "Deploy workloads on workers"
+# Deploy workloads
 cat <<EOF | kubectl apply -f -
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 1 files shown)


maelk:
@Xenwar So does this need a hold ? I am unclear if it is ready for merge, or if not, what should be waited on.

Xenwar:
> @Xenwar So does this need a hold ? I am unclear if it is ready for merge, or if not, what should be waited on.

Ready for merge, but it needs a hold as well as it will break the ci test runs.
There are other PRs to follow. And, the coming ones and this one need to get merged together.

scripts/feature_tests/upgrade/upgrade_common.sh
```diff
@@ -364,10 +356,10 @@ function cleanup_clusterctl_configuration() {
     mkdir /tmp/cluster-api-clone
 
     # clean up if previous test has failed.
-    rm -rf /home/"${USER}"/.cluster-api/dev-repository/cluster-api/v0.3.8
-    rm -rf /home/"${USER}"/.cluster-api/dev-repository/bootstrap-kubeadm/v0.3.8
-    rm -rf /home/"${USER}"/.cluster-api/dev-repository/control-plane-kubeadm/v0.3.8
-    rm -rf /home/"${USER}"/.cluster-api/overrides/infrastructure-metal3/v0.3.8
```
Jaakko-Os, line 370: keep these

scripts/feature_tests/upgrade/upgrade_common.sh
```diff
@@ -376,31 +368,31 @@ function create_clusterctl_configuration() {
 cat <<EOF >/home/"${USER}"/.cluster-api/clusterctl.yaml
 providers:
   - name: cluster-api
-    url: /home/$USER/.cluster-api/dev-repository/cluster-api/v0.3.5/core-components.yaml
```
Jaakko-Os, line 379: General comment: don't change the clusterctl versions and paths. The master branch content is correct.

scripts/feature_tests/upgrade/upgrade_common.sh
```diff
@@ -144,12 +148,10 @@ EOF
 }
 
 function manage_node_taints() {
-    kubectl get secrets "${CLUSTER_NAME}"-kubeconfig -n "${NAMESPACE}" -o json | \
-      jq -r '.data.value'| base64 -d > /tmp/kubeconfig-"${CLUSTER_NAME}".yaml
```
Jaakko-Os, line 148: General comment: I would do the kubeconfig handling in upgrade_common.sh where it's needed. Will this work after pivoting? I don't have an answer yet.

Like this:

function call
: 
point_to_management_cluster
get_target_kubeconfig
point_to_target_cluster

 // run commands on target cluster
: 
// before exiting
point_to_management_cluster

Jaakko-Os: This way the actual test scripts would not need to worry about kubeconfig.

Jaakko-Os (changes requested):

Jaakko-Os:
verification needs to pass with test-upgrade-features and test-upgrade-features-centos before merge

Xenwar:
> verification needs to pass with **test-upgrade-features** and **test-upgrade-features-centos** before merge
**test-upgrade-features** will not pass. 
given the comments, I think we can close this PR and take yours https://github.com/metal3-io/metal3-dev-env/pull/504

scripts/feature_tests/upgrade/upgrade_common.sh
Xenwar: Sorry, I missed this comment. 
If we take those rows to the common, then how does the test run ?
i.e The logic is in the non-common functions and order varies and cannot be pre-determined in the common functions file

PR #501, opened 2020-10-06
Update k8sversion to 1.18.8

kashifest:
This PR Updates k8s version to 1.18.8 to avoid mismatch in kubelet and kubectl versions since in node image we already use k8s version 1.18.8


Xenwar:
A description in the commit would make this PR easy to understand in the future.

kashifest:
> A description in the commit would make this PR easy to understand in the future.

Commit description is there, you mean PR I guess, updating it

Jaakko-Os (changes requested):
This change will impact the k8s version upgrade tests.
It should be verified within this PR that at least one k8s version upgrade test is successful.

kashifest:
> This change will impact the k8s version upgrade tests.
> It should be verified within this PR that at least one k8s version upgrade test is successful.

That is something that the upgrade tests should take care of since the environment is changing.  I am not going to change upgrade tests in this PR.

fmuyassarov (commented):
/lgtm

furkatgofurov7 (commented):
Looks good to me!

maelk (approved):

PR #502, opened 2020-10-06
[WIP] Make 1cp_1w_bootDiskImage and 1cp_3w_bootDiskImage run after pivoting

Xenwar:
This PR enables the 1cp_1w_bootDiskImage upgrade test to run after pivoting.

It depends on #500 and should be on hold until further notice.

### DIFF
--- a/scripts/feature_tests/upgrade/1cp_1w_bootDiskImage_cluster_upgrade.sh
+++ b/scripts/feature_tests/upgrade/1cp_1w_bootDiskImage_cluster_upgrade.sh
@@ -11,19 +11,30 @@ echo '' >~/.ssh/known_hosts
 
 start_logging "${1}"
 # Provision original nodes
-set_number_of_master_node_replicas 1
+set_number_of_master_node_replicas 3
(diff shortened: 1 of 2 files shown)


scripts/feature_tests/upgrade/1cp_1w_bootDiskImage_cluster_upgrade.sh
```diff
@@ -11,20 +11,31 @@ echo '' >~/.ssh/known_hosts
...
+get_target_kubeconfig
+point_to_target_cluster
 controlplane_is_provisioned
-controlplane_has_correct_replicas 1
-
-# apply CNI
-apply_cni
+controlplane_has_correct_replicas 3
 
+point_to_management_cluster
```
Jaakko-Os, line 25: like I commented in https://github.com/metal3-io/metal3-dev-env/pull/500, I would not do the kubeconfig magic in every test case but centralized in upgrade_common.sh

Jaakko-Os (changes requested):

PR #519, opened 2020-10-15
🐛 Fix removing all ironic containers properly

furkatgofurov7:
This patch tries to do a cleaning up of the the ironic containers properly in 01_*

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -112,10 +112,7 @@ elif [ "$(clusterctl version | grep -o -P '(?<=GitVersion:").*?(?=",)')" != "${C
 fi
 
 # Clean-up any old ironic containers
-for name in ironic ironic-inspector dnsmasq httpd mariadb ipa-downloader; do
-    sudo "${CONTAINER_RUNTIME}" ps | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" kill $name
-    sudo "${CONTAINER_RUNTIME}" ps --all | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" rm $name -f
-done
+remove_ironic_containers
 
 # Clean-up existing pod, if podman
 case $CONTAINER_RUNTIME in
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -426,3 +426,14 @@ function init_minikube() {
           --type network --config
     fi
 }
+
+#
+# Kill and remove the running ironic containers
+#
+function remove_ironic_containers() {
+  #shellcheck disable=SC2015
+  for name in ipa-downloader ironic ironic-api ironic-conductor ironic-inspector ironic-endpoint-keepalived dnsmasq httpd mariadb vbmc sushy-tools httpd-infra; do
+    sudo "${CONTAINER_RUNTIME}" ps | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" kill $name || true
+    sudo "${CONTAINER_RUNTIME}" ps --all | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" rm $name -f || true
+  done
+}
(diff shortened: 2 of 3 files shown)


maelk (approved):

01_prepare_host.sh
```diff
@@ -112,10 +112,7 @@ elif [ "$(clusterctl version | grep -o -P '(?<=GitVersion:").*?(?=",)')" != "${C
 fi
 
 # Clean-up any old ironic containers
-for name in ironic ironic-inspector dnsmasq httpd mariadb ipa-downloader; do
-    sudo "${CONTAINER_RUNTIME}" ps | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" kill $name
-    sudo "${CONTAINER_RUNTIME}" ps --all | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" rm $name -f
-done
+remove_ironic_containers
```
fmuyassarov, line 115: If you look at the [install_requirements.sh ](https://github.com/metal3-io/metal3-dev-env/blob/746bf689dad5a83a9f9bbbdf6e18af677c004eab/ubuntu_install_requirements.sh#L7) you can see that it is sourcing lib/common.sh before it installs the docker/podman. I think that's the reason why the CI failed.

fmuyassarov: Sorry ignore my comment, I didn't realize that you aren't actually calling the remove_ironic_containers func within the lib/common.sh

elfosardo (approved):

smoshiur1237 (commented):
/lgtm

lib/common.sh
```diff
@@ -426,3 +426,15 @@ function init_minikube() {
...
+
+#
+# Kill and remove the running ironic containers
+#
+function remove_ironic_containers() {
+  for name in ipa-downloader ironic ironic-api ironic-conductor ironic-inspector ironic-endpoint-keepalived dnsmasq httpd mariadb vbmc sushy-tools httpd-infra; do
+    sudo "${CONTAINER_RUNTIME}" ps | grep -w "$name$" || true
+    sudo "${CONTAINER_RUNTIME}" kill $name || true
+    sudo "${CONTAINER_RUNTIME}" ps --all | grep -w "$name$" || true
+    sudo "${CONTAINER_RUNTIME}" rm $name -f || true
```
maelk, line 438:
```suggestion
    sudo "${CONTAINER_RUNTIME}" ps | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" kill $name || true
    sudo "${CONTAINER_RUNTIME}" ps --all | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" rm $name -f || true
```

maelk (changes requested):

lib/common.sh
furkatgofurov7: I have tried this, shell will fail will following:
`In ./lib/common.sh line 435:
    sudo "${CONTAINER_RUNTIME}" ps | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" kill $name || true
                                                      ^-- SC2015: Note that A && B || C is not if-then-else. C may run when A is true.
In ./lib/common.sh line 436:
    sudo "${CONTAINER_RUNTIME}" ps --all | grep -w "$name$" && sudo "${CONTAINER_RUNTIME}" rm $name -f || true
                                                            ^-- SC2015: Note that A && B || C is not if-then-else. C may run when A is true.`

furkatgofurov7: I wanted to escape && and || to come together. Putting them separately works, but it will not fail silently:
`+ sudo podman ps
+ grep -w 'ipa-downloader$'
+ true
+ sudo podman kill ipa-downloader
Error: no container with name or ID ipa-downloader found: no such container
+ true`

maelk: That is a false positive from the shellcheck. We really want C to run even if A succeeds. We want C to run whenever B fails. So disable shellcheck there

fmuyassarov (approved):

PR #322, opened 2020-05-18
Remove v1alpha1 scripts

fmuyassarov:
Remove unused:

- `/scripts/v1alpha1/`
- `provision_host.sh`
- `deprovision_host.sh`

PR #526, opened 2020-10-21
🐛 Fix referencing to deleted "provision_host.sh" script

furkatgofurov7:
The `provision_host.sh` script was part of v1alpha1 scripts and was removed by [this patch ](https://github.com/metal3-io/metal3-dev-env/pull/322)


vars.md
```diff
@@ -11,19 +11,19 @@ assured that they are persisted.
...
 | EXTERNAL_SUBNET                | This is the subnet used on the "baremetal" libvirt network, created as the primary network interface for the virtual bare metalhosts.                                                                                                                    | CIDR                               | 192.168.111.0/24                                             |
-| SSH_PUB_KEY                    | This SSH key will be automatically injected into the provisioned host by the provision_host.sh script.                                                                                                                                                   |                           | ~/.ssh/id_rsa.pub                                            |
+| SSH_PUB_KEY                    | This SSH key will be automatically injected into the provisioned host by the lib/common.sh script.                                                                                                                                                   |                           | ~/.ssh/id_rsa.pub                                            |
```
namnx228, line 14: Could you please refer to code location where this happens?

namnx228 (commented):
Look good to me. Only one thing left is the thing I commented

vars.md
furkatgofurov7: Sure, [here](https://github.com/metal3-io/metal3-dev-env/blob/master/lib/common.sh#L60) it is

namnx228: I think we declare the key in `lib/common.sh`, but this key will be automatically injected into the provisioned host by another script, no?

furkatgofurov7: Which script do you mean by another script?

furkatgofurov7: Current `01_prepare_host.sh` is what `provision_host.sh` had been in my opinion, is that what you meant?

namnx228: No, I mean the script which does something like `ssh-add <key><host>`. If I am not mistaken, script 02 or 03 do that.

furkatgofurov7: You were right, I have changed it, thanks for review!

namnx228 (approved):

maelk (approved):

PR #536, opened 2020-10-29
WIP CI: backup and restore libvirt VM

Jaakko-Os:
To speed up feature testing, VMs are backed up when
an appropriate cluster setup created.
After a test is run the base setup can be restored.

Cluster deprovisioning, which is time consuming,
is no longer needed after every test case
because the same state can be achieved with the restore.

### DIFF
--- a/scripts/feature_tests/backup-restore-libvirt-vm.sh
+++ b/scripts/feature_tests/backup-restore-libvirt-vm.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+set -x 
+
+PROCEDURE="${1}"
+
+if [ "${PROCEDURE}" == "backup" ]; then
+  BMH_LIST=$(kubectl get bmh -A | awk 'NR>1' | awk '{{print $2}}' | grep node)
+
+  for bmh_node in ${BMH_LIST}; do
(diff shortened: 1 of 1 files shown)


Jaakko-Os:
A way to create an external disk and memory snapshot of the virsh VMs (node_0 as example) https://blog.programster.org/kvm-external-snapshots:

```
sudo virsh snapshot-create-as \
 --domain node_0 node_0_snap \
 --diskspec sda,file=/home/ubuntu/backups/node_0_disk_snap.qcow2,snapshot=external \
 --memspec file=/home/ubuntu/backups/node_0_mem_snap.qcow2,snapshot=external \
 --atomic
```

Only downside is that the reverting is currently only supported for internal snapshots:

```
$ sudo virsh snapshot-revert node_0 node_0_snap --running
 error: unsupported configuration: revert to external snapshot not supported yet
```

Would there be ways to go around this?
There is virt-backup tool https://github.com/aruhier/virt-backup, which I didn't manage to run successfully (python issues) and
then I found some qemu-img style hacks https://unix.stackexchange.com/questions/351004/kvm-live-external-snapshots-and-appropriate-disk-names, this site states that the memory revert part is not figured out and unclear if possible.

PR #548, opened 2020-11-03
CI: disable tls and basic auth in upgrade tests

Jaakko-Os:


Jaakko-Os:
After discussion it was decided that the change shall be implemented on the project-infra side.

PR #551, opened 2020-11-05
Add fetch manifest script to the feature tests.

smoshiur1237:
This PR will add the fetch manifest script to the feature tests.

### DIFF
--- a/scripts/feature_tests/pivoting/Makefile
+++ b/scripts/feature_tests/pivoting/Makefile
@@ -2,10 +2,13 @@ M3PATH := "$(dirname "$(readlink -f "${0}")")../../../"
 export NUM_OF_MASTER_REPLICAS := 3
 export NUM_OF_WORKER_REPLICAS := 1
 
-all: provision pivoting repivoting deprovision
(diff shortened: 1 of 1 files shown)


kashifest:
lgtm, run the feature tests also here to check if its working properly

Makefile
```diff
@@ -45,9 +45,11 @@ upgrade_test:
 	make -C ./scripts/feature_tests/upgrade/
 
 feature_tests: setup_env remediation_test cleanup_env pivoting_test
+	./scripts/fetch_manifests.sh
```
Jaakko-Os, line 48: would it be more consistent if you define a target,
fetch_manifests:
        ./scripts/fetch_manifests.sh

and then use the 'fetch_manifests' target
feature_tests: fetch_manifests setup_env remediation_test cleanup_env pivoting_test

smoshiur1237: Yes that can be done also.

smoshiur1237: Done

fmuyassarov:
There are no manifests apart from BMH unfortunately.

Jaakko-Os:
looks good!

PR #554, opened 2020-11-10
Fix ironic configmap for TLS

kashifest:
This PR fixes a bug for ironic bmo configmap specially for TLS setup. Since in 03 we modify the ironic configmap to reflect TLS changes , we also need the same setup post pivot. However, we were overriding those values once we were done with the source cluster setup. This PR keeps a copy of the configuration and uses that for target cluster.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -177,21 +177,9 @@ function update_images(){
 function launch_ironic() {
   pushd "${BMOPATH}"
 
-  if [ "${EPHEMERAL_CLUSTER}" != "minikube" ]; then
-    update_images
-    ${RUN_LOCAL_IRONIC_SCRIPT}
-  else
-    # Deploy Ironic using deploy.sh script
-
-    # Update container images to use local ones
-    cp "${BMOPATH}/ironic-deployment/ironic/ironic.yaml" "${BMOPATH}/ironic-deployment/ironic/ironic.yaml.orig"
-    cp "${BMOPATH}/ironic-deployment/keepalived/keepalived_patch.yaml" "${BMOPATH}/ironic-deployment/keepalived/keepalived_patch.yaml.orig"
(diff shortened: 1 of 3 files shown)


kashifest:
This also needs change, it will work for centos only

maelk:
I don't think this is the best way to do things. The issue with not restoring the files to their proper initial values, is that when you will try to changes branches, it will create conflicts. A better approach would be to generate properly the config wherever needed

kashifest:
> I don't think this is the best way to do things. The issue with not restoring the files to their proper initial values, is that when you will try to changes branches, it will create conflicts. A better approach would be to generate properly the config wherever needed

The thing is, this whole setup should be in one go. IMO the whole setup should be done once and then everything should be done as it is on the target cluster. Doing it in one place ensures you have the same setup. Doing it in all the places would make maintainability difficult.

kashifest:
However, I agree this would make the switching of the branch difficult. I dont foresee this to be an issue as compared to making it homogeneous for pivot, but then again its a problem. I ll try to do this as part of pivot script.

maelk:
> The thing is, this whole setup should be in one go. IMO the whole setup should be done once and then everything should be done as it is on the target cluster. Doing it in one place ensures you have the same setup. Doing it in all the places would make maintainability difficult.

I agree, but you could consider something like having a single function defined in a common place and called twice to ensure that everything is done in a single place

kashifest:
> > The thing is, this whole setup should be in one go. IMO the whole setup should be done once and then everything should be done as it is on the target cluster. Doing it in one place ensures you have the same setup. Doing it in all the places would make maintainability difficult.
> 
> I agree, but you could consider something like having a single function defined in a common place and called twice to ensure that everything is done in a single place

how does the current version look ?

maelk (approved):

fmuyassarov (commented):
mv didn't work?
/lgtm

PR #555, opened 2020-11-13
Fix serial log deletion

kashifest:
Serial logs were not getting deleted. This creates havoc in m3-dev-env iterative testing. This PR fixes the deletion.


furkatgofurov7 (commented):
/lgtm

PR #561, opened 2020-11-22
Migrate ubuntu_install_req*.sh from bash to Ansible

fmuyassarov:
Move ubuntu_install_requirements.sh to an Ansible script.


01_prepare_host.sh
```diff
@@ -28,11 +24,17 @@ source lib/images.sh
...
+if [[ $OS == ubuntu ]]; then
+  ANSIBLE_FORCE_COLOR=true ansible-playbook \
+    -e "working_dir=$WORKING_DIR" \
+    -e "metal3_dir=$SCRIPTDIR" \
+    -e "virthost=$HOSTNAME" \
+    -i vm-setup/inventory.ini \
+    -b -vvv vm-setup/install-package-playbook.yml
+else
+  # shellcheck disable=SC1091
+  source centos_install_requirements.sh
```
fmuyassarov, line 36: I will move centos one as well in a separate PR.

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -0,0 +1,101 @@
...
+        group: root
+        mode: '0644'
+
+    - name: Fetch OS release key
+      apt_key:
+        url: "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_{{ OS_VERSION_ID }}/Release.key"
+        state: present
+
+    - name: Update apt
+      apt: update_cache=yes name=docker-ce state=latest
```
kashifest, line 36: Same comment here as above

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -0,0 +1,101 @@
...
+      block:
+        - name: Install podman
+          apt: name=podman
+          
+        - name: Default registries configuration file
+          blockinfile:
+            path: /etc/containers/registries.conf
+            block: |
+              [registries.search]
+              registries = ['docker.io', 'quay.io']
```
kashifest, line 62: Could the right side of the `=` be replaced by env var like,  `${REGISTRY}` in the original .sh file?

vm-setup/roles/packages_installation/defaults/main.yml
```diff
@@ -22,15 +26,23 @@ packages:
         - libpolkit-agent-1-0
         - libpolkit-gobject-1-0
         - apache2-utils
+        - apt-transport-https
+        - ca-certificates
+        - gnupg-agent
+        - software-properties-common
```
kashifest, line 32: these packages are not needed when `CONTAINER_RUNTIME=podman`

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -0,0 +1,101 @@
...
+      when: CONTAINER_RUNTIME == "podman"
+
+    - name: Add Docker’s GPG key
+      apt_key:
+        url: https://download.docker.com/linux/ubuntu/gpg
+        state: present
+
+    - name: Add Docker Repository
+      apt_repository:
+        repo: deb https://download.docker.com/linux/ubuntu bionic stable
```
kashifest, line 73: Would it be bionic also in case of `focal`? Shouldn't we variablize this?

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -0,0 +1,101 @@
...
+        state: present
+
+    - name: Update apt
+      apt: update_cache=yes name=docker-ce state=latest
```
kashifest, line 77: Same comment as above

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -0,0 +1,101 @@
...
+        name: docker
+        state: started
+        enabled: yes
+
+    - name: Add current user to the docker group
+      user:
+        name: "{{ lookup('env','USER') }}"
+        groups: docker
+        append: yes
+  become: yes
```
kashifest, line 101: Line 66- 101 , these modules are not needed when `CONTAINER_RUNTIME=podman `

kashifest (changes requested):
Overall this is heading to right direction. Queries Inline

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -0,0 +1,101 @@
+---
+- name: Install required Ubuntu packages
+  block:
+    - name: Update apt
+      apt: update_cache=yes name=docker-ce state=latest
```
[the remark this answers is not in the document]
fmuyassarov: Good catch. It was updating only docker packages but not the rest. I replace docker-ce with "*"

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
fmuyassarov: Yes of course. But interestingly that's why I got after running make in a environment with container runtime set to podman. I'm not sure if it is the issue with the script that it is not setting REGISTRY in the list properly or something else. I could for example do like this:
```
name: Template registries configuration             
template: 
        src: "{{ DAEMON_JSON_PATH }}/registries.json"
        dest: /etc/containers/registries.conf
        owner: root
        group: root
# {{ DAEMON_JSON_PATH }}/registries.json before the templating

cat registries.json
 [registries.search]
 registries = ['docker.io', 'quay.io', "{{ REGISTRY }}"]
```

fmuyassarov: and then when we run template task, the output would be

```
 [registries.search]
 registries = ['docker.io', 'quay.io', '192.168.111.1:5000']
```

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
fmuyassarov: Good point. I can make it parameterized to check what is the current OS release. Something like `$(lsb_release -cs)`

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
fmuyassarov: Fixed. Thanks

vm-setup/roles/packages_installation/defaults/main.yml
fmuyassarov: Fixed.

vm-setup/roles/packages_installation/defaults/main.yml
```diff
@@ -2,13 +2,17 @@ packages:
   ubuntu:
     common:
       packages:
+        - python3-setuptools
+        - zlib1g-dev
+        - openssh-server
+        - wget
+        - libssl-dev
```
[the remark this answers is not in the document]
fmuyassarov: Sorry, missed that.

fmuyassarov:
@kashifest Please have a look at the current state. Thanks for the review.

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -0,0 +1,105 @@
...
+        url: "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_{{ OS_VERSION_ID }}/Release.key"
+        state: present
+
+    - name: Update all packages to their latest version
+      apt: update_cache=yes name="*" state=latest
+
+    - name: Set update-alternatives to python3
+      community.general.alternatives:
+        name: python
+        path: /usr/bin/python3.6
```
furkatgofurov7, line 41: original script had specific versions for python (3.6.1 & 3.8.1), would not be a problem in this case?

vm-setup/roles/packages_installation/defaults/main.yml
```diff
@@ -100,3 +115,8 @@ packages:
          - firewalld
          - python3-six
          - httpd-tools
+
+DAEMON_JSON_PATH: "{{ metal3_dir }}/vm-setup/roles/packages_installation/files"
```
furkatgofurov7, line 119: would it be sufficient not to point to the exact file name in this case?

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -0,0 +1,105 @@
...
+        line: deb https://dl.yarnpkg.com/debian/ stable main
+        create: yes
+        owner: root
+        group: root
+        mode: '0644'
+
+    - name: Add OS release key
+      lineinfile:
+        path: /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
+        line: "deb https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_{{ OS_VERSION_ID }}/ /"
```
furkatgofurov7, line 24: would be good for visibility if we could break the long line

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -0,0 +1,105 @@
...
+        create: yes
+        owner: root
+        group: root
+        mode: '0644'
+
+    - name: Fetch OS release key
+      apt_key:
+        url: "https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_{{ OS_VERSION_ID }}/Release.key"
```
furkatgofurov7, line 32: same here if possible

furkatgofurov7 (commented):
Good work! Mostly questions/suggestions inline.

vm-setup/roles/packages_installation/defaults/main.yml
fmuyassarov: Could be. I think both are fine. Of course I can change it, but I don't see what would be the difference.

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
fmuyassarov: I did in my local tests, and unfortunately was failing. I would rather keep it as it is now. And it is a URL link, that's why difficult to break.

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
fmuyassarov: I think no, since it passed the CI. But I will add a condition. Thanks.

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
fmuyassarov: Please see the comment above.

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
fmuyassarov: Addition. 
It wasn't 3.8.1 but 3.8 1 previously. And this has failed when I added 3.8 1. That's why I made 3.6/3.8 instead of 3.6 1/3.8 1.

fmuyassarov: I think it is fine to keep it like this @furkatgofurov7  ?

furkatgofurov7: My bad, I ignored reading the whole line and missed missing .(dot). I think it is fine as it is, sorry for the confusion.

fmuyassarov: No worries. I had the same confusion too :).

vm-setup/roles/packages_installation/defaults/main.yml
furkatgofurov7: Makes sense

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
fmuyassarov: Fixed

fmuyassarov:
@kashifest , @furkatgofurov7 is the current state ok for you ?

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -0,0 +1,113 @@
...
+            name: docker
+            state: started
+            enabled: yes
+
+        - name: Add current user to the docker group
+          user:
+            name: "{{ lookup('env','USER') }}"
+            groups: docker
+            append: yes
+      when: CONTAINER_RUNTIME != "podman"
```
maelk, line 111: I would rather check for docker here, to be a bit more explicit

maelk (approved):

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
fmuyassarov: Done

PR #563, opened 2020-11-23
Remove Centos7 support

fmuyassarov:
Stop Centos7 support in Metal3-dev-env and remove its related scripts.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -42,7 +42,18 @@ ROOT_DISK_NAME=${ROOT_DISK_NAME-"/dev/sda"}
 # Hostname format
 NODE_HOSTNAME_FORMAT=${NODE_HOSTNAME_FORMAT:-"node-%d"}
 # Check OS type and version
-OS=$(awk -F= '/^ID=/ { print $2 }' /etc/os-release | tr -d '"')
+# shellcheck disable=SC1091
+source /etc/os-release
+export DISTRO="${ID}${VERSION_ID%.*}"
+export OS="${ID}"
+export OS_VERSION_ID=$VERSION_ID
+export SUPPORTED_DISTROS=(centos8 rhel8 ubuntu18 ubuntu20)
+
+if [[ ! "${SUPPORTED_DISTROS[*]}" =~ $DISTRO ]]; then
+   echo "Supported OS distros for the host are: CentOS8 or RHEL8 or Ubuntu20.04"
+   exit 1
+fi
+
 # Container runtime
 if [[ "${OS}" == ubuntu ]]; then
   export CONTAINER_RUNTIME=${CONTAINER_RUNTIME:-"docker"}
@@ -58,7 +69,6 @@ else
   export POD_NAME_INFRA=""
 fi
 
-export OS
 export SSH_KEY=${SSH_KEY:-"${HOME}/.ssh/id_rsa"}
 export SSH_PUB_KEY=${SSH_PUB_KEY:-"${SSH_KEY}.pub"}
 # Generate user ssh key
@@ -214,35 +224,9 @@ if ! sudo -n uptime &> /dev/null ; then
   exit 1
 fi
 
-OS_VERSION=$(awk -F= '/^VERSION_ID=/ { print $2 }' /etc/os-release | tr -d '"' | cut -f1 -d'.')
-OS_VERSION_ID=$(awk -F= '/^VERSION_ID=/ { print $2 }' /etc/os-release | tr -d '"')
-export OS_VERSION
-export OS_VERSION_ID
-if [[ $OS == centos ]]; then
-  if [[ ${OS_VERSION} != 7 && ${OS_VERSION} != 8 ]]; then
-    echo "Required CentOS 7/8 or RHEL 8 or Ubuntu 20.04"
-    exit 1
-  fi
-elif [[ $OS == rhel ]]; then
-  if [[ ${OS_VERSION} -ne 8 ]]; then
-    echo "Required CentOS 7/8 or RHEL 8 or Ubuntu 20.04"
-    exit 1
-  fi
-elif [[ $OS == ubuntu ]]; then
-  if [[ ${OS_VERSION_ID} != "18.04" && ${OS_VERSION_ID} != "20.04" ]]; then
-    # Ubuntu 18.04 is supported, but the target VM cannot use UEFI, hence we are deprecating it
-    echo "Required CentOS 7/8 or RHEL 8 or Ubuntu 20.04"
-    exit 1
-  fi
-else
-  echo "Unsupported OS: $OS"
-  exit 1
-fi
-
 # Use firewalld on CentOS/RHEL, iptables everywhere else
 export USE_FIREWALLD=False
-if [[ ($OS == "rhel" || $OS = "centos") && ${OS_VERSION} == 8 ]]
-then
+if [[ $DISTRO == "rhel8" || $DISTRO == "centos8" ]]; then
   export USE_FIREWALLD=True
 fi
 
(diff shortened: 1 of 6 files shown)


vm-setup/roles/packages_installation/defaults/main.yml
```diff
@@ -38,41 +38,10 @@ packages:
...
-         - libvirt-devel
-         - libvirt-daemon-kvm
-         - libvirt-python
-         - libguestfs-tools
-         - python-lxml
-         - polkit-pkla-compat
-         - python-netaddr
-         - virt-install
-         - python-six
-         - httpd-tools
```
kashifest, line 70: some packages were python only, so we don't need to support them anymore I guess, because in common its only python3?

vm-setup/roles/packages_installation/tasks/main.yml
```diff
@@ -4,32 +4,33 @@
...
       package:
         name: "{{ packages.ubuntu.focal.packages}}"
         state: present
-      become: true
     - name: Install packages using pip3
       pip:
         executable: "pip3"
         name: "{{ packages.ubuntu.pip3 }}"
         state: present
-      become: true
```
kashifest, line 25: In your other PR I have seen you changing `true` to `yes`, should we be consistent?

kashifest (commented):
This looks good to me. Some small questions inline

vm-setup/roles/packages_installation/defaults/main.yml
fmuyassarov: I'm not quite sure. Because, python2 was shipped with Centos7. And we still have conditional checks for Centos7.
And I think it was needed for some other software ... but I again I'm  not 100% sure who is using them. We could of course deprecate them but maybe after we depreciate the support for Centos7 ?.

vm-setup/roles/packages_installation/tasks/main.yml
fmuyassarov: Sure, will fix it. Thanks.

vm-setup/roles/packages_installation/defaults/main.yml
kashifest: This can be answered by the community specially the Centos users. Meanwhile we can atleast keep python packages since you are completely removing them. I don't know whats the best way. @russellb @hardys do you guyz have any insight here?

fmuyassarov: Actually that's my bad. I didn't realize that I added 3 in front of python :). I would rather keep as it it now
```
         - libvirt-python
         - python-lxml
         - python-netaddr
         - python-six
```

fmuyassarov: Would ^ be okay for now @kashifest ? and if someone from the community with more insights on Centos confirms, then maybe we can do it with python3 ?

hardys: I'm not aware of anyone still testing with centos/RHEL7 so it may be best to de-duplicate this by deprecating centos7 support then removing the package list?

For now I don't think we can de-duplicate, because as you mention the python packages are named differently due to the change to python3 in centos8.

vm-setup/roles/packages_installation/tasks/main.yml
```diff
@@ -4,32 +4,33 @@
...
 
 
 - name: Install packages on CentOS
   block:
     - name: Install packages using standard package manager
       block:
+        - name: Install common packages using standard package manager for Centos
+          package:
+            name: "{{ packages.centos.common.packages }}"
+            state: present
```
hardys, line 33: If we do this it would be better to join the lists as it will be faster than running two install tasks

hardys:
Assuming this is just for cosmetic reasons I'd prefer not to do this - the python package names are not interchangeable so perhaps it'd be better to deprecate centos7 support, then remove it (which will remove the duplicate package list, and we can potentially also remove some other things e.g iptables support?

fmuyassarov:
> Assuming this is just for cosmetic reasons I'd prefer not to do this - the python package names are not interchangeable so perhaps it'd be better to deprecate centos7 support, then remove it (which will remove the duplicate package list, and we can potentially also remove some other things e.g iptables support)?

Thanks for the reply @hardys . I'm +1 for dropping Centos7.  And if everyone is okay with that, I can do the cleaning part too.

kashifest:
+1 for removing centos 7 support

jan-est:
+1 for removing centos 7 support

fmuyassarov:
/hold
Until I remove other necessary parts

fmuyassarov:
I have modified existing condition to be like this
```
if [[ $DISTRO != "centos8" && $DISTRO != "rhel8" && $DISTRO != "ubuntu20" && $DISTRO != "ubuntu18"  ]]; then
    echo "Supported OS distros for the host are: CentOS8 or RHEL8 or Ubuntu20.04"
    exit 1
fi
```
so that if a user tries to run Metal3-dev-env on Centos7, it will exit with the small text. (I hope that's what you meant @hardys ).
However, I didn't remove the iptables scripts, because we still need that for the Ubuntu cases. I think we can move it to firewalld for Ubuntu as well, but I would like to do it in a separate PR.

/test-centos-integration
/test-integration
/hold cancel

lib/common.sh
```diff
@@ -41,8 +41,20 @@ INT_IF=${INT_IF:-}
...
+
 # Check OS type and version
-OS=$(awk -F= '/^ID=/ { print $2 }' /etc/os-release | tr -d '"')
+# shellcheck disable=SC1091
+source /etc/os-release
+export DISTRO="${ID}${VERSION_ID%.*}"
+export OS="${ID}"
+export OS_VERSION_ID=$VERSION_ID
+
+if [[ $DISTRO != "centos8" && $DISTRO != "rhel8" && $DISTRO != "ubuntu20" && $DISTRO != "ubuntu18"  ]]; then
```
elfosardo, line 52: since we're here, I think would be much cleaner if we moved the supported distributions to a variable, like:
SUPPORTED_DISTROS="centos8|rhel8|ubuntu20|ubuntu18"
and then let the test evaluate if DISTRO is not in the list using the =~ operator

elfosardo (changes requested):

lib/common.sh
fmuyassarov: Great, now I got what Steven was saying :). Thanks @elfosardo .

maelk:
Should we rather first introduce a warning, before removing everything at once as done in this PR. This is full-on removal of the feature right now. We could put a deprecation notice for a couple of weeks, exiting if it is on CentOS 7 , but allowing to give an option to force deploying in CentOS 7.  Then removing the CentOS 7 later, to give possible users a warning period and some time to migrate.
But maybe we can consider that Centos7 should anyways not be in use anymore and merge this PR as is.

fmuyassarov:
> Should we rather first introduce a warning, before removing everything at once as done in this PR. This is full-on removal of the feature right now. We could put a deprecation notice for a couple of weeks, exiting if it is on CentOS 7 , but allowing to give an option to force deploying in CentOS 7. Then removing the CentOS 7 later, to give possible users a warning period and some time to migrate.
> But maybe we can consider that Centos7 should anyways not be in use anymore and merge this PR as is.

I think for the user that have already cloned metal3-dev-env they still will be able to use it for a while with CentOS7. But those who will try from the latest code base will need to have CentOS 8. There is anyways no way to migrate, since we don't have upgrading capabilities.
So, I would be +1 for dropping it now, but of course I will agree with what the community says.

fmuyassarov:
/hold 
To give some time for folks to reply on the Mailing list related to this PR, and share if they have any objections not to remove the CentOS7 support.

fmuyassarov:
/cancel hold Since there were no objections against so far.

maelk (commented):
/lgtm

PR #565, opened 2020-11-24
Couple of fixes related to #561

fmuyassarov:
This is related to https://github.com/metal3-io/metal3-dev-env/pull/561.
Fixes:
- run apt update before installing any Ubuntu packages
- export lib/network.sh so that `REGISTRY` variable is catched by the Ansible script and the value is added in daemon.json
- set python alternatives to python3 before running any python commands
- use systemd module of Ansible to reload docker daemon


fmuyassarov:
/hold
Until we investigate why REGISTRY variable is not populated in daemon.json file.

lib/network.sh
```diff
@@ -7,6 +7,16 @@
...
+
+# Set update-alternatives to python3
+if [[ ${OS_VERSION_ID} == "18.04" ]]; then
+  sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.6 1
+elif [[ ${OS_VERSION_ID} == "20.04" ]];
+  sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.8 1
+else
+  sudo alternatives --set python /usr/bin/python3
+fi
+
```
kashifest, line 19: These should not be here IMO. This is not related to networking and any installation should be a part of 01 or 02 not lib/ IMO

kashifest (changes requested):
Concern in line

lib/network.sh
fmuyassarov: Fixed.

smoshiur1237 (commented):
I think its in better shape now. I am testing the change is my VM and its working. Lets see the CI run. Looks good to me.

maelk (approved):

01_prepare_host.sh
```diff
@@ -12,13 +12,29 @@ if [[ $(id -u) == 0 ]]; then
 fi
 
 if [[ $OS == ubuntu ]]; then
+  sudo apt-get update
   sudo apt -y install python3-pip
+
+  # Set update-alternatives to python3
```
furkatgofurov7, line 18: There seems to be duplication, do we still need to keep the same configuration here?: https://github.com/metal3-io/metal3-dev-env/blob/deb9086f7e42c49a45265e58bc21e68b6f9f83be/vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml#L42

fmuyassarov: Good catch, thanks. Fixed. I didn't remove the same thing from centos_install*.sh, because that will be done when moving centos_install* script to Ansible.

maelk (approved):
/lgtm

PR #567, opened 2020-11-27
Use remove_local_ironic script from BMO 

kashifest:
Keeping the list of containers to be removed in BMO makes it simpler to maintain. 
This PR should go in after BMO PR# [729](https://github.com/metal3-io/baremetal-operator/pull/729)

### DIFF
--- a/cluster_cleanup.sh
+++ b/cluster_cleanup.sh
@@ -7,6 +7,10 @@ source lib/common.sh
 # Delete cluster
 if [ "${EPHEMERAL_CLUSTER}" == "kind" ]; then
   sudo su -l -c "kind delete cluster  || true" "${USER}"
+  # Kill and remove the running ironic containers
+  if [ -f "$BMOPATH/tools/remove_local_ironic.sh" ]; then
+    "$BMOPATH"/tools/remove_local_ironic.sh
+  fi
 fi
 
 if [ "${EPHEMERAL_CLUSTER}" == "minikube" ]; then
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -449,6 +449,10 @@ function start_management_cluster () {
 # -----------------------------
 
 clone_repos
+
+# Kill and remove the running ironic containers
+"$BMOPATH"/tools/remove_local_ironic.sh
+
 create_clouds_yaml
 if [ "${EPHEMERAL_CLUSTER}" != "tilt" ]; then
   start_management_cluster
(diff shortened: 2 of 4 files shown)


kashifest:
/hold 
until BMO PR#[729](https://github.com/metal3-io/baremetal-operator/pull/729) goes in

fmuyassarov:
lgtm

maelk (approved):

furkatgofurov7 (commented):
/lgtm

maelk (changes requested):
Baremetal operator is only cloned in 03* so any calls to remove_ironic_containers before that will fail with file not found as the CI did

kashifest:
> Baremetal operator is only cloned in 03* so any calls to remove_ironic_containers before that will fail with file not found as the CI did

Is it ok if I move the call to remove ironic containers after the clone in 03?

maelk (changes requested):
Calling make clean will not remove the ironic containers.

kashifest:
> Calling make clean will not remove the ironic containers.

Good catch. Fixed.

cluster_cleanup.sh
```diff
@@ -7,6 +7,8 @@ source lib/common.sh
 # Delete cluster
 if [ "${EPHEMERAL_CLUSTER}" == "kind" ]; then
   sudo su -l -c "kind delete cluster  || true" "${USER}"
+  # Kill and remove the running ironic containers
+  "$BMOPATH"/tools/remove_local_ironic.sh
```
maelk, line 11: That can be problematic, in case you did not run the 03* before calling make clean. Either you should ensure that this will not fail, or you should not call it if the executable is not there

maelk (changes requested):

cluster_cleanup.sh
kashifest: Humm... this is getting complicated!!

kashifest: How do you think about the current change?

cluster_cleanup.sh
```diff
@@ -7,6 +7,10 @@ source lib/common.sh
 # Delete cluster
 if [ "${EPHEMERAL_CLUSTER}" == "kind" ]; then
   sudo su -l -c "kind delete cluster  || true" "${USER}"
+  # Kill and remove the running ironic containers
+  if [ -d "$BMOPATH" ]; then
```
maelk, line 11: I would rather check for the executable itself. you might have an old fork without the executable

kashifest: How about now?

maelk (approved):

fmuyassarov (commented):
/lgtm

PR #570, opened 2020-12-04
Add console=ttyS0 to kernel params

derekhiggins:
In a virtual environment set console=ttyS0 so that the IPA
console can be logged for debugging.

To be used along with
https://github.com/metal3-io/ironic-image/pull/226

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -193,6 +193,10 @@ CACHEURL=http://$IRONIC_HOST/images
 IRONIC_FAST_TRACK=true
 EOF
 
+  if [ "$NODES_PLATFORM" == "libvirt" ] ; then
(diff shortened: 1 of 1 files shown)


03_launch_mgmt_cluster.sh
```diff
@@ -107,6 +107,10 @@ IRONIC_ENDPOINT=${IRONIC_URL}
 IRONIC_INSPECTOR_ENDPOINT=${IRONIC_INSPECTOR_URL}
 EOF
 
+  if [ "$NODES_PLATFORM" == "libvirt" ] ; then
+    echo "IRONIC_KERNEL_PARAMS=console=ttyS0" | sudo tee -a "${BMOPATH}/config/default/ironic.env"
+  fi
+
```
maelk, line 113: In this case we are not deploying ironic, only BMO, so this is not needed here (See the "true false" on the following line)

derekhiggins: thanks, fixed

fmuyassarov (approved):

PR #573, opened 2020-12-07
Fix bug regarding Ironic deployment in feature test

kashifest:
Feature test ironic deletion and re-creation was not synchronized for centos based feature test. This PR fixes the issue.


fmuyassarov (commented):
/lgtm

PR #574, opened 2020-12-07
Migrate centos_install_req*.sh from bash to Ansible

fmuyassarov:
Move centos_install_requirements.sh to an Ansible script.


vm-setup/roles/packages_installation/tasks/centos_required_packages.yml
```diff
@@ -0,0 +1,53 @@
+# Perform CentOS/RHEL8 related configurations
+- name: Example using fail and when together
```
namnx228, line 2: Look like a copy-paste error.

namnx228 (commented):
Look good to me. Just one small nit.

vm-setup/roles/packages_installation/tasks/main.yml
```diff
@@ -31,10 +31,13 @@
...
-  package:
-      name: "{{ packages.centos.rhel8.packages }}"
-      state: present
+- name: Install packages on CentOS/RHEL8
+  block:
+    - name: Install packages on CentOS/RHEL8
+      package:
+          name: "{{ packages.centos.rhel8.packages }}"
+          state: present
+    - name: Install required packages for Ubuntu
```
furkatgofurov7, line 40: For Ubuntu? Not CentOS?

furkatgofurov7 (commented):
/lgtm

vm-setup/roles/packages_installation/tasks/centos_required_packages.yml
fmuyassarov: D'oh, good catch. Thanks.

vm-setup/roles/packages_installation/tasks/main.yml
fmuyassarov: Ops. True, should be CentOS.

centos_install_requirements.sh
```diff
@@ -1,54 +0,0 @@
-#!/usr/bin/env bash
-set -ex
-
-# shellcheck disable=SC1091
-source lib/logging.sh
-# shellcheck disable=SC1091
-source lib/common.sh
-
-sudo dnf install -y libselinux-utils
```
furkatgofurov7, line 9: I think this is missing in package [list](https://github.com/metal3-io/metal3-dev-env/blob/c6f9b23c0906a068dded9056f10b9acf5d3c25bb/vm-setup/roles/packages_installation/defaults/main.yml#L56)?

fmuyassarov: This step is done via
```
- name: Enable SELinux
  ansible.posix.selinux:
    policy: targeted
    state: permissive
  when: ansible_selinux.status == "enabled"
```

centos_install_requirements.sh
```diff
@@ -1,54 +0,0 @@
...
-    sudo alternatives --set python /usr/bin/python3
-fi
-
-# Install required packages
-sudo dnf -y install \
-  redhat-lsb-core \
-  python3-pip \
-  wget
-
-sudo pip3 install ansible
```
namnx228, line 41: If we remove this, how can Ansible be installed?

fmuyassarov: We are installing it in 01* script https://github.com/metal3-io/metal3-dev-env/blob/c6f9b23c0906a068dded9056f10b9acf5d3c25bb/01_prepare_host.sh#L30

kashifest:
lgtm, the centos failure needs to be investigated

vm-setup/roles/packages_installation/tasks/centos_required_packages.yml
```diff
@@ -0,0 +1,53 @@
...
+
+- name: Upgrade all packages
+  dnf:
+    name: "*"
+    state: latest
+
+# Install additional repos as needed for each OS version
+- name: Enable external repository
+  dnf:
+    name: extras
```
furkatgofurov7, line 28:
```suggestion
    name: epel-release
```

furkatgofurov7 (commented):
CI is failing with: 
`"failures": [[0m
[0;31m        "No package extras available."`
Maybe this suggestion could work?

vm-setup/roles/packages_installation/tasks/centos_required_packages.yml
fmuyassarov: Thanks @furkatgofurov7 . Will try that!

hardys: Note that epel isn't the same as extras, and there is some package overlap with different versions ref http://mirror.centos.org/centos/8/extras/x86_64/os/Packages/

I wonder if we need this repo at all, since the main dependency there seems to be ansible, which we're now [installing via pip](https://github.com/metal3-io/metal3-dev-env/blob/master/01_prepare_host.sh#L30)?

fmuyassarov: Good catch. Removed it.

PR #583, opened 2021-01-12
Optimize selinux configuration

MaXinjian:
selinuxenabled exits with 0 when selinux is in permissive or enforcing status.
when selinux is already in permissive status, we don't need to configure selinux

Signed-off-by: Ma Xinjian <maxj.fnst@cn.fujitsu.com>

### DIFF
--- a/centos_install_requirements.sh
+++ b/centos_install_requirements.sh
@@ -7,7 +7,7 @@ source lib/logging.sh
 source lib/common.sh
 
 sudo dnf install -y libselinux-utils
-if selinuxenabled ; then
+if getenforce | grep Enforcing; then
     sudo setenforce permissive
     sudo sed -i "s/=enforcing/=permissive/g" /etc/selinux/config
 fi


centos_install_requirements.sh
```diff
@@ -7,7 +7,7 @@ source lib/logging.sh
 source lib/common.sh
 
 sudo dnf install -y libselinux-utils
-if selinuxenabled ; then
+if getenforce | grep Enforcing; then
```
fmuyassarov, line 10: Wouldn't be better to check  SELinux mode explicitly against permissive?
```
if  !  getenforce | grep  Permissive; then
// do the settings
fi
```

MaXinjian: Only when getenforce  returns Enforcing, the settings are needed. when getenforce returns Disabled. the settings are not needed.

fmuyassarov (approved):

fmuyassarov:
@MaXinjian note thhat `centos_install_requirements.sh` has been moved to ansible script in https://github.com/metal3-io/metal3-dev-env/pull/574.

MaXinjian:
Close this PR  since target file has been removed

PR #585, opened 2021-01-19
Fix bare variable evaluation deprecation issue

MaXinjian:
Fix the following warning:
  evaluating 'vm_nodes' as a bare variable, this behaviour will go away and you might need to add |bool to the expression in the future. Also see CONDITIONAL_BARE_VARS

Signed-off-by: Ma Xinjian <maxj.fnst@cn.fujitsu.com>

### DIFF
--- a/vm-setup/roles/libvirt/tasks/network_setup_tasks.yml
+++ b/vm-setup/roles/libvirt/tasks/network_setup_tasks.yml
@@ -22,7 +22,7 @@
     nodes: "{{ vm_nodes }}"
     networks: "{{ networks }}"
   register: node_mac_map
-  when: vm_nodes
+  when: vm_nodes | length > 0
 
 # Create the global, root-managed libvirt networks to which we will
 # attach the undercoud and vm virtual machines.
--- a/vm-setup/roles/libvirt/tasks/vm_setup_tasks.yml
+++ b/vm-setup/roles/libvirt/tasks/vm_setup_tasks.yml
@@ -53,7 +53,7 @@
     autostart: "yes"
     uri: "{{ libvirt_uri }}"
 
-- when: vm_nodes
+- when: vm_nodes | length > 0
   environment:
     LIBVIRT_DEFAULT_URI: "{{ libvirt_uri }}"
   block:
--- a/vm-setup/roles/libvirt/tasks/vm_teardown_tasks.yml
+++ b/vm-setup/roles/libvirt/tasks/vm_teardown_tasks.yml
@@ -17,7 +17,7 @@
     LIBVIRT_DEFAULT_URI: "{{ libvirt_uri }}"
   block:
 
-    - when: vm_nodes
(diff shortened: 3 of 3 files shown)


vm-setup/roles/libvirt/tasks/network_setup_tasks.yml
```diff
@@ -22,7 +22,7 @@
     nodes: "{{ vm_nodes }}"
     networks: "{{ networks }}"
   register: node_mac_map
-  when: vm_nodes
+  when: not vm_nodes|bool
```
dtantsur, line 25: Why `not`? Shouldn't this be `when: vm_nodes is defined`?

vm-setup/roles/libvirt/tasks/vm_setup_tasks.yml
```diff
@@ -53,7 +53,7 @@
     autostart: "yes"
     uri: "{{ libvirt_uri }}"
 
-- when: vm_nodes
+- when: not vm_nodes|bool
```
dtantsur, line 56: Same

vm-setup/roles/libvirt/tasks/network_setup_tasks.yml
MaXinjian: when: vm_nodes is ok until ansible 2.12. Now it's just a [DEPRECATION WARNING]
https://github.com/ansible/ansible/issues/53428

maelk:
Isn't the not inverting the effect we want ?

vm-setup/roles/libvirt/tasks/network_setup_tasks.yml
dtantsur: That's not what I'm asking, please see my comment.

hardys: Right `vm_nodes` is a list https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/common/tasks/main.yml#L8

So this should instead be changed to check if the length of that list is non-zero AFAICS?

dtantsur: Yep, something like `vm_nodes | length > 0`

MaXinjian (commented):
> Isn't the not inverting the effect we want ?

Force updated to `when: vm_nodes | length > 0`

MaXinjian:
Thank you all for the patience. 
I have updated to `when vm_nodes | length > 0`

MaXinjian:
> /test-integration
> /test-centos-integration

test-integration failed when 04_verify.sh. But I have compared the jenkins log with passed logs, seems it has no business with this PR.

MaXinjian:
ping

MaXinjian:
ping？

maelk (approved):

PR #587, opened 2021-01-22
Bypass docker pull limit by parameterizing registry

Xenwar:
This PR parameterizes the docker registry for pulling images.
Which registry is used determined by the following variable, with the give default value.

```
export OVERRIDE_DOCKER_REGISTRY_1=${OVERRIDE_DOCKER_REGISTRY_1:-"docker.io"}
```

If you wish to use a different registry, please assign value in a `HOST:PORT/path` format.

Note that when using your own registry, you need to pay attention to the path on which your images are located.
It is not necessarily the same as what is used in `docker.io` or `registry.hub.docker.com`


vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -39,61 +39,21 @@
...
\ No newline at end of file
+    - name: Registries configuration for podman
+      blockinfile:
+        path: /etc/containers/registries.conf
+        create: true
+        block: |
+            unqualified-search-registries = ['registry.nordix.org']
+            
+            [[registry]]
+            prefix = "\registry.nordix.org\"
```
fmuyassarov, line 55: I think we don't need backslashes here since this for source cluster and it will be added by Ansible not cloud-init.

fmuyassarov (changes requested):

maelk (commented):
This seems to remove the insecure configuration for the local registry. This is needed since we use the local registry to store the locally build images (that should not be pulled from nordix registry then)

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
Xenwar: removed

Xenwar:
> This seems to remove the insecure configuration for the local registry. This is needed since we use the local registry to store the locally build images (that should not be pulled from nordix registry then)

```
I agree. And, started to get the following error.
  Normal   Created      5m43s                  kubelet            Created container kube-rbac-proxy
  Normal   Pulling      5m5s (x3 over 6m5s)    kubelet            Pulling image "192.168.111.1:5000/localimages/cluster-api-provider-metal3:master"
  Warning  Failed       4m57s (x3 over 5m58s)  kubelet            Failed to pull image "192.168.111.1:5000/localimages/cluster-api-provider-metal3:master": rpc error: code = Unknown desc = error pinging docker registry 192.168.111.1:5000: Get "https://192.168.111.1:5000/v2/": http: server gave HTTP response to HTTPS client
  Warning  Failed       4m57s (x3 over 5m58s)  kubelet            Error: ErrImagePull
  Normal   BackOff      4m32s (x4 over 5m31s)  kubelet            Back-off pulling image "192.168.111.1:5000/localimages/cluster-api-provider-metal3:master"
  Warning  Failed       75s (x17 over 5m31s)   kubelet            Error: ImagePullBackOff
```

Xenwar:
The problem now is, the now-removed part for adding insecure registry and the new configuration are not compatible (v1 and v2). Now, trying to figure out if there is another way.

fmuyassarov:
> The problem now is, the now-removed part for adding insecure registry and the new configuration are not compatible (v1 and v2). Now, trying to figure out if there is another way.

I think this is because what @maelk mentioned above. You need to add `192.168.111.1:5000` registry as insecure registry.

Xenwar:
> > The problem now is, the now-removed part for adding insecure registry and the new configuration are not compatible (v1 and v2). Now, trying to figure out if there is another way.
> 
> I think this is because what @maelk mentioned above. You need to add `192.168.111.1:5000` registry as insecure registry.

I totally agree. Noted @maelk's comment and working on that.

fmuyassarov:
Can't you try to add one more registry block in the registries.conf as below?
```
          [[registry]]
          prefix = "192.168.111.1:5000"
          insecure = false
          blocked = false
          location = "192.168.111.1:5000"
```

Xenwar:
Added 
```
[[registry]]
prefix = "192.168.111.1:5000"
insecure = true
blocked = false
location = "192.168.111.1:5000"
```

Xenwar:
please do NOT approve or lgtm.

maelk:
Keep in mind also that this should be only for CI. If people are deploying their own metal3-dev-env, we don't want them to always fetch all images from Nordix. They should by default still fetch it from Docker (check discussion from issue). This means that this PR and the ones open on other repos cannot go as is, they need to be parameterized, with a default to normal Dockerhub

Xenwar:
> Keep in mind also that this should be only for CI. If people are deploying their own metal3-dev-env, we don't want them to always fetch all images from Nordix. They should by default still fetch it from Docker (check discussion from issue). This means that this PR and the ones open on other repos cannot go as is, they need to be parameterized, with a default to normal Dockerhub

No need to parameterized, we just change the order of repos. So that it goes do docker and goes to Nordix only when the limit has reached. 
`unqualified-search-registries = ['docker.io', 'registry.nordix.org']
`

lib/common.sh
```diff
@@ -142,8 +142,12 @@ export NUM_OF_WORKER_REPLICAS="${NUM_OF_WORKER_REPLICAS:-"1"}"
 export VM_EXTRADISKS=${VM_EXTRADISKS:-"false"}
 export NODE_DRAIN_TIMEOUT=${NODE_DRAIN_TIMEOUT:-"0s"}
 
+# Registry for pulling images
+export OVERRIDE_DOCKER_REGISTRY_1=${OVERRIDE_DOCKER_REGISTRY_1:-"docker.io"}
+export OVERRIDE_DOCKER_REGISTRY_2=${OVERRIDE_DOCKER_REGISTRY_2:-"registry.hub.docker.com"}
+
 # Docker registry for local images
-export DOCKER_REGISTRY_IMAGE=${DOCKER_REGISTRY_IMAGE:-"docker.io/registry:latest"}
+export DOCKER_REGISTRY_IMAGE=${DOCKER_REGISTRY_IMAGE:-"${OVERRIDE_DOCKER_REGISTRY_1}/library/registry:latest"}
```
smoshiur1237, line 150: I think `/library/` is not needed here. It will work without it.

resources/vbmc/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/centos:centos8
+FROM "${OVERRIDE_DOCKER_REGISTRY_1}/"library/centos:centos8
```
smoshiur1237, line 1: same here. `/library/` is not needed.

smoshiur1237 (changes requested):
Needs small change. Otherwise looks good to me

resources/vbmc/Dockerfile
Xenwar: fixed

lib/common.sh
Xenwar: fixed

namnx228:
lgtm

resources/sushy-tools/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.hub.docker.com/library/python:3.9
+FROM "${OVERRIDE_DOCKER_REGISTRY_2}/"library/python:3.9
```
fmuyassarov, line 1: This is not needed, we are not building it in dev-env.

resources/vbmc/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/centos:centos8
+FROM "${OVERRIDE_DOCKER_REGISTRY_1}/"centos:centos8
```
fmuyassarov, line 1: Same here

hack/shellcheck.sh
```diff
@@ -15,6 +15,6 @@ else
     --volume "${PWD}:/workdir:ro,z" \
     --entrypoint sh \
     --workdir /workdir \
-    registry.hub.docker.com/koalaman/shellcheck-alpine:stable \
+    "${OVERRIDE_DOCKER_REGISTRY_2}/"koalaman/shellcheck-alpine:stable \
```
fmuyassarov, line 18: same here

hack/markdownlint.sh
```diff
@@ -14,6 +14,6 @@ else
     --volume "${PWD}:/workdir:ro,z" \
     --entrypoint sh \
     --workdir /workdir \
-    registry.hub.docker.com/pipelinecomponents/markdownlint:latest \
+    "${OVERRIDE_DOCKER_REGISTRY_2}/"pipelinecomponents/markdownlint:latest \
```
fmuyassarov, line 17: same here

Xenwar: done

hack/shellcheck.sh
Xenwar: done

resources/vbmc/Dockerfile
Xenwar: done

resources/sushy-tools/Dockerfile
Xenwar: done

config_example.sh
```diff
@@ -104,6 +104,10 @@
 
 #export KUBERNETES_VERSION="v1.20.2"
 
+# Registry
+#export OVERRIDE_DOCKER_REGISTRY_1=${OVERRIDE_DOCKER_REGISTRY_1:-"docker.io"}
+#export OVERRIDE_DOCKER_REGISTRY_2=${OVERRIDE_DOCKER_REGISTRY_2:-"registry.hub.docker.com"}
```
maelk, line 109: Those variables are really not explicit. Please do not use _1 and _2 but explicit names. In addition, the _2 seems unused ?

Xenwar: renamed and removed _2.

fmuyassarov (commented):
/lgtm

Xenwar:
A new issue with respect to inheriting host environment variables from with Dockerfile
Given the following entries in the Dockerfile

```
ARG OVERRIDE_DOCKER_REGISTRY_2=${OVERRIDE_DOCKER_REGISTRY_2:-"registry.hub.docker.com"}
FROM "${OVERRIDE_DOCKER_REGISTRY_2}/"library/golang:1.15.3 AS builder
```

It is not always supported that the docker build command inhertis the host environmental variables.
This would require modifying the docker build command as follows.

`docker build —build-args OVERRIDE_DOCKER_HUB_REGISTRY=${OVERRIDE_DOCKER_HUB_REGISTRY:-“my-registry.com"}`

Xenwar:
replaced by an alternative solution.

PR #589, opened 2021-01-24
Fix basic_aut credentials vars of inspector

fmuyassarov:
Currently we are moving `ironic-username` & `ironic-password` to /opt/metal3/auth/ironic-inspector directory. This is okay as long as user doesn't provide custom username/password for inspector. However, if `IRONIC_INSPECTOR_USERNAME` & `IRONIC_INSPECTOR_PASSWORD` are given by the user, then we will ignore the actual provided username/password and instead move `ironic-username` & `ironic-password`.
This PR adds `IRONIC_INSPECTOR_USERNAME` & `IRONIC_INSPECTOR_PASSWORD` in the mv command to make sure that we don't ignore user provided credentials.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -130,8 +130,8 @@ EOF
       cp "${IRONIC_AUTH_DIR}ironic-password" /opt/metal3/auth/ironic/password
       sudo mkdir -p /opt/metal3/auth/ironic-inspector
       sudo chown "$USER":"$USER" /opt/metal3/auth/ironic-inspector
-      cp "${IRONIC_AUTH_DIR}ironic-username" /opt/metal3/auth/ironic-inspector/username
-      cp "${IRONIC_AUTH_DIR}ironic-password" /opt/metal3/auth/ironic-inspector/password
+      cp "${IRONIC_AUTH_DIR}${IRONIC_INSPECTOR_USERNAME}" /opt/metal3/auth/ironic-inspector/username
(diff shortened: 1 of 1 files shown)


03_launch_mgmt_cluster.sh
```diff
@@ -130,8 +130,8 @@ EOF
       cp "${IRONIC_AUTH_DIR}ironic-password" /opt/metal3/auth/ironic/password
       sudo mkdir -p /opt/metal3/auth/ironic-inspector
       sudo chown "$USER":"$USER" /opt/metal3/auth/ironic-inspector
-      cp "${IRONIC_AUTH_DIR}ironic-username" /opt/metal3/auth/ironic-inspector/username
-      cp "${IRONIC_AUTH_DIR}ironic-password" /opt/metal3/auth/ironic-inspector/password
```
jan-est, line 134: In case where user does not give username and password, how the pw and username is copied if these lines are removed?

fmuyassarov: Yes. For example, if `IRONIC_INSPECTOR_USERNAME` is not provided by the user, it will [default](https://github.com/metal3-io/metal3-dev-env/blob/40e3f79b78d4f317e117b98731230a7c929f5307/lib/ironic_basic_auth.sh#L25) to IRONIC_USERNAME, which in turn generates ironic-username [here](https://github.com/metal3-io/metal3-dev-env/blob/40e3f79b78d4f317e117b98731230a7c929f5307/lib/ironic_basic_auth.sh#L12).

smoshiur1237:
Looks good to me.

PR #590, opened 2021-01-25
K8s version upgrade from 1.20.0 to 1.20.2 for upgrade workflow.

smoshiur1237:
Kubernetes version change for upgrade workflow. Also added a fix for unqualified image error: ImageInspectError issue.


fmuyassarov (commented):
LGTM

namnx228:
lgtm

kashifest (approved):
/test-upgrade-features

kashifest (commented):
The kubernetes version is now by default 1.20.2, so your code will not work since 1.20.2 is not < than 1.20.2

smoshiur1237:
> The kubernetes version is now by default 1.20.2, so your code will not work since 1.20.2 is not < than 1.20.2

Yes you are right. I don't have any option to upgrade from 1.20.2 to 1.20.3. We need to wait for newer version of the kubernetes.

fmuyassarov (approved):

PR #596, opened 2021-02-03
Prepull container images hosted in docker hub

fmuyassarov:
In Nordix we are building Jenkins CI images on a daily bases. As part building those images we are running 01* script. This patch ensures that we download all the container images (hosted in Docker hub) needed during integration test in dev-env.
Fixes: https://github.com/metal3-io/project-infra/issues/166


fmuyassarov:
We have decided to prepull container images within Nordix/airship-dev-tools scripts.
/close

PR #599, opened 2021-02-11
Allow setting the lease expiry for DHCP

cybertron:
Newer versions of libvirt (6.3.0 and up) allow the lease expiry to
be set for hosts. We have some functionality that depends on using
infinite leases, so we would like to be able to set this through
metal3-dev-env.

### DIFF
--- a/vm-setup/roles/libvirt/templates/network.xml.j2
+++ b/vm-setup/roles/libvirt/templates/network.xml.j2
@@ -41,7 +41,9 @@
         {% set ironic_name = ironic_prefix + flavor + "_" + num|string %}
         {% set hostname_format = lookup('vars', flavor + '_hostname_format', default=flavor + '-%d') %}
         {% set hostname = hostname_format % num %}
-      <host mac='{{ node_mac_map.get(ironic_name).get(item.name)}}' name='{{hostname}}' ip='{{item.dhcp_range_v4[0]|ipmath(ns.index|int)}}'/>
+      <host mac='{{ node_mac_map.get(ironic_name).get(item.name)}}' name='{{hostname}}' ip='{{item.dhcp_range_v4[0]|ipmath(ns.index|int)}}'>
+        <lease expiry='{{ item.lease_expiry }}'/>
+      </host>
         {% set ns.index = ns.index + 1 %}
       {% endfor %}
     {% endfor %}
@@ -88,7 +90,9 @@
         {% set ironic_name = ironic_prefix + flavor + "_" + num|string %}
         {% set hostname_format = lookup('vars', flavor + '_hostname_format', default=flavor + '-%d') %}
         {% set hostname = hostname_format % num %}
-        <host id='00:03:00:01:{{ node_mac_map.get(ironic_name).get(item.name)}}' name='{{hostname}}' ip='{{item.dhcp_range_v6[0]|ipmath(ns.index|int)}}'/>
+        <host id='00:03:00:01:{{ node_mac_map.get(ironic_name).get(item.name)}}' name='{{hostname}}' ip='{{item.dhcp_range_v6[0]|ipmath(ns.index|int)}}'>
+          <lease expiry='{{ item.lease_expiry }}'/>
+        </host>
         {% set ns.index = ns.index + 1 %}
       {% endfor %}
     {% endfor %}
--- a/vm-setup/roles/common/defaults/main.yml
+++ b/vm-setup/roles/common/defaults/main.yml
@@ -64,6 +64,8 @@ networks:
     dhcp_range_v6:
       - "{{ virsh_dhcp_v6_start }}"
       - "{{ virsh_dhcp_v6_end }}"
+    # libvirt defaults to minutes as the unit
+    lease_expiry: 60
     nat_port_range:
       - 1024
       - 65535


hardys:
I triggered test-integration which may confirm - but do we know what libvirt version is expected in the supported Ubuntu release?

vm-setup/roles/libvirt/templates/network.xml.j2
```diff
@@ -41,7 +41,9 @@
         {% set ironic_name = ironic_prefix + flavor + "_" + num|string %}
         {% set hostname_format = lookup('vars', flavor + '_hostname_format', default=flavor + '-%d') %}
         {% set hostname = hostname_format % num %}
-      <host mac='{{ node_mac_map.get(ironic_name).get(item.name)}}' name='{{hostname}}' ip='{{item.dhcp_range_v4[0]|ipmath(ns.index|int)}}'/>
+      <host mac='{{ node_mac_map.get(ironic_name).get(item.name)}}' name='{{hostname}}' ip='{{item.dhcp_range_v4[0]|ipmath(ns.index|int)}}'>
+        <lease expiry='{{ item.lease_expiry }}'/>
```
hardys, line 45: Should this be conditional on item.lease_expiry being defined?

fmuyassarov:
> I triggered test-integration which may confirm - but do we know what libvirt version is expected in the supported Ubuntu release?

Description above says we need libvirt 6.3.0+. I just checked from the CI, and Ubuntu 20.04 based CI shows that we have 6.0.0 version, 
```
ubuntu@ci-test-vm-20210225133921-xphj:~$ sudo virsh --version
6.0.0
```
which is installed [here](https://github.com/metal3-io/metal3-dev-env/blob/c6f9b23c0906a068dded9056f10b9acf5d3c25bb/vm-setup/roles/packages_installation/defaults/main.yml#L2). I'm wondering if we can install [7.0.0](https://www.libvirt.org/news.html#v7-0-0-2021-01-15) version on Ubuntu by specifying the version implicitly during installation.

cybertron (commented):
I've been using this locally for a while now and have noticed a couple of odd things with the advanced-virtualization version of libvirt. Neither prevent use of it, but I'm wondering if we don't want to force installation of the newer version and just document what version of libvirt is needed for this particular feature? That would save us from having to deal with the distro-specific methods of getting the correct version.

maelk:
Could you please elaborate on the use case you are trying to cover ? as is now, there would not be any choice from the user point of view (except if the user is using the roles directly, not via the high level scripts). Would it make sense to introduce a configuration variable and pass it to ansible ?

cybertron:
I'm looking to support setting infinite leases from dev-scripts. I have a patch up that makes use of this configuration: https://github.com/openshift-metal3/dev-scripts/pull/1191

I'm not sure how broadly applicable this is. It's for an OpenShift feature that does some special handling of infinite leases, which I think is a little unusual. If you think others might want to use it I can wire something in though.

maelk:
If it is openshift specific, then maybe it should not be added here

cybertron:
OpenShift is based on metal3 though. Without this we can't use metal3-dev-env to do development and testing.

hardys:
@cybertron IIRC this is actually testing a feature of NetworkManager, so it's not specific to OpenShift (although the requirement originated there)?

@maelk the [dev-scripts](https://github.com/openshift-metal3/dev-scripts/) repo has some dependencies on metal3-dev-env, specifically `centos_install_requirements.sh` and the ansible roles - we could fork those pieces but the idea is to reduce duplicated effort when we need basically the exact same test setup for upstream metal3 testing, and metal3-enabled OpenShift.

fmuyassarov:
> @cybertron IIRC this is actually testing a feature of NetworkManager, so it's not specific to OpenShift (although the requirement originated there)?
> 
> @maelk the [dev-scripts](https://github.com/openshift-metal3/dev-scripts/) repo has some dependencies on metal3-dev-env, specifically `centos_install_requirements.sh` and the ansible roles - we could fork those pieces but the idea is to reduce duplicated effort when we need basically the exact same test setup for upstream metal3 testing, and metal3-enabled OpenShift.

Note that `centos_install_requirements.sh` has been replaced by Ansible [script](https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/packages_installation/tasks/centos_required_packages.yml) in https://github.com/metal3-io/metal3-dev-env/pull/574.

cybertron:
I've removed the requirements part entirely. I'm not sure we should be forcing people to pull in a newer version of libvirt if they don't need it. Older releases of libvirt will just drop the lease configuration, and anyone who wants to use custom lease expirations can make sure they have the necessary libvirt themselves. That way it's a conscious choice, not something that might come as a surprise.

cybertron:
Hmm, the centos job seems to have gotten stuck.

/test-centos-integration

furkatgofurov7:
Hi @cybertron seems there was a glitch in CI by that time, re-triggering all tests once again since we have had changes recently to run integration tests.  

/test-centos-integration
/test-integration
/test-v1a4-centos-integration
/test-v1a4-integration

furkatgofurov7 (commented):
/lgtm

cybertron:
> Hi @cybertron seems there was a glitch in CI by that time, re-triggering all tests once again since we have had changes recently to run integration tests.

Thanks, I've been meaning to ask about this. Looks like it's still working!

PR #604, opened 2021-02-19
Modify default_vcpu and use minimum resources in metal3-dev-env

smoshiur1237:
We are using` default_vcpu= 4` for each BMH(control plane and worker) in metal3-dev-env. This PR will edit the `default_vcpu=2` to limit the use of the resources.


fmuyassarov:
/approve
Let's keep an eye on the CI to see if there is any performance degradation in terms of runtime.

smoshiur1237:
> /approve
> Let's keep an eye on the CI to see if there is any performance degradation in terms of runtime.

I have checked the CI run on this PR. Performance and runtime is almost same.

PR #606, opened 2021-03-02
Bump sushy-tools and allow to ignore boot device

hardys:
The latest release adds a variable that allows us to test setting
the boot device via efibootmgr, when combined with the live-iso
disk format this can be used to install e.g using the Fedora
CoreOS iso.

https://review.opendev.org/c/openstack/releases/+/778023
https://review.opendev.org/c/openstack/sushy-tools/+/776660

### DIFF
--- a/vm-setup/roles/virtbmc/tasks/setup_tasks.yml
+++ b/vm-setup/roles/virtbmc/tasks/setup_tasks.yml
@@ -96,4 +96,5 @@
     dest: "{{ working_dir }}/virtualbmc/sushy-tools/conf.py"
     content: |
       SUSHY_EMULATOR_LIBVIRT_URI = "{{ vbmc_libvirt_uri }}"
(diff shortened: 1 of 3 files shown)


iurygregory (approved):

iurygregory (approved):

elfosardo (approved):

vm-setup/roles/virtbmc/tasks/setup_tasks.yml
```diff
@@ -96,4 +96,5 @@
     dest: "{{ working_dir }}/virtualbmc/sushy-tools/conf.py"
     content: |
       SUSHY_EMULATOR_LIBVIRT_URI = "{{ vbmc_libvirt_uri }}"
+      SUSHY_EMULATOR_IGNORE_BOOT_DEVICE = "{{ sushy_ignore_boot_device }}"
```
dtantsur, line 99: This is a string that always evaluates to True in Python. You need to remove the quotes.

PR #607, opened 2021-03-10
Reduce loop cycle time for upgrade script.

smoshiur1237:
Currently we are using larger loop cycle time in the upgrade common file, which are not necessary. This PR will reduce the loop cycle time in the script.

### DIFF
--- a/scripts/feature_tests/upgrade/upgrade_common.sh
+++ b/scripts/feature_tests/upgrade/upgrade_common.sh
@@ -88,7 +88,7 @@ function deprovision_cluster() {
 
 function wait_for_cluster_deprovisioned() {
     echo "Waiting for cluster to be deprovisioned"
-    for i in {1..3600}; do
+    for i in {1..600}; do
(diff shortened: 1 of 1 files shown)


smoshiur1237:
/cc @Xenwar 
/assign @kashifest 
/cc @fmuyassarov 
//test-upgrade-features

scripts/feature_tests/upgrade/upgrade_common.sh
```diff
@@ -88,7 +88,7 @@ function deprovision_cluster() {
 
 function wait_for_cluster_deprovisioned() {
     echo "Waiting for cluster to be deprovisioned"
-    for i in {1..3600}; do
```
kashifest, line 91: this has been changed to 600 but later 1800, I am not understanding the logic/reason

smoshiur1237: This is set to 600 because, cluster deprovisioning doesn't need longer time cycle.. For example controlplane provisioning needs longer time than this and that is the reason I use 1800 for 3 controlplane provisioning.

Xenwar (approved):

PR #608, opened 2021-03-16
Using a better lower-bound version for jinja2 to make sure behavior is right

osherdp:
Jinja2 namespaces were introduced in 2.10, hence not available on older versions.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -27,7 +27,7 @@ else
   sudo alternatives --set python /usr/bin/python3
 fi
 
-sudo pip3 install ansible
+sudo pip3 install ansible "jinja2>=2.10"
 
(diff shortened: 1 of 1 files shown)


hardys:
Thanks @osherdp!

For other reviewers, we use namespaces in the libvirt role templating ref https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/libvirt/templates/network.xml.j2#L37

This was introduced by me in https://github.com/metal3-io/metal3-dev-env/pull/129 but I neglected to set the lower bound for the jinja2 version.

/ok-to-test
/lgtm

furkatgofurov7:
There was an issue with our CI recently, so tests did not run properly in the previous runs. Hopefully should be fine now.

/test-centos-integration
/test-integration

osherdp:
Can anyone approve?

PR #611, opened 2021-03-17
Support NUM_NODES=0 metal3 deployments

macaptain:
This change allows NUM_NODES=0 to deploy Metal3 with no virtual machines acting as bare metal hosts. 

Prior to this change, NUM_NODES=0 would error during the 02 script while doing vm-setup (vm_nodes undefined).

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -252,7 +252,7 @@ function make_bm_hosts() {
 function apply_bm_hosts() {
   pushd "${BMOPATH}"
   list_nodes | make_bm_hosts > "${WORKING_DIR}/bmhosts_crs.yaml"
-  if [ "${EPHEMERAL_CLUSTER}" != "tilt" ]; then
+  if [[ -n "$(list_nodes)" ]]; then
     kubectl apply -f "${WORKING_DIR}/bmhosts_crs.yaml" -n metal3
   fi
   popd
--- a/04_verify.sh
+++ b/04_verify.sh
@@ -281,11 +281,13 @@ process_status $?
 echo ""
 
 ## Verify
-while read -r name address user password mac; do
-  iterate check_bm_hosts "${name}" "${address}" "${user}" \
-    "${password}" "${mac}"
-  echo ""
-done <<< "$(list_nodes)"
+if [[ -n "$(list_nodes)" ]]; then
+  while read -r name address user password mac; do
+    iterate check_bm_hosts "${name}" "${address}" "${user}" \
+      "${password}" "${mac}"
+    echo ""
+  done <<< "$(list_nodes)"
+fi
 
 # Verify that the operator are running locally
 if [[ "${BMO_RUN_LOCAL}" == true ]]; then
--- a/vm-setup/roles/common/tasks/main.yml
+++ b/vm-setup/roles/common/tasks/main.yml
@@ -3,7 +3,12 @@
 - set_fact:
     generate_vm_nodes: "{{vm_nodes is not defined}}"
 
-- name: Define vm_nodes if not already defined
+- name: Set an empty default for vm_nodes if not already defined
+  set_fact:
+    vm_nodes: []
+  when: generate_vm_nodes
+
+- name: Populate vm_nodes if not already defined
   when: generate_vm_nodes
   include_tasks: vm_nodes_tasks.yml
   loop: "{{flavors|dict2items}}"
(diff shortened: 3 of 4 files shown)


03_launch_mgmt_cluster.sh
```diff
@@ -252,7 +252,7 @@ function make_bm_hosts() {
 function apply_bm_hosts() {
   pushd "${BMOPATH}"
   list_nodes | make_bm_hosts > "${WORKING_DIR}/bmhosts_crs.yaml"
-  if [ "${EPHEMERAL_CLUSTER}" != "tilt" ]; then
```
macaptain, line 255: I've removed this check because `apply_bm_hosts` is only called from a place where `"${EPHEMERAL_CLUSTER}" != "tilt"` is already true.

fmuyassarov:
> /ok-to-test 

@macaptain I thought you are member of Metal3 github org.

fmuyassarov:
Because you need ok-to-test if you aren't member of the Metal3 guthub org.

macaptain:
> > /ok-to-test
> 
> @macaptain I thought you are a member of Metal3 github org.

@fmuyassarov I'm a member of the Metal3 organisation. This shows both on my profile and on my comments in this pull request. Is there any other permission I need to use the metal3-io-bot?

fmuyassarov:
> > > /ok-to-test
> > 
> > 
> > @macaptain I thought you are a member of Metal3 github org.
> 
> @fmuyassarov I'm a member of the Metal3 organisation. This shows both on my profile and on my comments in this pull request. Is there any other permission I need to use the metal3-io-bot?

Ok, I think you need to make you membership public here https://github.com/orgs/metal3-io/people

macaptain:
> > > > /ok-to-test
> > > 
> > > 
> > > @macaptain I thought you are a member of Metal3 github org.
> > 
> > 
> > @fmuyassarov I'm a member of the Metal3 organisation. This shows both on my profile and on my comments in this pull request. Is there any other permission I need to use the metal3-io-bot?
> 
> Ok, I think you need to make you membership public here https://github.com/orgs/metal3-io/people

Done! Thanks @fmuyassarov, let's see if that helps.

furkatgofurov7:
It might make sense to update [vars.md](https://github.com/metal3-io/metal3-dev-env/blob/cb568d3cfec1c1fa23f9ee847b9a05e290ce2144/vars.md#L42)  `Allowed values` column  for NUM_NODES with this support?

macaptain:
> It might make sense to update [vars.md](https://github.com/metal3-io/metal3-dev-env/blob/cb568d3cfec1c1fa23f9ee847b9a05e290ce2144/vars.md#L42) `Allowed values` column for NUM_NODES with this support?

Other numerical values have the `Allowed values` field blank, so to keep it consistent, I'm tempted to leave this field blank for NUM_NODES as well. I expected NUM_NODES=0 to work based on the description of the variable, but it didn't, so this PR could be considered more of a fix than a feature.

03_launch_mgmt_cluster.sh
```diff
@@ -252,7 +252,7 @@ function make_bm_hosts() {
...
+  if [[ -n "$(list_nodes)" ]]; then
```
kashifest, line 255: What if we `apply_bm_hosts` is called from some other place , the condition should not be changed IMO? Whats the motivation of changing this?

kashifest (commented):
Looks good. Whats the motivation of supporting NUM_NODES=0 ?

03_launch_mgmt_cluster.sh
macaptain: The condition `${EPHEMERAL_CLUSTER}" != "tilt"` is already checked at the only site where `apply_bm_hosts` is called in the codebase (https://github.com/metal3-io/metal3-dev-env/blob/f089833114e5a722f6ed69e42eea660353168f4b/03_launch_mgmt_cluster.sh#L471-L474), so this function will never be called for a tilt cluster. Removing the check slightly improves the intent of the function. When I looked at it, I wondered why bmhosts_crs.yaml would be generated but not applied for a tilt cluster. But it turns out in reality, it is neither generated nor applied for a tilt cluster.

macaptain:
> Looks good. Whats the motivation of supporting NUM_NODES=0 ?

I'd like to deploy a Metal3 environment with the management cluster in Minikube, but I don't need to spin up virtual machines to act as bare metal hosts (I want to apply my own Kubernetes manifests for different hosts). With this small change, I can set `NUM_NODES=0` in the config_<user>.sh file, run `make` and the environment comes up, without spinning up node-0 and node-1, which I don't need for the environment.

fmuyassarov (approved):

PR #614, opened 2021-03-22
Uplift k8s to 1.20.4

kashifest:
Since kind node image version is not a 1 to 1 mapping to k8s version we also introduce `KIND_NODE_IMAGE_VERSION` variable.


kashifest:
/hold 
until node images are built

fmuyassarov (commented):
Maybe https://github.com/kubernetes/kubernetes/releases/tag/v1.20.5 ?

kashifest:
> Maybe https://github.com/kubernetes/kubernetes/releases/tag/v1.20.5 ?

Nope, we always keep a gap to allow upgrade

fmuyassarov (approved):

lib/common.sh
```diff
@@ -199,6 +199,7 @@ export KUSTOMIZE_VERSION=${KUSTOMIZE_VERSION:-"v3.8.5"}
 
 # Kind version
 export KIND_VERSION=${KIND_VERSION:-"v0.10.0"}
+export KIND_NODE_IMAGE_VERSION=${KIND_NODE_IMAGE_VERSION:-"1.20.2"}
```
furkatgofurov7, line 202: I have two questions:

1. why kind image version is not 1 to 1 map to k8s version? 
2. why do we have different versions of kind image in this patch and in [airship-dev-tools repo](https://github.com/Nordix/airship-dev-tools/pull/271/files#diff-54b22a5b934099e29518bd94112e4f547eaddc3d8002c8510e59bd4188bc38b8R13)

kashifest: 1.  Probably kind folks are the best person to answer, why did they stop building kind images after 1.20.2 
2. We do not have different kind version in airship-dev-tools repo. Check the latest master.

furkatgofurov7: >     2\. We do not have different kind version in airship-dev-tools repo. Check the latest master.

Missed this [patch](https://github.com/Nordix/airship-dev-tools/pull/272) in between.

furkatgofurov7 (commented):
/lgtm

PR #617, opened 2021-03-24
Fix compatibility with newer ironic-client images

dtantsur:
See also: https://github.com/metal3-io/ironic-client/pull/4


fmuyassarov (approved):
`baremetal node list` would still work though. But I think some people still prefer to use `openstack baremetal node list`

PR #620, opened 2021-04-01
Parameterize CAPI maxSurge field

jan-est:
The Cluster API control plane now allows for scaling in [4293](https://github.com/kubernetes-sigs/cluster-api/pull/4293#event-4525955116). Documentation can be found [rolling-update-strategy](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20191017-kubeadm-based-control-plane.md#rolling-update-strategy)
This PR parameterizes the required metal3_dev_env fields and adds a new environment variable MAX_SURGE_VALUE.

### DIFF
--- a/config_example.sh
+++ b/config_example.sh
@@ -31,6 +31,14 @@
 #
 #export SSH_PUB_KEY=~/.ssh/id_rsa.pub
 
+# Set the controlplane replica count
+#export NUM_OF_MASTER_REPLICAS=1
+
+# This variable defines if controlplane should scale-in or scale-out during upgrade
+# The field values can be 0 or 1. Default is 1. When set to 1 controlplane scale-out
+# When set to 0 controlplane scale-in. In case of scale-in NUM_OF_MASTER_REPLICAS must be >=3. 
+# export MAX_SURGE_VALUE=1
+
 #
 # Select the Container Runtime, can be "podman" or "docker"
 # Defaults to "podman"
--- a/vars.md
+++ b/vars.md
@@ -9,6 +9,8 @@ assured that they are persisted.
 
 | Name | Option | Allowed values | Default |
 | :------ | :------- | :--------------- | :-------- |
+| NUM_OF_MASTER_REPLICAS | Set the controlplane replica count. ||1| 
+| MAX_SURGE_VALUE | This variable defines if controlplane should scale-in or scale-out during upgrade. | 0 (scale-in) or 1 (scale-out) |1| 
 | EPHEMERAL_CLUSTER | Tool for running management/ephemeral cluster. | minikube, kind, tilt | Ubuntu default is kind. Only minikube is supported on CentOS |
 | EXTERNAL_SUBNET                | This is the subnet used on the "baremetal" libvirt network, created as the primary network interface for the virtual bare metalhosts.                                                                                                                    | CIDR                               | 192.168.111.0/24                                             |
 | SSH_PUB_KEY                    | This SSH key will be automatically injected into the provisioned host by the clusterctl environment template files.                                                                                                                                                   |                           | ~/.ssh/id_rsa.pub                                            |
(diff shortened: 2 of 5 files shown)


fmuyassarov (commented):
LGTM, will run the tests first

fmuyassarov (approved):

furkatgofurov7 (commented):
LGTM

Small nit: you have to update  [config_example.sh](https://github.com/metal3-io/metal3-dev-env/blob/master/config_example.sh) and [vars.md ](https://github.com/metal3-io/metal3-dev-env/blob/master/vars.md) as well, but it can be tackled in a separate PR.

fmuyassarov:
> LGTM
> 
> Small nit: you have to update [config_example.sh](https://github.com/metal3-io/metal3-dev-env/blob/master/config_example.sh) and [vars.md ](https://github.com/metal3-io/metal3-dev-env/blob/master/vars.md) as well, but it can be tackled in a separate PR.

@jan-est  can we fix it now ?

maelk:
I'd rather tackle it in this PR, since we're doing the modification, let's do everything at once.

furkatgofurov7 (commented):
I forgot to mention one more thing, btw: if we want to set `MAX_SURGE_VALUE` to 0 (default is 1), `NUM_OF_MASTER_REPLICAS` MUST be <= 3 (default is 1) according to [xref](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20191017-kubeadm-based-control-plane.md#rolling-update-strategy). Should we note and/or take into account that case as well?

jan-est:
> I forgot to mention one more thing, btw: if we want to set `MAX_SURGE_VALUE` to 0 (default is 1), `NUM_OF_MASTER_REPLICAS` MUST be <= 3 (default is 1) according to [xref](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20191017-kubeadm-based-control-plane.md#rolling-update-strategy). Should we note and/or take into account that case as well?

Is that something we can take account in the individual test cases?

config_example.sh
```diff
@@ -31,6 +31,14 @@
 #
 #export SSH_PUB_KEY=~/.ssh/id_rsa.pub
 
+# Set the controlplane replica count
+#export NUM_OF_MASTER_REPLICAS=1
+
+# This variable defines if controlplane should scale-in or scale-out during upgrade
+# The field values can be 0 or 1. Default is 1. When set to 1 controlplane scale-out
+# When set to 0 controlplane scale-in. In case of scale-in NUM_OF_MASTER_REPLICAS must be <=3.
```
kashifest, line 39: Is this statement correct? `In case of scale-in NUM_OF_MASTER_REPLICAS must be <=3`

vars.md
```diff
@@ -9,6 +9,8 @@ assured that they are persisted.
 
 | Name | Option | Allowed values | Default |
 | :------ | :------- | :--------------- | :-------- |
+| NUM_OF_MASTER_REPLICAS | Set the controlplane replica count. ||1|
```
kashifest, line 12: Should it be ||1| ? There are double |

kashifest (commented):
lgtm, couple of small nits, by the way `CONTROL_PLANE_MACHINE_COUNT` is now replaced with `NUM_OF_MASTER_REPLICAS`. So do we need this `CONTROL_PLANE_MACHINE_COUNT` anywhere else?

config_example.sh
jan-est: good catch `NUM_OF_MASTER_REPLICAS` must be 3 or greater. Fixed

vars.md
jan-est: > Should it be ||1| ? There are double |

Third column is empty, so I assume this is correct?

jan-est:
> lgtm, couple of small nits, by the way `CONTROL_PLANE_MACHINE_COUNT` is now replaced with `NUM_OF_MASTER_REPLICAS`. So do we need this `CONTROL_PLANE_MACHINE_COUNT` anywhere else?

`CONTROL_PLANE_MACHINE_COUNT` is not used anywhere else. Originally this variable was not even initialized or defaulted anywhere in our code, which caused CAPI webhook to default replica count as 1.

vars.md
kashifest: Aha ok true

PR #621, opened 2021-04-01
Remove undeclared WORKER_MACHINE_COUNT variable

jan-est:
Metal3 dev-env uses undeclared WORKER_MACHINE_COUNT variable in cluster-template-workers.yaml This PR replaces WORKER_MACHINE_COUNT with defined and defaulted NUM_OF_WORKER_REPLICAS.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/templates/cluster-template-workers.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/cluster-template-workers.yaml
@@ -8,7 +8,7 @@ metadata:
     nodepool: nodepool-0
 spec:
   clusterName: ${ CLUSTER_NAME }
-  replicas: ${ WORKER_MACHINE_COUNT }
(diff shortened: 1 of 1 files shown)


fmuyassarov (commented):
LGTM
/test-integration 
/test-centos-integration

furkatgofurov7 (commented):
We could update controlplane template same way, since it is also using undeclared var. 
See [here](https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml#L8), we could replace it with [this](https://github.com/metal3-io/metal3-dev-env/blob/3d123df5249aeb4dd19f4415b629e7d36521835f/vm-setup/roles/v1aX_integration_test/vars/main.yml#L13).

fmuyassarov:
> mplate same way,

it is done in https://github.com/metal3-io/metal3-dev-env/pull/620

PR #622, opened 2021-04-01
Add tests for inspection API

fmuyassarov:
This is adding ansible tasks to check [inspection API](https://github.com/metal3-io/metal3-docs/blob/master/design/baremetal-operator/inspection-api.md) feature while running feature tests CI.


vm-setup/roles/v1aX_integration_test/tasks/inspection.yml
```diff
@@ -0,0 +1,50 @@
+---
+  - name: Wait until "{{ NUM_NODES }}" bmhs become ready.
```
fmuyassarov, line 2: This might look odd that we are using shell in Ansible :), but unfortunately field selector doesn't work for CRDs it seems, so the only way I can see it to rely on jq.
It would be nice to have something like this to filter out ready bmh, but that's not possible it seems right now. This not the limitation of Ansible, but rather Kubernetes. Similar discussion in [BMO](https://github.com/metal3-io/baremetal-operator/issues/779). 
```
- name: Search for all running pods
  community.kubernetes.k8s_info:
    kind: Pod
    field_selectors:
      - status.phase=Running
```

namnx228:
LGTM

scripts/feature_tests/setup_env.sh
```diff
@@ -11,3 +11,8 @@ if [[ "${1}" == "ug" ]]; then
 fi
 pushd "${M3PATH}" || exit
 make
+
+if [[ "${1}" != "ug" ]]; then
+  export ACTION="inspection"
+  "${M3PATH}"/scripts/run.sh
+fi
```
kashifest, line 18: Setup_env is supposed to setup the env only, Not sure if calling a test script is a good idea here ....

vm-setup/roles/v1aX_integration_test/tasks/inspection.yml
kashifest: This is done also in this PR. https://github.com/metal3-io/metal3-dev-env/pull/623. Depending on which one lands first, this would be redundant I suppose ?

vm-setup/roles/v1aX_integration_test/tasks/main.yml
```diff
@@ -82,4 +82,8 @@
 
 - name: Node remediation
   include: remediation.yml
-  when: v1aX_integration_test_action == "remediation"
\ No newline at end of file
+  when: v1aX_integration_test_action == "remediation"
+
+- name: test inspection API
+  include: inspection.yml
+  when: v1aX_integration_test_action in "inspection"
```
kashifest, line 89: I suppose since it is done before provisioning `inspection` wont require any other script like `verify` ?

kashifest (changes requested):

vm-setup/roles/v1aX_integration_test/tasks/inspection.yml
fmuyassarov: Right. I think this can go first. I was interested to test it in 623, and that's why I had to add it there too. But I will anyways need to rebase 623, once this will get merged.

scripts/feature_tests/setup_env.sh
fmuyassarov: Good point! Updated with by adding a new entry in the Makefile. Thanks for suggestions.

vm-setup/roles/v1aX_integration_test/tasks/main.yml
fmuyassarov: I think no, because this is happening after `make` is finished, and is suppose to put baremetalhosts in the same state (`Ready`). I added some checks to make sure that baremetalhosts are in Ready state before we move on with provisioning.

vm-setup/roles/v1aX_integration_test/tasks/inspection.yml
```diff
@@ -0,0 +1,58 @@
...
+          name: "{{  item  }}"
+          namespace: "{{ NAMESPACE }}"
+          annotations:
+            inspect.metal3.io: ""
+    loop: "{{ bmh_list | json_query('resources[*].metadata.name') }}"
+
+  - name: Wait until "{{ NUM_NODES }}" bmhs are in inspecting or inspect wait state.
+    shell: |
+        kubectl get bmh -n {{ NAMESPACE }} -o json | jq -r '[ .items[]
+        | select (.status.provisioning.state == "inspecting" or .status.provisioning.state == "inspect wait")
```
furkatgofurov7, line 43: Some related questions, after reading ref. proposal: 

- I was not able to find the mention in the proposal stating BMH might be in  `inspect wait` state while it is inspected, so why we should consider that state? 
- Is it possible to check once the inspection is finished, whether the previous inspection data is updated on the spec of the BareMetalHost object or that is out of scope of this test case?

furkatgofurov7 (commented):
Overall looks good, some questions inline:

vm-setup/roles/v1aX_integration_test/tasks/inspection.yml
```diff
@@ -0,0 +1,58 @@
...
+    register: ready_bmhs
+    retries: 150
+    delay: 10
+    until: ready_bmhs.stdout == NUM_NODES
+
+  - name: Get a list of all BareMetalHosts
+    community.kubernetes.k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: metal3
```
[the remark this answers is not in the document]
fmuyassarov: Ah true, I missed this.

vm-setup/roles/v1aX_integration_test/tasks/inspection.yml
fmuyassarov: > Some related questions, after reading ref. proposal:
> 
> * I was not able to find the mention in the proposal stating BMH might be in  `inspect wait` state while it is inspected, so why we should consider that state?

When node is being inspected, it will have either inspecting or inspect wait. This is to ensure that inspection was really triggered. 

> * Is it possible to check once the inspection is finished, whether the previous inspection data is updated on the spec of the BareMetalHost object or that is out of scope of this test case?

It would be nice to have that sort of check, but I'm not sure how to do it. Any suggestions?

furkatgofurov7: Not sure either if that is possible to check, so that's why was asking if current logic is already enough to say that BMH spec is already updated with new HW details.

PR #623, opened 2021-04-03
WIP: Refactor integration scripts

fmuyassarov:
Refactor integration scripts. Avoid using shell within ansible.


vm-setup/roles/v1aX_integration_test/tasks/move_back.yml
```diff
@@ -7,7 +7,7 @@
       kind: Deployment
       state: absent
       namespace: "{{ NAMESPACE }}"
-      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+      kubeconfig: "{{ TARGET_KUBECONFIG }}"
```
kashifest, line 10: If `move_back`  is called in local deployment separately after `move` was done separately, would the variable `{{ TARGET_KUBECONFIG }}` have the correct value which was registered during `move` step?

vm-setup/roles/v1aX_integration_test/tasks/post_pivot.yml
```diff
@@ -6,19 +6,19 @@
     shell: |
         kubectl taint nodes --all node-role.kubernetes.io/master-
     environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+      KUBECONFIG: "{{ TARGET_KUBECONFIG }}"
```
kashifest, line 9: Same concern as above

vm-setup/roles/v1aX_integration_test/vars/main.yml
```diff
@@ -18,6 +18,7 @@ BMOPATH: "{{ lookup('env', 'BMOPATH') }}"
 IRONIC_DATA_DIR: "{{ lookup('env', 'IRONIC_DATA_DIR') }}"
 KUBECONFIG_PATH: "/home/ubuntu/.kube/config"
 KUBERNETES_VERSION: "{{ lookup('env', 'KUBERNETES_VERSION') | default('v1.20.4', true) }}"
+TARGET_KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
```
kashifest, line 21: Ow, I guess this answers my previous two questions right? So this value is always fixed? no defaults?

fmuyassarov: Yep, that's right. It is hardcoded once here instead of several times within the tasks.

PR #627, opened 2021-04-07
Fix missing upgrade vars in feature tests

Insullone:
This fix the problem that upgrade vars are not used in feature tests.


kashifest (approved):
@fmuyassarov  @maelk  it needs approval

smoshiur1237 (approved):
/lgtm

PR #575, opened 2020-12-09
Fix centos minikube networking

kashifest:
Not restarting libvirt is messing up minikube network. This PR makes sure minikube network is synced with libvirtd.

PR #635, opened 2021-04-16
Remove libvirtd restart in init_minikube

macaptain:
This libvirtd restart shouldn't be needed after minikube network configuration. Removing the libvirtd restart will also avoid a timeout issue in minikube itself.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -403,9 +403,6 @@ function init_minikube() {
           --model virtio --source baremetal \
           --type network --config
     fi
-
-    # Restart libvirtd
-    sudo systemctl restart libvirtd
 }
 
 #


lib/common.sh
```diff
@@ -406,6 +406,9 @@ function init_minikube() {
 
     # Restart libvirtd
     sudo systemctl restart libvirtd
+
+    # Wait a moment for libvirtd to come up before continuing with other commands
+    sleep 2
```
kashifest, line 411: This seems a bit hackish, do we have a away to tell minikube to try again for wait longer for libvirt?

macaptain: The problem is that there's a timeout that's too short in minikube. I've raised an issue and submitted a patch: https://github.com/kubernetes/minikube/issues/11113. Until this is merged and a new version of minikube is released, we'll need to `sleep` here.

macaptain: The other alternative is to delete the `sudo systemctl restart libvirtd` line. But this PR suggests that it's necessary to fix a networking issue? https://github.com/metal3-io/metal3-dev-env/pull/575

fmuyassarov (approved):
This looks good to me until we have the actual fix in minikube landed.

lib/common.sh
kashifest: I don't recall the actual problem which intrigued the restart, you can try to remove it to test if it solve the issue @macaptain

macaptain: @kashifest I didn't run into any problems when I removed the libvirtd in my metal3-dev-env, so let's try removing it instead of sleeping.

PR #636, opened 2021-04-20
🐛 Fix the issue that tripleo-repo requires a stream version of OS to run properly. 

namnx228:
This temporary fix the issue `__main__.InvalidArguments: --stream provided, but OS is not the Stream version. Please ensure the host is Stream.` in Dockerfile of vbmc

### DIFF
--- a/resources/vbmc/Dockerfile
+++ b/resources/vbmc/Dockerfile
@@ -1,11 +1,11 @@
 FROM docker.io/centos:centos8
 
 RUN dnf install -y python3 python3-requests python3-pip && \
-    curl https://raw.githubusercontent.com/openstack/tripleo-repos/master/tripleo_repos/main.py | python3 - -b master current && \
-    dnf update -y && \
-    dnf install -y python3-virtualbmc && \
-    dnf clean all && \
-    rm -rf /var/cache/{yum,dnf}/*
+     curl https://raw.githubusercontent.com/openstack/tripleo-repos/0543f7664a6d2b2a24b21164a9256e75b886becb/tripleo_repos/main.py | python3 - -b master current && \
+     dnf update -y && \
+     dnf install -y python3-virtualbmc && \
+     dnf clean all && \
+     rm -rf /var/cache/{yum,dnf}/*
 
 # NOTE(dtantsur): work around old pyghmi in RDO; this operation will become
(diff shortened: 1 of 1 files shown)


fmuyassarov (commented):
Can't we still use a single RUN?
Can you please update the PR title to reflect what you are fixing here?

namnx228:
> Can't we still use a single RUN?
> Can you please update the PR title to reflect what you are fixing here?

Sure, I will change the title

kashifest:
Can't we use the `[--no-stream](https://github.com/openstack/tripleo-repos/blob/master/releasenotes/notes/stream_by_default-783cc6a2208b942c.yaml#L4-L5)` flag instead ? Or a centros-stream image? By the way, isn't is so that the `vbmc` image is built on master jobs only? Integration tests are still not broken with this right?

namnx228:
> Can't we use the [--no-stream](https://github.com/openstack/tripleo-repos/blob/master/releasenotes/notes/stream_by_default-783cc6a2208b942c.yaml#L4-L5) flag instead ? Or a centros-stream image? By the way, isn't is so that the `vbmc` image is built on master jobs only? Integration tests are still not broken with this right?

I have tried to to used the `--no-stream` flag, but it doesn't work. 
I am not sure about other PRs, but I think the integration tests for this PR should pass the CI since the vbmc is built based on the Dockerfile of this PR.

namnx228:
@kashifest Using centos stream can be a long term solution, but we need to have more discussion before adopting it. Therefore, let's use this temporary fix to make the CI master jobs pass asap.

fmuyassarov:
> Can't we use the [--no-stream](https://github.com/openstack/tripleo-repos/blob/master/releasenotes/notes/stream_by_default-783cc6a2208b942c.yaml#L4-L5) flag instead ? Or a centros-stream image? By the way, isn't is so that the `vbmc` image is built on master jobs only? Integration tests are still not broken with this right?

No, we enabled them for every PR on metal3-dev-env

fmuyassarov (approved):
Merging this now to unblock the Jenkins CI. This is happening after https://github.com/openstack/tripleo-repos/commit/e91bd735f7cb2124e9bf873614dcd47e24e1ea88 change. We need to probably pass the stream flag.
/cc @hardys

PR #638, opened 2021-04-22
🐛 A proper fix for the Centos stream issue that tripleo-repo requires a stream version of OS to run properly

namnx228:
This fix is a replacement for the temporary fix in https://github.com/metal3-io/metal3-dev-env/pull/636


fmuyassarov (approved):
Force merging it.

PR #639, opened 2021-04-23
Make pivot and repivot separate in feature test

kashifest:


furkatgofurov7 (commented):
/lgtm

In feature tests as it is now, we re-pivot after pivoting, which is not desirable as we need a way to test scale in & node reuse features right after pivoting. For that, this patch gets the re-pivot action out of pivoting, and so that re-pivoting will be called after node reuse tests.

PR #641, opened 2021-04-28
Pin Minikube version to v1.18.1

macaptain:
There's a flaky issue with getting an IP address in v1.19.0 which is still under investigation. By pinning the minikube version (like we pin the kind version), we'll make the master CentOS builds more reliable until the root cause of the issue can be found.


fmuyassarov (approved):

namnx228:
LGTM

fmuyassarov:
Merging it manually for now.

PR #642, opened 2021-04-28
Make ipv4 hostname the fqdn

hardys:
Otherwise we end up with inconsistent inspection results from IPA in
dual-stack environments, initially the hostname is the shortname
provided via the ipv4 lease, then later it's updated to the FQDN
provided via ipv6.

To ensure consistent behavior, append the domain so the ipv4 host_name
option contains the FQDN.

### DIFF
--- a/vm-setup/roles/libvirt/templates/network.xml.j2
+++ b/vm-setup/roles/libvirt/templates/network.xml.j2
@@ -41,7 +41,11 @@
         {% set ironic_name = ironic_prefix + flavor + "_" + num|string %}
         {% set hostname_format = lookup('vars', flavor + '_hostname_format', default=flavor + '-%d') %}
         {% set hostname = hostname_format % num %}
+  {% if item.domain is defined %}
+      <host mac='{{ node_mac_map.get(ironic_name).get(item.name)}}' name='{{hostname}}.{{item.domain}}' ip='{{item.dhcp_range_v4[0]|ipmath(ns.index|int)}}'/>
+  {% else %}
       <host mac='{{ node_mac_map.get(ironic_name).get(item.name)}}' name='{{hostname}}' ip='{{item.dhcp_range_v4[0]|ipmath(ns.index|int)}}'/>
+  {% endif %}
         {% set ns.index = ns.index + 1 %}
       {% endfor %}
     {% endfor %}


hardys:
Some more data - I was initially seeing results like this in a dual-stack environment:

```
$ for bmh in master-0 master-1 master-2 worker-0 worker-1; do yq .status.hardware.hostname ostest-$bmh.yaml; done
"master-0"
"master-1.ostest.test.metalkube.org"
"master-2.ostest.test.metalkube.org"
"worker-0.ostest.test.metalkube.org"
"worker-1"
```

Looking at the DHCP data, we can see ipv4 provides the shortname, ipv6 provides the FQDN (despite both being the shortname in the libvirt generated dnsmasq config):

```
# On the target system/IPA
Apr 28 11:00:22 worker-0 NetworkManager[1704]: <info>  [1619607622.7169] dhcp4 (enp2s0): option host_name            => 'worker-0'
Apr 28 11:00:24 worker-0 NetworkManager[1704]: <info>  [1619607624.6423] dhcp6 (enp2s0): option fqdn_fqdn            => 'worker-0.ostest.test.metalkube.org'

# On the virthost
cat ostestbm.hostsfile  | grep worker-0
00:38:f6:47:fd:1a,192.168.111.23,worker-0
id:00:03:00:01:00:38:f6:47:fd:1a,worker-0,[fd2e:6f44:5dd8:c956::17]
```

In the IPA logs, we see the hostname changes, which means the IPA results are inconsistent/racy:

```
NetworkManager[461]: <info>  [1619546674.7467] policy: set-hostname: set hostname to 'worker-1' (from DHCPv4)
2021-04-27T18:06:50.520352181Z 920035dc-8687-4b1b-a716-494c73de8fa8_20210427-180650.207144.tar.gz: Apr 27 14:06:11 worker-1.ostest.test.metalkube.org NetworkManager[461]: <info>  [1619546771.7028] policy: set-hostname: current hostname was changed outside NetworkManager: 'worker-1.ostest.test.metalkube.org'
```

I think the "changed outside NetworkManager" is because of the IPA `/etc/NetworkManager/dispatcher.d/01-hostname` (workaround for NM ignoring ipv6 option fqdn_fqdn IIRC).

Making this consistent by forcing the ipv4 host_name to the fqdn gives consistent results AFAICT:

```
$ kubectl get bmh -n openshift-machine-api -o yaml | grep hostname
      hostname: master-0.ostest.test.metalkube.org
      hostname: master-1.ostest.test.metalkube.org
      hostname: master-2.ostest.test.metalkube.org
      hostname: worker-0.ostest.test.metalkube.org
      hostname: worker-1.ostest.test.metalkube.org
```

It does make the dnsmasq config look a bit weird, but I couldn't find another way to avoid this racy behavior:

```
$ cat /var/lib/libvirt/dnsmasq/ostestbm.hostsfile | grep worker-0
00:2a:e1:01:cc:be,192.168.111.23,worker-0.ostest.test.metalkube.org
id:00:03:00:01:00:2a:e1:01:cc:be,worker-0,[fd2e:6f44:5dd8:c956::17]
```

hardys:
Also /cc @dtantsur @derekhiggins 

I'm not sure where `/etc/NetworkManager/dispatcher.d/01-hostname`  can be found - IIRC the NM bug related to ignoring the ipv6 DHCP name is resolved (@cybertron can you confirm?) but even if we remove the dispatcher I think we've got a race here as I'm not sure which lease NM set-hostname will prefer?

stbenjam (approved):
This looks reasonable to me.

cybertron:
> I'm not sure where `/etc/NetworkManager/dispatcher.d/01-hostname` can be found - IIRC the NM bug related to ignoring the ipv6 DHCP name is resolved (@cybertron can you confirm?) but even if we remove the dispatcher I think we've got a race here as I'm not sure which lease NM set-hostname will prefer?

Yes, that was fixed. Hostnames provided via DHCP should be respected now. I'm not sure whether that would have affected metal3-dev-env anyway though. NM would have been able to lookup the hostname from reverse DNS on libvirt instances. I suppose it's possible that would have affected the priority of the ipv4/ipv6 hostnames though.

PR #645, opened 2021-05-02
Stop using Travis for mandatory CI jobs

fmuyassarov:
As our Prow cluster is back on track, we should stop using Travis for running markdown & shellcheck in this repo.


kashifest:
Isn't it possible to keep this file but just disable travis CI in repo configs ?

fmuyassarov:
> Isn't it possible to keep this file but just disable travis CI in repo configs ?

Probably disabling it from Github status check would suffice in this case, or will it be still running behind?

kashifest:
This is not needed anymore since we have disabled travis from the settings

fmuyassarov:
We have disabled Travis instead of deleting the configuration file.
/close

PR #648, opened 2021-05-04
Add support for external VLAN in integration tests

macaptain:
This change allows setting the `EXTERNAL_VLAN_ID` environment variable in metal3-dev-env. When this is set, Metal3DataTemplates on the target cluster configure the node interfaces on the bare metal network to use this VLAN. When this variable is not set, all existing behaviour is preserved.

Currently, the VM nodes have two interfaces, one used for the provisioning network, the other used for the bare metal network. This change allows for a setup where target hosts have a single interface with link and a dedicated VLAN for the bare metal network.

As part of the implementation, the templates that are used for creating Metal3DataTemplates for CentOS/Ubuntu are consolidated into a single template. Because interface names vary between the distributions, I've implemented automatic interface name templating, using an Ansible filter plugin and the spec from BMH.


vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
```diff
@@ -108,8 +109,7 @@ export CTLPLANE_KUBEADM_EXTRA_CONFIG="
           }
           vrrp_instance VI_1 {
               state MASTER
-              interface eth1
-              virtual_router_id 1
```
kashifest, line 112: why are you removing virtual_router_id ?

kashifest (approved):
overall lgtm, I would like  @maelk to have a look also.

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
macaptain: Good catch. This is just a mistake. I've put this line back.

config_example.sh
```diff
@@ -21,8 +21,12 @@
 # V4 default of 192.168.111.0/24 set in lib/network.sh
 # V6 default of fd55::/64 is set in lib/network.sh
 #
+# If the network is on a specific VLAN, specific EXTERNAL_VLAN_ID (default: no
```
furkatgofurov7, line 24: nit: `specify`

vm-setup/roles/v1aX_integration_test/templates/metal3datatemplate-template.yaml
```diff
@@ -0,0 +1,69 @@
...
+      ethernets:
+{% for nic_name in bmh_nic_names %}
+        - type: phy
+          id: {{ nic_name }}
+          macAddress:
+            fromHostInterface: {{ nic_name }}
+{% endfor %}
+{% if EXTERNAL_VLAN_ID != "" %}
+      vlans:
+        - id: "vlan{{ EXTERNAL_VLAN_ID }}"
```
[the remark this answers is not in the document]
macaptain: I don't think so. What's that? See here for the API documentation on the `vlans:` section of the configuration: https://github.com/metal3-io/cluster-api-provider-metal3/blame/7a3b72b35bf1a6b0dedffaabe087e3a7ee2d6b38/docs/api.md#L780-L786

vm-setup/roles/v1aX_integration_test/templates/metal3datatemplate-template.yaml
```diff
@@ -0,0 +1,69 @@
...
+          mtu: 1500
+          macAddress:
+            fromHostInterface: {{ bmh_nic_names | first }}
+          vlanID: {{ EXTERNAL_VLAN_ID }}
+          vlanLink: {{ bmh_nic_names | first }}
```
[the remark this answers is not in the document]
macaptain: We have to pick a link (interface) on which to create a vlan interface, it will be something like `eno1` or `eth0` or `enp0s1`, depending on the system and what 'Predictable Network Naming' comes out with. The vlan interface gets the name `eth0.3` or similar. When the bare metal IP / keepalived IP is added to this vlan interface, traffic will be tagged with the vlan id.

vm-setup/roles/v1aX_integration_test/templates/metal3datatemplate-template.yaml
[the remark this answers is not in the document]
furkatgofurov7: I was thinking of if we need a `type` for vlans (i.e default, management, native) too as we do for [ethernets](https://github.com/metal3-io/metal3-dev-env/pull/648/files/136c7bfd7c260169ddf088e5f00894a6b5d59a33#diff-6f62231fe18aa7cad7c54c4fb0c1a7c01c24b9c9a7ba4c55ee04721b48285743R25), seems not relevant.

config_example.sh
macaptain: Fixed!

config_example.sh
```diff
@@ -21,8 +21,12 @@
 # V4 default of 192.168.111.0/24 set in lib/network.sh
 # V6 default of fd55::/64 is set in lib/network.sh
 #
+# If the network is on a specific VLAN, specify EXTERNAL_VLAN_ID (default: no
+# special VLAN handling, variable is empty.)
+#
 #export EXTERNAL_SUBNET_V4="192.168.111.0/24"
 #export EXTERNAL_SUBNET_V6="fd55::/64"
+#export EXTERNAL_VLAN_ID=""
```
namnx228, line 29: Please also add this variable to vars.md

namnx228 (commented):
LGTM, just one nit

config_example.sh
macaptain: Thanks for the spot, I've added a line to vars.md.

vm-setup/roles/v1aX_integration_test/filter_plugins/bmh_metadata.py
```diff
@@ -0,0 +1,16 @@
+"""Filter plugins for managing k8s BMH metadata"""
+
+def bmh_nic_names(bmh):
+    """Returns the names of the interfaces of a BareMetalHost as a sorted
+    list.""" 
+    nics = bmh["status"]["hardware"]["nics"]
```
maelk, line 6: would this panic if one of the BMHs did not go through introspection properly and does not have "hardware" key?

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
```diff
@@ -108,7 +109,7 @@ export CTLPLANE_KUBEADM_EXTRA_CONFIG="
           }
           vrrp_instance VI_1 {
               state MASTER
-              interface eth1
+              interface {% if EXTERNAL_VLAN_ID == "" %}eth1{% else %}eth0.{{ EXTERNAL_VLAN_ID }}{% endif %}
```
maelk, line 112: I would expect something similar for Ubuntu, no ?

vm-setup/roles/v1aX_integration_test/filter_plugins/bmh_metadata.py
macaptain: Yes, it will. Ansible will error and exit. By the time we use this filter in the code, all BMHs are passed inspection and they are provisioned. 

I think this is fine. Ansible's mantra is "fail fast". I would rather we error if something is not right than add safety here.

vm-setup/roles/v1aX_integration_test/templates/clusterctl_env_template_centos.rc
macaptain: Good spot, thanks. I've added the keepalived config for the Ubuntu VLAN interface too.

PR #659, opened 2021-05-18
Allow the ironic container to restart when certificates are updated

namnx228:
This PR enables the ironic to restart when TLS certificates are updated.
Related PRs: 
https://github.com/metal3-io/ironic-image/pull/247 (Merged)
https://github.com/metal3-io/baremetal-operator/pull/854


lib/common.sh
```diff
@@ -48,6 +48,8 @@ export DISTRO="${ID}${VERSION_ID%.*}"
 export OS="${ID}"
 export OS_VERSION_ID=$VERSION_ID
 export SUPPORTED_DISTROS=(centos8 rhel8 ubuntu18 ubuntu20)
+# Enable ironic restart feature when the TLS certificate is updated
+export RESTART_CONTAINER_CERTIFICATE_UPDATED=${RESTART_CONTAINER_CERTIFICATE_UPDATED:-"true"}
```
kashifest, line 52: Should this be also synced with TLS setup ? What if TLS and basic auth is disabled ?

namnx228: Yes, that is true. Although when TLS is disabled, `RESTART_CONTAINER_CERTIFICATE_UPDATED=true` doesn't cause any harm, but we don't need to enable this feature in that case. Thanks for spotting out. I will fix it

namnx228: I will give the top priority to the user option. They may want to enable the feature without TLS for testing purpose. If the user doesn't set the variable manually, we can sync it with TLS setup.

kashifest: Sounds good.

kashifest:
Would this pass until BMO PR 854 is merged?

fmuyassarov:
> Would this pass integration test until BMO PR 854 is merged?

probably yes because this is only exporting vars.

fmuyassarov (approved):

PR #668, opened 2021-05-25
Fetch the manifests even when the verification fails

namnx228:
Currently, the manifests of objects can be fetched only when the CI passes the verification step in script 04_verify.sh. This PR aims to take the manifests even if the CI cannot pass that step.

### DIFF
--- a/04_verify.sh
+++ b/04_verify.sh
@@ -311,4 +311,5 @@ done
 
 
 echo -e "\nNumber of failures : $FAILS"
+. "$(dirname "$(readlink -f "${0}")")"/scripts/fetch_manifests.sh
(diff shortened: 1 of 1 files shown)


04_verify.sh
```diff
@@ -311,4 +311,5 @@ done
 
 
 echo -e "\nNumber of failures : $FAILS"
+. "$(dirname "$(readlink -f "${0}")")"/scripts/fetch_manifests.sh
```
kashifest, line 314: At this moment we haven't deployed any target cluster so essentially no metal3 objects, so what would it fetch?

fmuyassarov:
@namnx228 do we still need it ?

PR #669, opened 2021-05-26
Refactor some usage of shell into native kubernetes modules + small fixes

lentzi90:
The major part of this PR is converting use of the Ansible `shell` module into `k8s_info`. In addition to this it changes all `include`s to `include_tasks` to make the behavior more predictable and avoid problems later when it is deprecated. (See [the documentation](https://docs.ansible.com/ansible/2.9/modules/include_module.html#include-module).)

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -118,8 +118,6 @@ function install_clusterctl() {
   curl -L https://github.com/kubernetes-sigs/cluster-api/releases/download/"${CAPIRELEASE}"/clusterctl-linux-amd64 -o clusterctl
   chmod +x ./clusterctl
   sudo mv ./clusterctl /usr/local/bin/clusterctl
-  mkdir -p home/"${USER}"/.cluster-api
-  touch home/"${USER}"/.cluster-api/version.yaml
 }
 
 if ! [ -x "$(command -v clusterctl)" ]; then
--- a/vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
@@ -23,52 +23,74 @@
           (IMAGE_OS == "")
 
   - name: Check if cluster deprovisioning started.
-    shell: "kubectl get cluster -n {{ NAMESPACE }} -o json | jq -r '.items[] | .status.phase'"
+    k8s_info:
+      # TODO: This should be set based on CAPI_VERSION when v1alpha4 is working.
+      api_version: cluster.x-k8s.io/v1alpha3
+      kind: Cluster
+      namespace: "{{ NAMESPACE }}"
     register: deprovision_cluster
     retries: 2
     delay: 20
-    until: (deprovision_cluster.stdout == "deleting") or
-           (deprovision_cluster.stdout == "Deleting") or
-           (deprovision_cluster.stdout == "")
+    until: (deprovision_cluster is succeeded) and (
+             (deprovision_cluster.resources | length == 0) or
+             (deprovision_cluster.resources[0].status.phase | lower == "deleting"))
 
   - name: Wait until "{{ NUM_NODES }}" bmhs become ready again.
-    shell: |
-        kubectl get bmh -n {{ NAMESPACE }} -o json | jq -r '[ .items[]
-        | select (.status.provisioning.state == "ready")
-        | .metadata.name ] | length'
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
     register: deprovisioned_nodes
     retries: 150
     delay: 10
-    until: deprovisioned_nodes.stdout ==  NUM_NODES
+    vars:
+      query: "[? status.provisioning.state=='ready'].status.provisioning.state"
+    until: (deprovisioned_nodes is succeeded) and
+           (deprovisioned_nodes.resources | length > 0) and
+           (deprovisioned_nodes.resources | json_query(query) | length == (NUM_NODES | int))
 
   - name: Wait until no metal3machines are remaining
-    shell: |
-        kubectl get metal3machine -n {{ NAMESPACE }} -o json | jq -r '[ .items[] ] | length'
+    k8s_info:
+      api_version: infrastructure.cluster.x-k8s.io/{{ CAPM3_VERSION }}
+      kind: Metal3Machine
+      namespace: "{{ NAMESPACE }}"
     register: deprovisioned_m3m
     retries: 150
     delay: 3
-    until: deprovisioned_m3m.stdout ==  "0"
+    until: (deprovisioned_m3m is succeeded) and
+           (deprovisioned_m3m.resources | length == 0)
 
   - name: Wait until no machines are remaining
-    shell: |
-        kubectl get machine -n {{ NAMESPACE }} -o json | jq -r '[ .items[] ] | length'
+    k8s_info:
+      # TODO: This should be set based on CAPI_VERSION when v1alpha4 is working.
+      api_version: cluster.x-k8s.io/v1alpha3
+      kind: Machine
+      namespace: "{{ NAMESPACE }}"
     register: deprovisioned_machines
     retries: 150
     delay: 3
-    until: deprovisioned_machines.stdout ==  "0"
+    until: (deprovisioned_machines is succeeded) and
+           (deprovisioned_machines.resources | length == 0)
 
   - name: Wait until no metal3cluster is remaining
-    shell: |
-        kubectl get metal3cluster -n {{ NAMESPACE }} -o json | jq -r '[ .items[] ] | length'
+    k8s_info:
+      api_version: infrastructure.cluster.x-k8s.io/{{ CAPM3_VERSION }}
+      kind: Metal3Cluster
+      namespace: "{{ NAMESPACE }}"
     register: deprovisioned_metal3cluster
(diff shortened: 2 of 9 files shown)


vm-setup/roles/v1aX_integration_test/tasks/verify.yml
```diff
@@ -84,23 +97,27 @@
...
+        - status.phase=Running
     retries: 150
     delay: 20
     register: running_pods
-    failed_when: >
-      (running_pods.stderr != "") or
-      (running_pods.rc > 1) or
-      (running_pods.stdout != "")
-    until: running_pods.stdout == ""
+    until: running_pods.resources | length == 0
```
macaptain, line 110: `grep -v` is an inverted match, so in the shell we are listing the pods that are _not_ in running phase and waiting until the output is empty. However, I think the field selector with the until clause here is doing the opposite? If so, I expect we have to fix this.

These "check" task names could be clearer as well to indicate that we're waiting for a condition.

macaptain (changes requested):
I like the extra information we will get in the Ansible output. We need to check that the conditions are exactly the same as the old shell though.

vm-setup/roles/v1aX_integration_test/tasks/verify.yml
lentzi90: Oh gotcha! :facepalm:

macaptain (approved):
This looks good to me, and I like this change. In particular, the increased log output could be helpful for debugging problems in future. I don't have permissions to add approvals which will allow merging, so other members of the project will have to take a look.

01_prepare_host.sh
```diff
@@ -119,8 +119,6 @@ function install_clusterctl() {
   curl -L https://github.com/kubernetes-sigs/cluster-api/releases/download/"${CAPIRELEASE}"/clusterctl-linux-amd64 -o clusterctl
   chmod +x ./clusterctl
   sudo mv ./clusterctl /usr/local/bin/clusterctl
-  mkdir -p home/"${USER}"/.cluster-api
-  touch home/"${USER}"/.cluster-api/version.yaml
```
lentzi90, line 123: Note the missing `/` before `home`. This creates a folder on the relative path which doesn't seem to be used for anything. Later references to `.cluster-api` all use the absolute path `/home/${USER}/.cluster-api`.

kashifest: @furkatgofurov7 any idea why it was so and should we add a `/` before `home` instead of removing it?

vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
```diff
@@ -23,52 +23,70 @@
           (IMAGE_OS == "")
 
   - name: Check if cluster deprovisioning started.
-    shell: "kubectl get cluster -n {{ NAMESPACE }} -o json | jq -r '.items[] | .status.phase'"
+    k8s_info:
+      api_version: cluster.x-k8s.io/v1alpha3
```
kashifest, line 27: I think we also need to parameterize this since we would have two versions of these soon

vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
```diff
@@ -23,52 +23,70 @@
...
+    vars:
+      query: "[? status.provisioning.state=='ready'].status.provisioning.state"
+    until: (deprovisioned_nodes is succeeded) and
+           (deprovisioned_nodes.resources | json_query(query) | length == (NUM_NODES | int))
 
   - name: Wait until no metal3machines are remaining
-    shell: |
-        kubectl get metal3machine -n {{ NAMESPACE }} -o json | jq -r '[ .items[] ] | length'
+    k8s_info:
+      api_version: infrastructure.cluster.x-k8s.io/v1alpha4
```
kashifest, line 52: I think we also need to parameterize this since we would have two versions of these soon

vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
```diff
@@ -23,52 +23,70 @@
...
     delay: 3
-    until: deprovisioned_m3m.stdout ==  "0"
+    until: (deprovisioned_m3m is succeeded) and
+           (deprovisioned_m3m.resources | length == 0)
 
   - name: Wait until no machines are remaining
-    shell: |
-        kubectl get machine -n {{ NAMESPACE }} -o json | jq -r '[ .items[] ] | length'
+    k8s_info:
+      api_version: cluster.x-k8s.io/v1alpha3
```
kashifest, line 63: same

vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
```diff
@@ -23,52 +23,70 @@
...
     delay: 3
-    until: deprovisioned_machines.stdout ==  "0"
+    until: (deprovisioned_machines is succeeded) and
+           (deprovisioned_machines.resources | length == 0)
 
   - name: Wait until no metal3cluster is remaining
-    shell: |
-        kubectl get metal3cluster -n {{ NAMESPACE }} -o json | jq -r '[ .items[] ] | length'
+    k8s_info:
+      api_version: infrastructure.cluster.x-k8s.io/v1alpha4
```
kashifest, line 74: same

kashifest (commented):
Lgtm. Some comments inline

01_prepare_host.sh
furkatgofurov7: Please see the related [issue](https://github.com/metal3-io/metal3-dev-env/issues/470) and [fix](https://github.com/metal3-io/metal3-dev-env/pull/485) where these were added. IMO we need `.cluster-api ` part with missing `/` but `version.yaml` most probably might not be needed.

vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
lentzi90: Done

vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
lentzi90: Done

vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
lentzi90: Done

vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
lentzi90: Done

lentzi90:
Thanks for the comments @kashifest !
I used the `CAPI_VERSION` and `CAPM3_VERSION` that were already defined to parametrize the API versions. Do you think we should also parametrize the BMO version? I didn't find a variable for this.

01_prepare_host.sh
lentzi90: That's weird. I haven't had any problems without this `.cluster-api` folder. :thinking:  Could it have been some strange value set for `$HOME`? I mean `clusterctl` should use `$HOME/.cluster-api`, right? Unless we set `--config` to something different.

lentzi90: Could it be actually this issue https://github.com/kubernetes-sigs/cluster-api/issues/3573 that was the root cause for #470? The timing is suspicious and the error seems to be very much the same.

kashifest (approved):

kashifest:
/hold
Squash the commits please

lentzi90:
Looks like I need to check also for empty list, not just unreachable API.
```
TASK [v1aX_integration_test : Wait until cluster becomes provisioned.] *********
task path: /home/****/tested_repo/vm-setup/roles/v1aX_integration_test/tasks/verify.yml:6
    "msg": "The conditional check '(provisioned_cluster is succeeded) and (provisioned_cluster.resources[0].status.phase | lower == \"provisioned\")' failed. The error was: error while evaluating conditional ((provisioned_cluster is succeeded) and (provisioned_cluster.resources[0].status.phase | lower == \"provisioned\")): list object has no element 0"
}
```

Will try to address this. It seems very rare though since I haven't seen it in any of the previous runs.

01_prepare_host.sh
furkatgofurov7: Strange, how come then we were facing that reported issue? Most probably CAPI had issue and we fixed it before them and now the fix seems to be redundant.

lentzi90:
I'm confused about the version changes. Maybe someone can explain?
After #685 was merged I get an error in the integration tests:
```
Failed to find API for resource with apiVersion \"cluster.x-k8s.io/v1alpha4\" and kind \"Cluster\"
```
So CAPI_VERSION must be set to v1alpha4 that is why it is trying to check for it. But when I check the CRD there is only v1alpha2 and v1alpha3 there.

I noticed also that this is hard-coded in the templates, e.g. here: https://github.com/metal3-io/metal3-dev-env/blob/8b61eaee442c7468d07f0209ed2e43f54ef16017/vm-setup/roles/v1aX_integration_test/templates/cluster-template-cluster.yaml#L1
I'm not sure really how these things should work but I think `CAPI_VERSION` should be set to `v1alpha3` in the tests until the CRDs support `v1alpha4`.

lentzi90:
Based on offline discussion we are going with hard coding the CAPI version until v1alpha4 support has landed. Then we should do a followup and parameterize it.
/test-integration
/test-v1a4-centos-integration

PR #672, opened 2021-05-27
Replace ansible shells by k8s_info module in the integration tests move tasks

mboukhalfa:
This PR replaces Ansible shells by `k8s_info` module in `move` tasks from the integration tests and adds Ansible `filter_plugins` to help filtering the `k8s_info resources` based on their status.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
+++ b/vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
@@ -1,0 +1,77 @@
+""" filtering k8s_info resources """
+
+from ansible.utils.display import Display
+
+display = Display()
+
+
+def msg(key, resources):
+    return "could not find %r key in %r" % (key, resources)
+
+
+def filter_phase(resources, phase):
+    """Filter resources based on a defined phase
+
+    Args:
+        resources : Json object contains a list of k8s resources.
+        phase (str): The status phase of the filtered resources
+                    e.g. 'running', 'provisioning', 'deleting'.
+
+    Returns:
+        list: A list of resources in the defined phase.
+    """
+
+    filtered = []
+    for r in resources:
+        try:
+            if r["status"]["phase"].lower() == phase:
+                filtered.append(r)
+        except KeyError:
+            display.warning(msg("['status']['phase']", resources))
+
+    return filtered
+
+
+def filter_ready(resources):
+    """return resources based on defined ready status"""
+    filtered = []
+    for r in resources:
+        try:
+            if r["status"]["ready"]:
+                filtered.append(r)
+        except KeyError:
+            display.warning(msg("['status']['ready']", resources))
+
+    return filtered
+
+
+def filter_provisioning(resources, state):
+    """Filter resources based on a defined provisioning state
+
+    Args:
+        resources : Json object contains a list of k8s resources.
+        state (str): The provisioning state of the filtered resources
+                    e.g. 'provisioned', 'ready', 'deprovisioning'.
+
+    Returns:
+        list: A list of resources in the defined provisioning state.
+    """
+    filtered = []
+    for r in resources:
+        try:
+            if r["status"]["provisioning"]["state"].lower() == state:
+                filtered.append(r)
+        except KeyError:
+            display.warning(msg("['status']['provisioning']['state']", resources))
+
+    return filtered
+
+
+class FilterModule:
+    def filters(self):
+        filters = {
+            "filter_phase": filter_phase,
+            "filter_ready": filter_ready,
+            "filter_provisioning": filter_provisioning,
+        }
+        return filters
--- a/vm-setup/roles/v1aX_integration_test/tasks/move.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/move.yml
@@ -30,7 +30,16 @@
     when: EPHEMERAL_CLUSTER == "minikube"
 
   - name: Obtain target cluster kubeconfig
-    shell: "kubectl get secrets {{ CLUSTER_NAME }}-kubeconfig -n {{ NAMESPACE }} -o json | jq -r '.data.value'| base64 -d > /tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    k8s_info:
+      kind: secrets
+      name: "{{ CLUSTER_NAME }}-kubeconfig"
+      namespace: "{{ NAMESPACE }}"
+    register: metal3_kubeconfig
+
+  - name: Decode and save cluster kubeconfig
+    copy:
+      content: "{{ metal3_kubeconfig.resources[0].data.value | b64decode }}"
+      dest: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
 
   - name: Create namespace
     k8s:
@@ -50,14 +59,16 @@
 
   # Check for cert-manager pods on the target cluster
   - name: Check if cert-manager  pods in running state
-    shell: "kubectl get pods -n cert-manager -o json | jq -r '.items[].status.phase' | grep -cv Running"
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    retries: 20 
-    delay: 20
+    k8s_info:
+      kind: pods
+      namespace: cert-manager
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+      field_selectors:
+        - status.phase!=Running
     register: target_running_pods
-    until: target_running_pods.stdout|int == 0
-    failed_when: target_running_pods.stdout|int > 0
+    retries: 20
+    delay: 20
+    until: target_running_pods.resources | length == 0
 
   # Install Ironic
   - name: Install Ironic
@@ -72,53 +83,47 @@
     
   # Check for pods & nodes on the target cluster
   - name: Check if pods in running state
-    shell: "kubectl get pods -A -o json | jq -r '.items[].status.phase' | grep -v Running"
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    k8s_info:
+      kind: pods
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+      field_selectors:
+        - status.phase!=Running
+    register: target_running_pods
     retries: 150
     delay: 20
-    register: target_running_pods
-    failed_when: >
-      (target_running_pods.stderr != "") or
-      (target_running_pods.rc > 1) or
-      (target_running_pods.stdout != "")
-    until: target_running_pods.stdout == ""
+    until: target_running_pods.resources | length == 0
 
   - name: Pivot objects to target cluster
     shell: "clusterctl move --to-kubeconfig /tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml -n {{ NAMESPACE }} -v 10"
 
   - name: Check if machines become running.
-    shell: |
-        kubectl get machines -n {{ NAMESPACE }} -o json | jq -r '[ .items[]
-        | select (.status.phase == "Running" or .status.phase == "running")
-        | .metadata.name ] | length'
+    k8s_info:
+      api_version: cluster.x-k8s.io/v1alpha3
+      kind: machines
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     register: provisioned_machines
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     retries: 50
     delay: 20
-    until: provisioned_machines.stdout == NUMBER_OF_BMH
+    until: provisioned_machines.resources | filter_phase("running") | length == (NUMBER_OF_BMH | int)
 
-  - name: Check if metal3machines become provisioned.
-    shell: |
-        kubectl get m3m -n {{ NAMESPACE }} -o json | jq -r '[ .items[]
-        | select (.status.ready == true)
-        | .metadata.name ] | length'
-    register: provisioned_m3m_machines
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+  - name: Check if metal3machines become ready.
+    k8s_info:
+      api_version: infrastructure.cluster.x-k8s.io/v1alpha4
+      kind: metal3machine
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: m3m_machines
     retries: 10
     delay: 20
-    until: provisioned_m3m_machines.stdout == NUMBER_OF_BMH
+    until: m3m_machines.resources | filter_ready | length == (NUMBER_OF_BMH | int)
 
   - name: Check if bmh is in provisioned state
-    shell: |
-        kubectl get bmh -n {{ NAMESPACE }} -o json | jq -r '[ .items[]
-        | select (.status.provisioning.state == "provisioned")
-        | .metadata.name ] | length'
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    register: provisioned_bmh
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: bmh
     retries: 10
     delay: 20
-    until: provisioned_bmh.stdout ==  NUMBER_OF_BMH
+    until: bmh.resources | filter_provisioning("provisioned") | length == (NUMBER_OF_BMH | int)


vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -95,53 +108,53 @@
...
+      kind: BareMetalHost 
+      namespace: "{{ NAMESPACE }}"
+    register: provisioned_bmh
     environment:
       KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    register: provisioned_bmh
     retries: 10
     delay: 20
-    until: provisioned_bmh.stdout ==  NUMBER_OF_BMH
+    until: provisioned_bmh | provisioned | length == NUMBER_OF_BMH
```
macaptain, line 159: The LHS is an `int` but the RHS is a `string`. You'll need to cast the RHS to an `int` to do the comparison.

macaptain (changes requested):

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -95,53 +107,49 @@
...
     delay: 20
-    register: target_running_pods
-    failed_when: >
-      (target_running_pods.stderr != "") or
-      (target_running_pods.rc > 1) or
-      (target_running_pods.stdout != "")
-    until: target_running_pods.stdout == ""
-
+    until: target_running_pods.resources | length == 0
+    failed_when: target_running_pods.resources | length > 0
```
kashifest, line 119: Does this one line of failed_when simulates the three lines removed above?

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -95,53 +107,49 @@
...
+      kind: machines 
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: provisioned_machines 
+    retries: 50 
     delay: 20
-    until: provisioned_machines.stdout == NUMBER_OF_BMH
-
+    until: provisioned_machines | filter_running | length == NUMBER_OF_BMH | int
+  
   - name: Check if metal3machines become provisioned.
```
kashifest, line 134: the check is for being ready, can you change the name also accordingly

kashifest (commented):
lgtm. Few small nits

vm-setup/roles/v1aX_integration_test/tasks/move.yml
mboukhalfa: IMHO the shell was executing `kubectl get pods -A -o json` then filters phase and finally `grep` lines that do not contain the `Running` word. As a result the `stdout` will contain lines of non running phases or `""` if all pods are in `Running` phase.

So, the deleted failed_when was checking if `target_running_pods.stdout != ""` and this is handled by `target_running_pods.resources | length == 0` the other `target_running_pods.stderr != "" `and `target_running_pods.rc > 1` As far as I can tell are shell cmd related and this should be handled by the module `k8s_info`

Docs:
- stderr: https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#stderr
- rc: https://docs.ansible.com/ansible/latest/reference_appendices/common_return_values.html#rc

vm-setup/roles/v1aX_integration_test/tasks/move.yml
mboukhalfa: Good catch

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
```diff
@@ -0,0 +1,55 @@
...
+    """return resources in running phase"""
+    running = []
+    for r in a["resources"]:
+        try:
+            if r["status"]["phase"].lower() == "running":
+                running.append(r)
+        except KeyError:
+            if skip_missing:
+                break
+            raise AnsibleFilterError(
```
macaptain, line 14: I expect you will need to import this error from `ansible.errors`.

smoshiur1237:
Checked and seems that the issue is resolved.
/lgtm

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
```diff
@@ -0,0 +1,63 @@
+""" filtering k8s_info resources """
+
+from ansible.errors import AnsibleFilterError
+from ansible.utils.display import Display
+
+display = Display()
+
+
+def filter_running(a, skip_missing=False):
```
macaptain, line 9: I think the default behaviour should be to warn, rather than error. If there's no `["status"]["phase"]` key, I think we can safely assume that whatever the resource is, it's not in the running state that we want, so let's warn + print the details about the resource, and continue, not appending the resource to the list.

Would it also be possible to add this behaviour to the other filters you've written? I believe all of them could encounter the same KeyError.

macaptain (changes requested):

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
```diff
@@ -0,0 +1,70 @@
...
+    filtered = []
+    for r in resources:
+        try:
+            if r["status"]["phase"].lower() == phase:
+                filtered.append(r)
+        except KeyError:
+            if skip_missing:
+                display.warning(msg("['status']['phase']", resources))
+                break
+            raise AnsibleFilterError(msg("['status']['phase']", resources))
```
macaptain, line 24: None of the code in this change runs with `skip_missing` False. Rather than write code for logic branches that aren't used at all (in the assumption that someone later will use your parameter), I'd rather we remove the `skip_missing` parameter entirely, and make the `skip_missing` behaviour the default and only behaviour, rather than giving users the option to have the filter plugin error.

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
```diff
@@ -0,0 +1,70 @@
...
+            if skip_missing:
+                display.warning(msg("['status']['ready']", resources))
+                break
+            raise AnsibleFilterError(msg("['status']['ready']", resources))
+
+    return filtered
+
+
+def filter_provisioning(resources, state, skip_missing=True):
+    """return resources in a defined provisioning state"""
```
macaptain, line 46: It's not clear to me what I would need to pass as `state` to this filter to use properly. Instead, could we call the filter `filter_provisioned` which only takes in resources as an argument and returns those which are provisioned? This would be very clear to me.

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
```diff
@@ -0,0 +1,70 @@
...
+from ansible.utils.display import Display
+
+display = Display()
+
+
+def msg(key, resources):
+    return "could not find %r key in %r" % (key, resources)
+
+
+def filter_phase(resources, phase, skip_missing=True):
```
macaptain, line 13: It isn't clear to me what all the different `phase` values I would use here are. Could we make this simpler? I'd rather all these filters take only one argument and then do what they say on the filter name, instead than doing logic based on strings from Jinja.

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -95,53 +106,47 @@
...
+    k8s_info:
+      api_version: infrastructure.cluster.x-k8s.io/v1alpha4
+      kind: metal3machine
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: m3m_machines
     retries: 10
     delay: 20
-    until: provisioned_m3m_machines.stdout == NUMBER_OF_BMH
+    until: m3m_machines.resources | filter_ready | length == NUMBER_OF_BMH | int
```
macaptain, line 141:
```suggestion
    until: m3m_machines.resources | filter_ready | length == (NUMBER_OF_BMH | int)
```

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -95,53 +106,47 @@
...
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: bmh
     retries: 10
     delay: 20
-    until: provisioned_bmh.stdout ==  NUMBER_OF_BMH
+    until: bmh.resources | filter_provisioning("provisioned") | length == NUMBER_OF_BMH | int
```
macaptain, line 152:
```suggestion
    until: bmh.resources | filter_provisioning("provisioned") | length == (NUMBER_OF_BMH | int)
```

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -95,53 +106,47 @@
...
+      api_version: cluster.x-k8s.io/v1alpha3
+      kind: machines
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     register: provisioned_machines
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     retries: 50
     delay: 20
-    until: provisioned_machines.stdout == NUMBER_OF_BMH
+    until: provisioned_machines.resources | filter_phase("running") | length == NUMBER_OF_BMH | int # skip_missing to fix when missing status key
```
macaptain, line 130:
```suggestion
    until: provisioned_machines.resources | filter_phase("running") | length == (NUMBER_OF_BMH | int)
```



I don't think this comment will mean a lot to people reading this Ansible code in the future in the future. Also, to make this clause clearer, you could wrap the NUMBER_OF_BMH | int in parentheses, so it's clear that the == is comparing LHS to RHS.

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
```diff
@@ -0,0 +1,70 @@
...
+        except KeyError:
+            if skip_missing:
+                display.warning(msg("['status']['phase']", resources))
+                break
+            raise AnsibleFilterError(msg("['status']['phase']", resources))
+
+    return filtered
+
+
+def filter_ready(resources, ready=True, skip_missing=True):
```
macaptain, line 29: Again, I don't think `ready` should be a parameter here. I don't think a filter called `filter_ready` should get a boolean argument that flips the logic backwards. (Often boolean arguments for a function are a code smell.) I'd rather have `filter_ready` do one thing and one thing only, and return me resources which are in ready status.

macaptain (changes requested):
I think the key to making this code nice and readable and usable in future is to make the filters as 'dumb' as possible. Let's not have them take arguments from Jinja, and instead the filter should do what it says on the name: `filter_ready`, `filter_provisioned`, `filter_running`.

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
mboukhalfa: Agree ! that was the idea in the beginning, however going through the different tasks different `phase` was needed such as `running`, `provisioning` and `deleting` creating a filter for each end up duplicating the same code which is a bit difficult to maintain moreover we will get confusing with other filter `phase_ready`, `provisioning_ready` in the provisioning filter. and when` ready = True / False`  So I still believe sticking to the json behavior and having small number of parameterized filters will be more efficient in the future to use, maintain, and easy to understand

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
mboukhalfa: Same as phase, we will end up having `filter_provisioning_provisioned `and `filter_provisioning_ready ` with duplicated code

Check details: https://github.com/metal3-io/metal3-dev-env/pull/672#discussion_r644925590

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
mboukhalfa: Next commit `Skip_missing` will be removed only keep `display.warning`

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
mboukhalfa: Agree with `filter_ready` `param: ready` is not needed

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
macaptain: In my opinion, duplication isn't bad when it makes the code easy to use. However, I can see that you have thought about this. It is not a blocker for me. I think it would be helpful in the docstring to give some example of usage and more details about the inputs that we expect to use.

kashifest (commented):
/approve
Please squash your commit. I ll put a hold , unhold it once squashing is done.
/hold

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
```diff
@@ -0,0 +1,104 @@
...
+    for r in resources:
+        try:
+            names.append(r["metadata"]["name"])
+        except KeyError:
+            display.warning(msg("['metadata']['name']", resources))
+
+    return names
+
+
+def filter_k8s_version(resources, version):
```
macaptain, line 82: It looks like this function isn't used. Did you intend to use it somewhere?

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -30,7 +30,16 @@
     when: EPHEMERAL_CLUSTER == "minikube"
 
   - name: Obtain target cluster kubeconfig
-    shell: "kubectl get secrets {{ CLUSTER_NAME }}-kubeconfig -n {{ NAMESPACE }} -o json | jq -r '.data.value'| base64 -d > /tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    k8s_info:
+      kind: secrets
+      name: "{{ CLUSTER_NAME }}-kubeconfig"
+      namespace: "{{ NAMESPACE }}"
+    register: metal3_kubconfig
```
macaptain, line 37: I'd prefer if this were named 'kubeconfig' rather than 'kubconfig', but it does seem to be consistently named in your Ansible, so I don't think there is a problem.

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
```diff
@@ -0,0 +1,104 @@
...
+        try:
+            if r["status"]["provisioning"]["state"].lower() == state:
+                filtered.append(r)
+        except KeyError:
+            display.warning(msg("['status']['provisioning']['state']", resources))
+
+    return filtered
+
+
+def get_names(resources):
```
macaptain, line 70: This filter doesn't seem to be used anywhere in your code. Did you have a use for it?

macaptain (commented):
The intent of the code looks to be the same before and after the refactor, which is good. I'd remove any filters you aren't using, it does not make sense to write functions that are not invoked. With this change, I'd approve, since the refactor itself looks fine and the extra information in the logs for these tests is great to have.

vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
mboukhalfa: Agree! Those filters will be included with the PR: https://github.com/metal3-io/metal3-dev-env/pull/679 and removed from here

macaptain (approved):

furkatgofurov7 (commented):
Great work! 

Maybe we have to rename the patch since it doesn't fully replace the shell from move file and is there a plan to fully replace it to that nice looking ansible modules in the future?

mboukhalfa:
> Great work!
> 
> Maybe we have to rename the patch since it doesn't fully replace the shell from move file and is there a plan to fully replace it to that nice looking ansible modules in the future?

I think it indicates that those changes are related to k8s_info. While the pending shells are not relevant to k8s_info and can be replaced in a separate PR:
```
  # Configure Ironic configmap
  - name: Configure Ironic Configmap
    shell: |
      cp {{ BMOPATH }}/ironic-deployment/keepalived/ironic_bmo_configmap.env {{ BMOPATH }}/ironic-deployment/keepalived/ironic_bmo_configmap.env.orig
      cp {{ IRONIC_DATA_DIR }}/ironic_bmo_configmap.env  {{ BMOPATH }}/ironic-deployment/keepalived/ironic_bmo_configmap.env

  - name: Initialize Provider component in target cluster
    shell: "clusterctl init --kubeconfig /tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml --core cluster-api:{{ CAPIRELEASE }} --bootstrap kubeadm:{{ CAPIRELEASE }} --control-plane kubeadm:{{ CAPIRELEASE }} --infrastructure metal3:{{ CAPM3RELEASE }} -v 5"
```
```
# Install Ironic
  - name: Install Ironic
    shell: "{{ BMOPATH }}/tools/deploy.sh false true {{ IRONIC_TLS_SETUP }} {{ IRONIC_BASIC_AUTH }} true"
    environment:
      IRONIC_HOST: "{{ IRONIC_HOST }}"
      IRONIC_HOST_IP: "{{ IRONIC_HOST_IP }}"
      KUBECTL_ARGS: "{{ KUBECTL_ARGS }}"

  - name: Reinstate Ironic Configmap
    shell: "mv {{ BMOPATH }}/ironic-deployment/keepalived/ironic_bmo_configmap.env.orig {{ BMOPATH }}/ironic-deployment/keepalived/ironic_bmo_configmap.env"
```
```
  - name: Pivot objects to target cluster
    shell: "clusterctl move --to-kubeconfig /tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml -n {{ NAMESPACE }} -v 10"
```

furkatgofurov7:
> While the pending shells are not relevant to k8s_info

Oh I see, I was not aware of that! 

> can be replaced in a separate PR:

Yeah, agree remaining parts can be replaced as a separate follow-up PR.

furkatgofurov7 (commented):
/lgtm

PR #676, opened 2021-05-31
🐛 Fix kubectl apply issue in nodereuse

furkatgofurov7:
Feature test CI sometimes failing with:

```
`TASK [v1aX_integration_test : Put maxSurge field in KubeadmControlPlane back to default value(1).] ***
...
[0;31m        "Resource: \"controlplane.cluster.x-k8s.io/v1alpha3, Resource=kubeadmcontrolplanes\", GroupVersionKind: \"controlplane.cluster.x-k8s.io/v1alpha3, Kind=KubeadmControlPlane\"",[0m
[0;31m        "Name: \"test1\", Namespace: \"metal3\"",[0m
[0;31m        "for: \"STDIN\": Operation cannot be fulfilled on kubeadmcontrolplanes.controlplane.cluster.x-k8s.io \"test1\": the object has been modified; please apply your changes to the latest version and try again"[0m`
```

It seems, since we already changed maxSurge field once before also and by that time `resoruceVersion` gets incremented and once we try to change the field again, it uses old cached `resoruceVersion` number to apply the object and fails. For that, we might need to drop some of the fields before reapplying them in KCP.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
@@ -28,7 +28,7 @@
 
   - name: Update maxSurge field in KubeadmControlPlane to 0 to start testing KCP scale-in feature. 
     shell: |
-        kubectl get kubeadmcontrolplane "{{ CLUSTER_NAME }}" -n "{{ NAMESPACE }}" -o json | jq '.spec.rolloutStrategy.rollingUpdate.maxSurge=0' | kubectl apply -f-
+        kubectl patch -n "{{ NAMESPACE }}" kubeadmcontrolplane "{{ CLUSTER_NAME }}" --type=merge -p '{"spec": {"rolloutStrategy": {"rollingUpdate": {"maxSurge": 0}}}}'
     environment:
       KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
 
@@ -161,7 +161,7 @@
 
   - name: Put maxSurge field in KubeadmControlPlane back to default value(1).
     shell: |
-        kubectl get kubeadmcontrolplane "{{ CLUSTER_NAME }}" -n "{{ NAMESPACE }}" -o json | jq '.spec.rolloutStrategy.rollingUpdate.maxSurge=1' | kubectl apply -f-
+        kubectl patch -n "{{ NAMESPACE }}" kubeadmcontrolplane "{{ CLUSTER_NAME }}" --type=merge -p '{"spec": {"rolloutStrategy": {"rollingUpdate": {"maxSurge": 1}}}}'
     environment:
       KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
 
@@ -224,7 +224,7 @@
 
   - name: Update maxSurge and maxUnavailable fields in MachineDeployment.
     shell: |
-        kubectl get machinedeployment "{{ CLUSTER_NAME }}" -n "{{ NAMESPACE }}" -o json | jq '.spec.strategy.rollingUpdate.maxSurge=0|.spec.strategy.rollingUpdate.maxUnavailable=1' | kubectl apply -f-
+        kubectl patch -n "{{ NAMESPACE }}" machinedeployment "{{ CLUSTER_NAME }}" --type=merge -p '{"spec": {"strategy": {"rollingUpdate": {"maxSurge": 0,"maxUnavailable": 1}}}}'
     environment:
       KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
 


vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -159,12 +159,25 @@
...
+    shell: |
+        kubectl get kubeadmcontrolplane "{{ CLUSTER_NAME }}" -n "{{ NAMESPACE }}" -o json | jq '.metadata.resourceVersion'
+    register: resource_version_kcp 
+    environment:
+      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+
   - name: Put maxSurge field in KubeadmControlPlane back to default value(1).
     shell: |
-        kubectl get kubeadmcontrolplane "{{ CLUSTER_NAME }}" -n "{{ NAMESPACE }}" -o json | jq '.spec.rolloutStrategy.rollingUpdate.maxSurge=1' | kubectl apply -f-
+        kubectl get kubeadmcontrolplane "{{ CLUSTER_NAME }}" -n "{{ NAMESPACE }}" -o json | jq '.spec.rolloutStrategy.rollingUpdate.maxSurge=1' | jq '.metadata|.resourceVersion=="{{ resource_version_kcp }}"' | kubectl apply -f-
```
Xenwar, line 171: Are you assigning a value ? or selecting or running conditionally ?

Xenwar (changes requested):

lentzi90 (changes requested):
I think it would be better to use `kubectl patch` or the `k8s` Ansible module for these things.

Something like this:

```yaml
- name: Put maxSurge field in KubeadmControlPlane back to default value(1).
    community.kubernetes.k8s:
      kind: KubeadmControlPlane
      name: "{{ CLUSTER_NAME }}"
      namespace: "{{ NAMESPACE }}"
      resource_definition:
        spec:
          rolloutStrategy:
            rollingUpdate:
              maxSurge: 1
```

Since this is a patch/merge and we only specify the relevant field we don't have to worry about the resource version.
What do you think?

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
furkatgofurov7: Yes, I am trying to assign a value of `resourceVersion` fetched just before applying it to object.

furkatgofurov7:
> I think it would be better to use `kubectl patch` or the `k8s` Ansible module for these things.
> 
> Something like this:
> 
> ```yaml
> - name: Put maxSurge field in KubeadmControlPlane back to default value(1).
>     community.kubernetes.k8s:
>       kind: KubeadmControlPlane
>       name: "{{ CLUSTER_NAME }}"
>       namespace: "{{ NAMESPACE }}"
>       resource_definition:
>         spec:
>           rolloutStrategy:
>             rollingUpdate:
>               maxSurge: 1
> ```
> 
> Since this is a patch/merge and we only specify the relevant field we don't have to worry about the resource version.
> What do you think?

Might be, the problem is not with using shell or ansible module here, because it works as is even without this change, but it is flaky. The actual problem is KCP deployment has an entry i.e for resourceVersion (because we are changing it second time at this point) which don't belong in a declarative configuration file and when we are trying to update again, it complains.

lentzi90:
> Might be, the problem is not with using shell or ansible module here, because it works as is even without this change, but it is flaky. The actual problem is KCP deployment has an entry i.e for resourceVersion (because we are changing it second time at this point) which don't belong in a declarative configuration file and when we are trying to update KCP again, it complains.

Yes I know. I'm saying this is a simpler way to achieve what you want. The only thing you want to change is `maxSurge`, right? The problem comes if you try to "change" also the `resourceVersion`, which happens because we grab the whole `KubeadmControlPlane` and then (with jq) changes `maxSurge`. This means that all fields will be applied, including `resourceVersion`, `uid` and what not. Since the resource was changed in between this is a problem because we try to set `resourceVersion` back to the old value.

To solve this you can try to remove all irrelevant fields, or set the correct value for `resourceVersion` like you do in this patch. I'm saying it is much simpler to do a patch that only touches the `maxSurge` field and ignores everything else. You can do this like I described above with `k8s` or you can use `kubectl patch` like this:

```console
$ kubectl patch -n {{ NAMESPACE }} kubeadmcontrolplane {{ CLUSTER_NAME }} -p '{"spec": {"rolloutStrategy": {"rollingUpdate": {"maxSurge": 1}}}}'
```

This will never try to change `resourceVersion` or `creationTimestamp` or anything like that. And you don't need to do `kubeactl get ...` first, or use `jq` or anything like that.

lentzi90:
I see it is failing with
```
Error from server (UnsupportedMediaType): the body of the request was in an unknown format - accepted media types include: application/json-patch+json, application/merge-patch+json, application/apply-patch+yaml
```
This is probably because `KubeadmControlPlane` is a custom resource. Sorry I should have though about that.
Anyway, I think you can solve it by adding `--type=merge`. The default is `strategic` and this won't work for custom resources and is unfortunately not well documented. Ref: https://github.com/kubernetes/kubernetes/issues/71024

It should work with the following:

```console
$ kubectl patch -n {{ NAMESPACE }} kubeadmcontrolplane {{ CLUSTER_NAME }} --type=merge -p '{"spec": {"rolloutStrategy": {"rollingUpdate": {"maxSurge": 1}}}}'

```

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -224,7 +224,7 @@
 
   - name: Update maxSurge and maxUnavailable fields in MachineDeployment.
     shell: |
-        kubectl get machinedeployment "{{ CLUSTER_NAME }}" -n "{{ NAMESPACE }}" -o json | jq '.spec.strategy.rollingUpdate.maxSurge=0|.spec.strategy.rollingUpdate.maxUnavailable=1' | kubectl apply -f-
+        kubectl patch -n "{{ NAMESPACE }}" machinedeployment "{{ CLUSTER_NAME }}" --type=merge -p '{"spec": {"strategy": {"rollingUpdate": [{"maxSurge": 0,"maxUnavailable": 1}]}}}'
```
lentzi90, line 227:
```suggestion
        kubectl patch -n "{{ NAMESPACE }}" machinedeployment "{{ CLUSTER_NAME }}" --type=merge -p '{"spec": {"strategy": {"rollingUpdate": {"maxSurge": 0,"maxUnavailable": 1}}}}'
```

lentzi90 (commented):
Now it worked for the others :tada:  but the last one had a typo. :slightly_smiling_face:

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
furkatgofurov7: It was suggested to use `[]` if patch is for more than a single value, but seems that is not the [case](https://github.com/kubernetes/kubernetes/issues/72663#issuecomment-452325790), we could simply pass them without `[]`? Thanks for the suggestion @lentzi90

lentzi90: Hmm I see. I think the suggestion is only for `--type=json` that has a different format. It is very confusing with these `--type` things because all of them takes a JSON-formatted string as input but they work very differently.

For the `merge` type that we are using here it will just use the string to override the resource in the cluster, so the format must be the same. This means that the `rollingUpdate` field should have the same type in the patch as it has in the cluster, and this is a map, not a list. I.e. `kubectl explain machinedeployment.spec.strategy.rollingUpdate` says `RESOURCE: rollingUpdate <Object>` not `<[]Object>`.

I'm not sure if I'm explaining it in a good way, but anyway the point is that it should be without `[]`.

P.S. For `--type=json` it would look something like the below (but I don't think we should use this).
```
kubectl patch ... --type=json '[{ "op": "replace", "path": "/spec/strategy/rollingUpdate/maxSurge", "value": 0 },
    { "op": "replace", "path": "/spec/strategy/rollingUpdate/maxUnavailable", "value": 1 }]'
```

furkatgofurov7:
cc @mboukhalfa

lentzi90 (approved):

kashifest (approved):

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -161,7 +161,7 @@
 
   - name: Put maxSurge field in KubeadmControlPlane back to default value(1).
```
Xenwar, line 162: There is no default or original value in the process, the process is stateless.
`Put maxSurge field in KubeadmControlPlane value to 1.`

lentzi90: What do you mean? According to `kubeactl explain kubeadmcontrolplane.spec.rolloutStrategy.rollingUpdate` it has default 1.

Xenwar: I agree. I am wrong.

PR #677, opened 2021-06-01
Fix minikube start

kashifest:
Minikube still fails to starts if the IP is not returned in two retries. There is no way to configure number of retries in minikube. So instead of failing early, we ignore the failure and try once more in 03.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -388,9 +388,14 @@ differs(){
 #
 function init_minikube() {
     #If the vm exists, it has already been initialized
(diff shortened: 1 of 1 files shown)


furkatgofurov7 (commented):
Hopefully this is the end of minikube failure story, thanks for the patch @kashifest. 

/lgtm

smoshiur1237:
Well done.
/lgtm

maelk (approved):

PR #678, opened 2021-06-01
Remove httpd-reverse-proxy from ironic container list

macaptain:
After https://github.com/metal3-io/baremetal-operator/pull/886 and https://github.com/metal3-io/ironic-image/pull/264, we don't have httpd-reverse-proxy as an expected ironic container anymore.


fmuyassarov (commented):
/approve

PR #679, opened 2021-06-02
Replace node_reuse shells by k8s and k8s_info modules

mboukhalfa:
This PR changes Ansible tasks in the `node_reuse.yml` file from shell to Ansible modules such as `k8s_info` and `k8s` and adds Ansible third-party module to patch json files.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
@@ -5,44 +5,58 @@
       NUMBER_OF_ALL_BMH: "{{ NUM_OF_WORKER_REPLICAS|int + NUM_OF_MASTER_REPLICAS|int }}"
 
   - name: Scale worker down to 0 to start testing KubeadmControlPlane node reuse test scenario.
-    shell: |
-        kubectl scale machinedeployment "{{ CLUSTER_NAME }}" -n "{{ NAMESPACE }}" --replicas=0
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    k8s:
+      api_version: cluster.x-k8s.io/{{ CAPI_VERSION }}
+      kind: MachineDeployment
+      name: "{{ CLUSTER_NAME }}"
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+      definition:
+        spec:
+          replicas: 0
 
   - name: Wait until worker is scaled down and "{{ NUMBER_OF_BMH_MD }}" BMH is Ready.
-    shell: kubectl get bmh -n "{{ NAMESPACE }}" | grep -w ready | wc -l
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: saved_bmhs
     retries: 200
     delay: 20
-    register: ready_hosts
-    until: ready_hosts.stdout == NUMBER_OF_BMH_MD
+    until:
+      - saved_bmhs is succeeded
+      - saved_bmhs.resources | filter_provisioning("ready") | length == (NUMBER_OF_BMH_MD | int)
 
   - name: Get provisioned BMH names and UUIDs mapping before upgrade in KubeadmControlPlane node reuse test scenario.
-    shell: |
-        kubectl get bmh -A -o json | jq '.items[]| select (.status.provisioning.state == "provisioned")
-        | "metal3/"+.metadata.name+"="+"metal3://"+.metadata.uid' | cut -f2 -d\" | sort > /tmp/before_upgrade_mapping.txt
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-
-  - name: Update maxSurge field in KubeadmControlPlane to 0 to start testing KCP scale-in feature. 
-    shell: |
-        kubectl patch -n "{{ NAMESPACE }}" kubeadmcontrolplane "{{ CLUSTER_NAME }}" --type=merge -p '{"spec": {"rolloutStrategy": {"rollingUpdate": {"maxSurge": 0}}}}'
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    set_fact:
+      bmh_before_upgrade_kcp: "{{ bmh_before_upgrade_kcp | default([]) + ['metal3/' + item.metadata.name + '=metal3://' + item.metadata.uid] }}"
+    with_items: "{{ saved_bmhs.resources | filter_provisioning('provisioned') }}"
 
   - name: Update Metal3MachineTemplate nodeReuse field to 'True'.
-    shell: |
-        kubectl get m3mt "{{ CLUSTER_NAME }}"-controlplane -n "{{ NAMESPACE }}" -ojson | jq '.spec.nodeReuse=true' | kubectl apply -f-
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    k8s:
+      api_version: infrastructure.cluster.x-k8s.io/{{ CAPM3_VERSION }}
+      kind: Metal3MachineTemplate
+      name: "{{ CLUSTER_NAME }}-controlplane"
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+      definition:
+        spec:
+          nodeReuse: true
 
   - name: Upgrade KubeadmControlPlane k8s version from "{{ KUBERNETES_VERSION }}" to "{{ UPGRADED_K8S_VERSION }}" to test KCP scale-in feature.
-    shell: |
-        kubectl get kubeadmcontrolplane "{{ CLUSTER_NAME }}" -n "{{ NAMESPACE }}" -o json | jq '.spec.version="{{ UPGRADED_K8S_VERSION }}"' | kubectl apply -f-
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    k8s:
+      api_version: controlplane.cluster.x-k8s.io/{{ CAPI_VERSION }}
+      kind: KubeadmControlPlane
+      name: "{{ CLUSTER_NAME }}"
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+      definition:
+        spec:
+          version: "{{ UPGRADED_K8S_VERSION }}"
+          rolloutStrategy:
+            rollingUpdate:
+              maxSurge: 0
 
   - pause:
       minutes: 1
@@ -66,45 +80,47 @@
     failed_when: provisioning_machine.stdout != "0"
 
   - name: Wait until "{{ NUMBER_OF_BMH_MD }}" BMH is in deprovisioning state.
-    shell: |
-        kubectl get bmh -n "{{ NAMESPACE }}" -o json | jq -r '[ .items[]
-        | select (.status.provisioning.state == "deprovisioning")
-        | .metadata.name ] | length'
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    register: bmh_in_deprovisioning
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: saved_bmhs
     retries: 150
     delay: 10
-    until: bmh_in_deprovisioning.stdout == NUMBER_OF_BMH_MD
+    until:
+      - saved_bmhs is succeeded
+      - saved_bmhs.resources | filter_provisioning("deprovisioning") | length == (NUMBER_OF_BMH_MD | int)
 
-  - name: Get the name of the deprovisioning BMH.
-    shell: |
-        kubectl get bmh -n "{{ NAMESPACE }}" -o json | jq -r '[ .items[]
-        | select (.status.provisioning.state == "deprovisioning")
-        | .metadata.name ] | add'
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    register: deprovisioning_bmh_name
-
-  - name: Wait until above deprovisioning BMH is in Ready state again.
-    shell: |
-        kubectl get bmh -n "{{ NAMESPACE }}" | grep -w ready | awk '{print $1}'
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    register: ready_bmh
+  - name: Wait until above deprovisioning BMHs are in Ready state again.
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: bmhs
     retries: 200
     delay: 2
-    until: deprovisioning_bmh_name.stdout in ready_bmh.stdout_lines
+    until:
+      - bmhs is succeeded
+      - (saved_bmhs.resources | filter_provisioning("deprovisioning") | map(attribute='metadata.name')) |
+        intersect(bmhs.resources | filter_provisioning("ready") | map(attribute='metadata.name')) |
+        length ==  (NUMBER_OF_BMH_MD | int)
 
   - name: Check if just deprovisioned and became ready BMH is re-used for next provisioning.
-    shell: |
-        kubectl get bmh -n "{{ NAMESPACE }}" | grep -w provisioning | awk '{print $1}'
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    register: provisioning_bmh
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: bmhs
     retries: 150
     delay: 2
-    until: deprovisioning_bmh_name.stdout in provisioning_bmh.stdout_lines
+    until:
+      - bmhs is succeeded
+      - (saved_bmhs.resources | filter_provisioning("deprovisioning") | map(attribute='metadata.name')) |
+        intersect(bmhs.resources | filter_provisioning("provisioning") | map(attribute='metadata.name')) |
+        length == (NUMBER_OF_BMH_MD | int)
 
   - name: Wait until two machines become running and updated with new "{{ UPGRADED_K8S_VERSION }}" k8s version.
     shell: |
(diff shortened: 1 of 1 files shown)


furkatgofurov7:
@mboukhalfa **just a note**, we have to run a feature test in this PR, once it is ready.

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -5,121 +5,167 @@
...
+          value: "{{ UPGRADED_K8S_VERSION }}"
+
+  - name: Apply maxSurge field and upgrade k8s version on KubeadmControlPlane k8s
+    k8s:
+      state: present
+      src: /tmp/KubeadmControlPlane.json
+      apply: yes
+  # End tasks to update maxSurge field and upgrade k8s version in KubeadmControlPlane to 0 to start testing KCP scale-in feature
+
+  # Tasks to update Metal3MachineTemplate nodeReuse field to 'True'
```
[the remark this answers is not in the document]
mboukhalfa: I have switched the sequence to: 

1. First update Metal3MachineTemplate nodeReuse field to 'True'
2. Update maxSurge field
3. Upgrade KubeadmControlPlane k8s version

Since I prefer to do update maxSurge and Upgrade KCP in the same task to avoid repeating the same tasks.
This should be ok if it is not necessary to Update maxSurge before updating nodeReuse?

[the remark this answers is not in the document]
furkatgofurov7: Yes, it is fine, updating maxSurge before or after updating nodeReuse field is not important, but we must ensure nodeReuse field update before KCP upgrade.

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -335,22 +429,34 @@
...
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: provisioned_host
     retries: 200
     delay: 20
-    register: provisioned_host
-    until: provisioned_host.stdout == "4"
+    until: provisioned_host.resources | filter_provisioning("provisioned") | length == 4
```
lentzi90, line 450:
```suggestion
    until: provisioned_host.resources | filter_provisioning("provisioned") | length == (NUM_NODES | int)
```

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -335,22 +429,34 @@
...
+      api_version: cluster.x-k8s.io/v1alpha3
+      kind: machines
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     register: running_machines
     retries: 200
     delay: 20
-    until: running_machines.stdout|int == 4
\ No newline at end of file
+    until: running_machines.resources | filter_phase("running") | length == 4
```
lentzi90, line 461:
```suggestion
    until: running_machines.resources | filter_phase("running") | length == (NUM_NODES | int)
```

lentzi90 (commented):
Great work so far! :slightly_smiling_face: 
There is a lot going on so easy to break things. If it becomes too hard to figure out what works and what doesn't, it may be worth splitting this up into smaller PRs. For example one for the control plane and one for the workers.

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -9,117 +9,132 @@
...
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     register: ready_bmh
     retries: 200
     delay: 2
-    until: deprovisioning_bmh_name.stdout in ready_bmh.stdout_lines
+    until: (deprovisioning_bmh.resources | filter_provisioning("deprovisioning") | get_names) |
+           intersect(ready_bmh.resources | filter_provisioning("ready") | get_names) |
+           length ==  (NUMBER_OF_BMH_MD | int)
```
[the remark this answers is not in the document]
mboukhalfa: Here if I well followed we are trying to make sure that the set of BMHs `n(BMHs) = NUMBER_OF_BMH_MD ` that was in `deprovisioning` state is now in `ready` state knowing that I tried to intersect the two sets (deprovisioning ⋂ ready ) and ensure that the intersect equals the deprovisioning length (that's equivalent to `deprovisioning_names difference(ready_name) == 0 )`.

I will see if it can be simplified

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -128,48 +143,63 @@
...
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: provisioning_bmh
 
+  - name: Get provisioned BMH names and UUIDs mapping after upgrade in KubeadmControlPlane node reuse test scenario.
+    set_fact:
+      bmh_after_upgrade_kcp: "{{ bmh_after_upgrade_kcp | default([]) + [item.metadata.name + '=metal3://' + item.metadata.uid] }}"
+    with_items: "{{ provisioning_bmh.resources | filter_provisioning('provisioned') | json_query('[*]') }}"
```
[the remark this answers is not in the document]
mboukhalfa: Indeed it is not and will be removed

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -182,145 +212,209 @@
...
   - name: Wait until controlplane is scaled down and "{{ NUMBER_OF_BMH_KCP }}" BMHs' are Ready.
-    shell: kubectl get bmh -n "{{ NAMESPACE }}" | grep -w ready | wc -l
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: all_ready_hosts
```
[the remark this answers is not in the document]
mboukhalfa: I have had this in mind I am always filtering the vars after register and if I named the var by its content before filtering I will end up with the same name in all tasks which might cause conflicts so I am naming this vars by what we will get after filtering. I am not sure if it is a good idea to have same name in multiple tasks !

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -182,145 +212,209 @@
...
   - name: Get provisioned BMH names and UUIDs mapping after upgrade in MachineDeployment node reuse test scenario.
-    shell: |
-        kubectl get bmh -A -o json | jq '.items[]| select (.status.provisioning.state == "provisioned")
-        | "metal3/"+.metadata.name+"="+"metal3://"+.metadata.uid' | cut -f2 -d\" | sort > /tmp/after_upgrade_mapping_md.txt
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-
+    set_fact:
+      bmh_after_upgrade_md: "{{ bmh_after_upgrade_md | default([]) + [item.metadata.name + '=metal3://' + item.metadata.uid] }}"
+    with_items: "{{ provisioning_bmh.resources | filter_provisioning('provisioned') | json_query('[*]') }}"
```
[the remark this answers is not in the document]
mboukhalfa: Will be removed

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -182,145 +212,209 @@
...
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: provisioned_host
     retries: 200
     delay: 20
-    register: provisioned_host
-    until: provisioned_host.stdout == "2"
+    until: provisioned_host.resources | filter_provisioning("provisioned") | length == 2
```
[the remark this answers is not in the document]
mboukhalfa: Checking this I have found that:
`NUMBER_OF_BMH_MD: "{{ NUM_OF_WORKER_REPLICAS|int }}"`
While:
`./scripts/feature_tests/node_reuse/Makefile:2:export NUM_OF_WORKER_REPLICAS := 1`

Which is not equal to the actual value `2` ! I think the var in the task name means 1 more BMH to be provisioned while another one already provisioned. @furkatgofurov7 probably can confirm and suggest if possible using a var instead of static value in these two tasks

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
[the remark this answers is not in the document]
lentzi90: I think it is safe to have the same name in multiple tasks. Naming them according to what we get after filtering can be quite confusing when reading the query. But if you want different variable names between tasks, maybe it would be better to name them according to "where we are" or "what we are doing". For example, this one could be `bmh_scaling_down` or something like this.

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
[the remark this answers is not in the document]
lentzi90: Oh interesting! Maybe we could set it to `NUM_OF_WORKER_REPLICAS + NUMBER_OF_BMH_MD`?

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -3,161 +3,180 @@
...
+      apply: yes
+      definition:
+        spec:
+          rolloutStrategy:
+            rollingUpdate:
+              maxSurge: 0
+          version: "{{ UPGRADED_K8S_VERSION }}"
+  
   - pause:
       minutes: 1
```
furkatgofurov7, line 66: Last run has been hanging here forever and upgrade has proceeded well. Wondering why it should hang when you specify exact time? We can remove pause in the next run and see again.

furkatgofurov7 (changes requested):

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
lentzi90: I think the pause was OK but it got stuck on the next step.
```
Pausing for 60 seconds
(ctrl+C then 'C' = continue early, ctrl+C then 'A' = abort)

TASK [v1aX_integration_test : pause] *******************************************
ok: [localhost]
```

`ok: [localhost]` should mean that it is done with the task and moving on to the next. Not sure why it is hanging there though...

furkatgofurov7: Hmm, in that case it is hanging forever while
`- name: Check if only single machine is in Deleting state and no other new machine is in Provisioning state while upgrade triggered to test the KCP scale-in feature. `

I was inside VM while last test was running and it did exactly delete 1 machine and no provisioning machine was there, but it hang, so it means there is a problem with check either it is a one time issue.

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -3,161 +3,180 @@
...
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    register: bmh_in_deprovisioning
+    until: 
+      - provisioning_machine is succeeded
+      - provisioning_machine.resources | filter_phase("deleting") | length == (NUMBER_OF_BMH_MD | int)
+    failed_when: 
+      - provisioning_machine is succeeded
+      - provisioning_machine.resources | filter_phase("Provisioning") | length  != 0
+
```
furkatgofurov7, line 83: why we have `provisioning_machine is succeeded` in both the desired and failure conditions?

mboukhalfa: I was trying to avoid api calls fails when resources does not exist ! Do not think this would cause an issue

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
mboukhalfa: I think this is a one time issue since it does not used to hangs there and I have just tested locally it failed in  Wait until "1" BMH is in deprovisioning state 2. same as yesterday !

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
lentzi90: Hmm yes that looks weird... I think it should only be in `until`. Actually, now that I think about it, I'm not sure why we need `failed_when` at all? It was necessary when using `kubectl` since it can give an error code if there is some temporary problem with the connection, but now it should not be a problem since we use the module and have `provisioning_machine is succeeded` in `until`.

mboukhalfa: IMHO `failed_when` ensures that `no other new machine is in Provisioning state while upgrade triggered` as mentioned in the task name we can use it in the until condition by inverting the condition to `provisioning_machine.resources | filter_phase("Provisioning") | length  == 0`. I did not get how it can be related to `kubectl` only

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
mboukhalfa: I think here it is better to use `NUMBER_OF_ALL_BMH`, isn't it ?

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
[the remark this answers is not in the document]
mboukhalfa:
```
---
  - set_fact:
      NUMBER_OF_BMH_KCP: "{{ NUM_OF_MASTER_REPLICAS|int }}"
      NUMBER_OF_BMH_MD: "{{ NUM_OF_WORKER_REPLICAS|int }}"
      NUMBER_OF_ALL_BMH: "{{ NUM_OF_WORKER_REPLICAS|int + NUM_OF_MASTER_REPLICAS|int }}"
```
`NUMBER_OF_BMH_MD` is initialized with `NUM_OF_WORKER_REPLICAS` It may work for this case but if vars values chaged I am afraid it won't

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
mboukhalfa: In fact, this issue is frequent although it is not always I have just increased the delay from 2 to 4 and I am testing I am not sure if this could have an impact

mboukhalfa:
\unhold

mboukhalfa:
\test-v1a4-centos-integration

lentzi90 (commented):
I have reviewed a bit more thoroughly now. The changes looks alright and if the flaky issues are gone then I think we can merge this. However, I have some comments on things that I think could be improved. Let's see what others think.

- There is some extra white space in some places (e.g. lines 150 and 157 of node_reuse.yml). Please remove it.
- This adds more filter plugins and I'm not sure we really need them. I understand that it is convenient it have shortcuts for things like `get_names`, but I would still prefer to stick to built in filters as much as possible. Every custom filter we add will make the code harder for "outsiders" to understand since they cannot be expected to know how our filters work. Basically I would like to stick to a few "standard" tools that people can be expected to know inside and out unless there is a very good reason to add custom things.
- When we do add filter plugins I think we need to be careful with the naming. This PR adds `get_names` and `filter_k8s_version`. Both are very generic and it is unclear what resources they can be used on. I think `get_names` will work on almost everything since it is practically a rule to set the name as `metadata.name`. But the k8s version is far from obvious. Nodes have a k8s version but they cannot be used with this filter, as it only supports Machines I think.

Here is an example of how to get the names of all Machines with a specific k8s version using only built in filters: https://github.com/metal3-io/metal3-dev-env/blob/30385579f7260a532d44022287db1d28b05e7e42/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml#L424-L428

lentzi90 (approved):
Great work @mboukhalfa !

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -5,48 +5,74 @@
...
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    set_fact:
+      bmh_before_upgrade_kcp: "{{ bmh_before_upgrade_kcp | default([]) + [item.metadata.name + '=metal3://' + item.metadata.uid] }}"
+    with_items: "{{ saved_bmhs.resources | filter_provisioning('provisioned') }}"
+
+  - name: Update maxSurge field in KubeadmControlPlane to 0 to start testing KCP scale-in feature.
+    k8s:
+      api_version: controlplane.cluster.x-k8s.io/{{ CAPI_VERSION }}
+
```
fmuyassarov, line 39: redundant ?

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -5,48 +5,74 @@
...
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-
-  - name: Update maxSurge field in KubeadmControlPlane to 0 to start testing KCP scale-in feature. 
-    shell: |
-        kubectl patch -n "{{ NAMESPACE }}" kubeadmcontrolplane "{{ CLUSTER_NAME }}" --type=merge -p '{"spec": {"rolloutStrategy": {"rollingUpdate": {"maxSurge": 0}}}}'
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    set_fact:
+      bmh_before_upgrade_kcp: "{{ bmh_before_upgrade_kcp | default([]) + [item.metadata.name + '=metal3://' + item.metadata.uid] }}"
```
furkatgofurov7, line 33: is it missing `"metal3/"` at the beginning? the prev. version was:
`"metal3/"+.metadata.name+"="+"metal3://"+.metadata.uid'`

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -128,48 +156,61 @@
...
+    register: saved_bmhs
 
   - name: Get provisioned BMH names and UUIDs mapping after upgrade in KubeadmControlPlane node reuse test scenario.
-    shell: |
-        kubectl get bmh -A -o json | jq '.items[]| select (.status.provisioning.state == "provisioned")
-        | "metal3/"+.metadata.name+"="+"metal3://"+.metadata.uid' | cut -f2 -d\" | sort > /tmp/after_upgrade_mapping.txt
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    set_fact:
+      bmh_after_upgrade_kcp: "{{ bmh_after_upgrade_kcp | default([]) + [item.metadata.name + '=metal3://' + item.metadata.uid] }}"
```
furkatgofurov7, line 181: same question as above here in all other occurences of this task in MD scenario

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -182,175 +223,247 @@
...
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: bmhs
     retries: 200
     delay: 20
-    register: all_ready_hosts
-    until: all_ready_hosts.stdout == "3"
+    until:
+      - bmhs is succeeded
+      - bmhs.resources | filter_provisioning("ready") | length == (NUMBER_OF_BMH_KCP   | int)
```
furkatgofurov7, line 236:
```suggestion
      - bmhs.resources | filter_provisioning("ready") | length == (NUMBER_OF_BMH_KCP | int)
```

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -182,175 +223,247 @@
...
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: machines
     retries: 200
     delay: 20
-    until: running_machines.stdout|int == 4
\ No newline at end of file
+    until:
+      - machines is succeeded
+      - machines.resources | filter_phase("running") | length == (NUMBER_OF_ALL_BMH  | int)
```
furkatgofurov7, line 469:
```suggestion
      - machines.resources | filter_phase("running") | length == (NUMBER_OF_ALL_BMH | int)
```

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -182,175 +223,247 @@
...
+        spec:
+          nodeReuse: True
+
+  - name: List BMHs'.
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: bmhs
```
furkatgofurov7, line 318: you should list only **Ready** BMHs' here, not all of them, are not you? Because at this point you should have 2 of them and those need to be marked unhealthy

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -182,175 +223,247 @@
...
+    until:
+      - machines is succeeded
+      - machines.resources | filter_phase("running") | length == 2
+
+  - name: Get provisioned BMH before upgrade in MachineDeployment node reuse test scenario.
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: saved_bmhs
```
furkatgofurov7, line 280: where is it checking for `provisioned` BMHs'?

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -66,45 +92,47 @@
...
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: bmhs
     retries: 150
     delay: 2
-    until: deprovisioning_bmh_name.stdout in provisioning_bmh.stdout_lines
+    until:
+      - bmhs is succeeded
+      - (saved_bmhs.resources | filter_provisioning("deprovisioning") | map(attribute='metadata.name')) |
+        intersect(bmhs.resources | filter_provisioning("provisioning") | map(attribute='metadata.name')) |
+        length == (NUMBER_OF_BMH_MD | int)
```
furkatgofurov7, line 135: this should check if last deprovisioned but currently in `Ready` state BMH has been picked up or not, but you are filtering for `deprovisioning` BMH here? how are you making sure that just deprovisioned BMH in previous step has been reused?

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -128,48 +156,61 @@
...
+    until:
+      - machines is succeeded
+      - machines.resources | filter_phase("running") | selectattr('spec.version', 'match', UPGRADED_K8S_VERSION) | length == (NUMBER_OF_BMH_KCP | int)
+
+  - name: Get provisioned BMH after upgrade in KubeadmControlPlane node reuse test scenario.
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: saved_bmhs
```
furkatgofurov7, line 177: where is the BMH state check condition?

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -182,175 +223,247 @@
...
-    shell: |
-        kubectl get bmh -A -o json | jq '.items[]| select (.status.provisioning.state == "provisioned")
-        | "metal3/"+.metadata.name+"="+"metal3://"+.metadata.uid' | cut -f2 -d\" | sort > /tmp/before_upgrade_mapping_md.txt
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    set_fact:
+      bmh_before_upgrade_md: "{{ bmh_before_upgrade_md | default([]) + [item.metadata.name + '=metal3://' + item.metadata.uid] }}"
+    with_items: "{{ saved_bmhs.resources | filter_provisioning('provisioned') }}"
 
   - name: Update maxSurge and maxUnavailable fields in MachineDeployment.
```
furkatgofurov7, line 287: I think it is better to explicitly define the values of `maxSurge` and `maxUnavailable` here too as you are doing in other places?

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -182,175 +223,247 @@
...
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: bmhs
     retries: 200
     delay: 2
-    until: deprovisioning_bmh_name.stdout in provisioning_bmh.stdout_lines
+    until:
+      - bmhs is succeeded
+      - (saved_bmhs.resources | filter_provisioning("deprovisioning") | map(attribute='metadata.name')) |
+        intersect(bmhs.resources | filter_provisioning("provisioning") | map(attribute='metadata.name')) |
+        length == (NUMBER_OF_BMH_MD | int)
```
furkatgofurov7, line 400: same question as in the KCP scenario

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -182,175 +223,247 @@
...
+    until:
+      - machines is succeeded
+      - machines.resources | filter_phase("running") | selectattr('spec.version', 'match', UPGRADED_K8S_VERSION) | length == 2
+
+  - name: Get provisioned BMH after upgrade in MachineDeployment node reuse test scenario.
+    k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: saved_bmhs
```
furkatgofurov7, line 420: same, condition missing?

furkatgofurov7 (commented):
Overall this is going in the right direction and looks good, but there are some questions inline:

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
mboukhalfa: Could you please mention what is redundant ?

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
mboukhalfa: I am saving previous bmhs in `saved_bmhs` and current in `bmhs` then intersect deprovisioning from the **last** with provisioning from the **current** and ensure that all of the previous deprovisioning become in provisioning state. 
Is not that enough ?

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
mboukhalfa: I do not think there is a way to register only provisioned bmhs after filter so I am saving the whole bmhs set (including provisioned) and filter it in the next set fact would you like to change task name or use both tasks in block ?

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
mboukhalfa: Because I could not register bmhs after filter this being done in the next task so I am marking only Ready bmhs as unhealthy

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
fmuyassarov: Ah sorry forgot to mention the exact line. L39 was empty and I was wondering why is it so. But now it is seems you have removed it.

fmuyassarov (commented):
/approve

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
furkatgofurov7: Okay got it

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
furkatgofurov7: this is not answered, or rename the task name appropriately

furkatgofurov7 (changes requested):

PR #689, opened 2021-06-10
Set NUM_NODES for feature tests

lentzi90:
NUM_NODES defaults to 2, which doesn't work with the feature tests since they set `NUM_OF_MASTER_REPLICAS=3` and `NUM_OF_WORKER_REPLICAS=1`. In the CI jobs we override NUM_NODES to make things work.

Now NUM_NODES is set to 4 for these tests (without having to override it).

Why I think we need this: When I want to manually test something by running for example `make feature_tests_upgrade`, I must first remember to do `export NUM_NODES=4` or the test will fail since there are not enough nodes. Since are already setting values for the number of masters/workers we might as well do it for the total number of nodes also.

Delete duplicate variable declarations for CAPM3RELEASE and CAPIRELEASE.

### DIFF
--- a/scripts/feature_tests/node_reuse/Makefile
+++ b/scripts/feature_tests/node_reuse/Makefile
@@ -1,5 +1,6 @@
 export NUM_OF_MASTER_REPLICAS := 3
 export NUM_OF_WORKER_REPLICAS := 1
+export NUM_NODES := 4
 
 all: node_reuse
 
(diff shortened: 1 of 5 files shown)


lentzi90:
This won't work unfortunately. The VMs are created before the tests are run which means that specifying `NUM_NODES` in the tests Makefiles doesn't help. That file is read only after the VMs are created.

Issue #688, opened 2021-06-10
Parametrize CAPI version


lentzi90:
Currently the CAPI version is hard-coded in multiple places and omitted in other places even though we have a `CAPI_VERSION` variable. We should start using this variable instead of hard-coding or omitting the version.

Unfortunately this is blocked until v1alpha4 works properly, as `CAPI_VERSION` is already set to `v1alpha4` for the integration tests. This means that we will first need to add support for v1alpha4 (or set CI to v1alpha3) before we can start using `CAPI_VERSION`.

Ref:
- Hard coded version: https://github.com/metal3-io/metal3-dev-env/blob/8b61eaee442c7468d07f0209ed2e43f54ef16017/vm-setup/roles/v1aX_integration_test/templates/cluster-template-cluster.yaml#L1
- Omitted version: https://github.com/metal3-io/metal3-dev-env/blob/eb07765eada06663933e27eec46c9c263b73e9bb/vm-setup/roles/v1aX_integration_test/tasks/remediation.yml#L280

PR #690, opened 2021-06-10
Refactor upgrade for control plane components

lentzi90:
- Use Ansible modules k8s and k8s_info instead of shell for interaction
with the API.
- Use relevant Ansible modules instead of shell for file operations
(edit/copy)
- Use label selectors instead of regex search for relevant k8s resources
- Add Ansible filter plugin for removing autogenerated fields (e.g.
resourceVersion and uid)

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/filter_plugins/k8s_fields.py
+++ b/vm-setup/roles/v1aX_integration_test/filter_plugins/k8s_fields.py
@@ -1,0 +1,36 @@
+"""
+Filter out managed fields, uids, timestamps and similar from Kubernetes
+resources. These fields are often "in the way" when trying to find the relevant
+parts or when "restoring" resources to a new cluster.
+"""
+
+import copy
+
+
+def k8s_backup(resources):
+    """
+    Take a backup of k8s resources by stripping away problematic fields.
+    Fields that are automatically created and should normally not be touched
+    will be removed so that the resources can safely be restored from this
+    "backup".
+    Removed fields under .metadata: uid, managedFields, resourceVersion,
+    creationTimestamp
+    """
+    unwanted_metadata_fields = ["uid", "resourceVersion", "creationTimestamp",
+                                "managedFields"]
+
+    filtered = copy.deepcopy(resources)
+    for resource in filtered:
+        for field in unwanted_metadata_fields:
+            if "metadata" in resource and field in resource["metadata"]:
+                del resource["metadata"][field]
+
+    return filtered
+
+
+class FilterModule:
+
+    def filters(self):
+        return {
+            'k8s_backup': k8s_backup
+        }
--- a/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
@@ -4,72 +4,101 @@
       NUMBER_OF_BMH: "{{ NUM_OF_MASTER_REPLICAS|int +  NUM_OF_WORKER_REPLICAS|int }}"
 
   - name: Update maxSurge and maxUnavailable fields
-    shell: |
-           kubectl get machinedeployment -n {{NAMESPACE}} test1 -o json | 
-           jq '.spec.strategy.rollingUpdate.maxSurge=1|.spec.strategy.rollingUpdate.maxUnavailable=1' | 
-           kubectl apply -f-
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    kubernetes.core.k8s:
+      api_version: cluster.x-k8s.io/v1alpha3
+      kind: MachineDeployment
+      name: "{{ CLUSTER_NAME }}"
+      namespace: "{{ NAMESPACE }}"
+      resource_definition:
+        spec:
+          strategy:
+            rollingUpdate:
+              maxSurge: 1
+              maxUnavailable: 1
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
 
   - name: Untaint all nodes
-    shell: |
-        kubectl taint nodes --all node-role.kubernetes.io/master-
+    ansible.builtin.command: kubectl taint nodes --all node-role.kubernetes.io/master-
     environment:
       KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     ignore_errors: yes
 
   - name: Scale worker down to 0
-    shell: |
-        kubectl scale machinedeployment "{{ CLUSTER_NAME }}"  -n "{{ NAMESPACE }}" --replicas=0
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    kubernetes.core.k8s:
+      api_version: cluster.x-k8s.io/v1alpha3
+      kind: MachineDeployment
+      name: "{{ CLUSTER_NAME }}"
+      namespace: "{{ NAMESPACE }}"
+      resource_definition:
+        spec:
+          replicas: 0
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
 
   - name: Wait until worker is scaled down and one bmh is Ready
-    shell: kubectl get bmh -n "{{ NAMESPACE }}" | grep -w ready | wc -l
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    kubernetes.core.k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     retries: 200
     delay: 20
-    register: ready_hosts
-    until: ready_hosts.stdout == "1"
+    register: bare_metal_hosts
+    vars:
+      query: "[? status.provisioning.state=='ready']"
+    until: (bare_metal_hosts is succeeded) and
+           (bare_metal_hosts.resources | json_query(query) | length == 1)
 
 # ---------- ---------- ---------- ---------- ---------- ---------- ---------- ----------
 #                       Upgrade controlplane components                                 |
 # ---------- ---------- ---------- ---------- ---------- ---------- ---------- ----------
   - name: Gather variables that were missing on the source cluster
-    shell: | 
-            kubectl get cm -n {{ NAMEPREFIX }}-system {{ NAMEPREFIX }}-baremetal-operator-ironic -o json |
-            jq -r '.data' | sed 's/: /=/'| cut -sf1- -d\" | sed 's/^/export/' | 
-            tr -d ',' | tr -d '"' | grep -E "_URL|_ENDPOINT" > /tmp/configmap.env.values
-            . /tmp/configmap.env.values
-            echo "export IRONIC_URL=${IRONIC_ENDPOINT}" >> /tmp/configmap.env.values
-            echo "export IRONIC_INSPECTOR_URL=${IRONIC_INSPECTOR_ENDPOINT}" >> /tmp/configmap.env.values
-    environment:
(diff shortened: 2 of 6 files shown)


lentzi90:
/hold
This PR is using the `CAPI_VERSION` variable, which is currently set to `v1alpha4`. I'll need to hard code to `v1alpha3` until #688 is fixed.

macaptain (approved):
Great work. The filter plugin looks really nice, great way of solving that problem, and I appreciate the unit tests for it too.

I checked the build time for the upgrade tests, and there's no change. The extra build output we get about the BareMetalHosts will help with debugging problems as we'll be able to see the full state of the k8s resource on failures.

There are some v1alpha3 references here. I am not sure what we need to do to prepare for v1alpha4 -> v1alpha5 testing, but I think this is outside the scope of your refactor.

furkatgofurov7 (commented):
This is great to see, good job!

furkatgofurov7:
> There are some v1alpha3 references here. I am not sure what we need to do to prepare for v1alpha4 -> v1alpha5 testing, but I think this is outside the scope of your refactor.

I think we have to make sure that if this PR lands before that API uplift work, we need to update those references most probably in #693 after rebasing.

lentzi90:
> I think we have to make sure that if this PR lands before that API uplift work in other repos and corresponding changes in this repo (#693), we need to update those references most probably in #693 after rebasing.

I can parameterize the versions in this PR and then wait for the uplift to be ready before merging. The reason I now hard-coded `v1alpha3` was to be able to run the tests (they are broken if I use `CAPI_VERSION` since that variable is already set to `v1alpha4`). But now that I know that the tests are green I can put the variable back and wait for the uplift. Do you think this would be ok @furkatgofurov7 ?

Xenwar:
> > I think we have to make sure that if this PR lands before that API uplift work in other repos and corresponding changes in this repo (#693), we need to update those references most probably in #693 after rebasing.
> 
> I can parameterize the versions in this PR and then wait for the uplift to be ready before merging. The reason I now hard-coded `v1alpha3` was to be able to run the tests (they are broken if I use `CAPI_VERSION` since that variable is already set to `v1alpha4`). But now that I know that the tests are green I can put the variable back and wait for the uplift. Do you think this would be ok @furkatgofurov7 ?

+1 for temporarily hard coding. Then creating a new PR for parameterize them.

lentzi90:
> +1 for temporarily hard coding. Then creating a new PR for parameterize them.

If this is OK then we can merge it directly. :slightly_smiling_face: It should be simple to parameterize later, just find/replace. :+1: 
But if the uplift is very close to ready then I think this can wait so we don't cause extra work.

Xenwar:
> > +1 for temporarily hard coding. Then creating a new PR for parameterize them.
> 
> If this is OK then we can merge it directly. 🙂 It should be simple to parameterize later, just find/replace. 👍
> But if the uplift is very close to ready then I think this can wait so we don't cause extra work.

I would go with merging it. In the worst case scenario, we just just create another hardcode value with the version after the uplift.

furkatgofurov7:
> But now that I know that the tests are green I can put the variable back and wait for the uplift. Do you think this would be ok @furkatgofurov7 ?

I was just referring that if we take this in now, we have to do small change on #693 regarding the API version, which is fine and not a blocker for this to be merged. Changes lgtm.

lentzi90:
Thanks @furkatgofurov7 ! Can you add `/lgtm` then? :wink:

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -4,72 +4,101 @@
...
-    shell: |
-           kubectl get machinedeployment -n {{NAMESPACE}} test1 -o json | 
-           jq '.spec.strategy.rollingUpdate.maxSurge=1|.spec.strategy.rollingUpdate.maxUnavailable=1' | 
-           kubectl apply -f-
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    community.kubernetes.k8s:
+      api_version: cluster.x-k8s.io/v1alpha3
+      kind: MachineDeployment
+      name: test1
```
kashifest, line 10:
```suggestion
      name: {{ CLUSTER_NAME }}
```

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -125,91 +152,141 @@
...
   - name: Perform upgrade on the source cluster
-    shell: |
-           # Add missing environment variables
-           . /tmp/configmap.env.values
-           clusterctl upgrade apply --management-group capi-system/cluster-api --contract v1alpha3
-           sleep 30
-           # restore secrets to fill missing secret fields
-           kubectl replace -f /tmp/secrets.with.values.yaml
-    ignore_errors: yes
+    ansible.builtin.command: clusterctl upgrade apply --management-group capi-system/cluster-api --contract v1alpha3
```
kashifest, line 223: @smoshiur1237 I hope you are also parameterizing these v1alpha3s in your PR

kashifest (commented):
lgtm, small nit. Good job

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
lentzi90: Thanks, fixed!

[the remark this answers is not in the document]
lentzi90: Thanks, fixed!

vm-setup/roles/v1aX_integration_test/filter_plugins/k8s_fields.py
```diff
@@ -0,0 +1,36 @@
...
+
+def k8s_backup(resources):
+    """
+    Take a backup of k8s resources by stripping away problematic fields.
+    Fields that are automatically created and should normally not be touched
+    will be removed so that the resources can safely be restored from this
+    "backup".
+    Removed fields under .metadata: uid, managedFields, resourceVersion,
+    creationTimestamp
+    """
```
mboukhalfa, line 18: k8s `apply` https://docs.ansible.com/ansible/latest/collections/community/kubernetes/k8s_module.html#parameter-apply supposed to ignore properties that are automatically generated. Does not it help with those fields ?

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -4,72 +4,101 @@
...
 
   - name: Update maxSurge and maxUnavailable fields
-    shell: |
-           kubectl get machinedeployment -n {{NAMESPACE}} test1 -o json | 
-           jq '.spec.strategy.rollingUpdate.maxSurge=1|.spec.strategy.rollingUpdate.maxUnavailable=1' | 
-           kubectl apply -f-
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    community.kubernetes.k8s:
+      api_version: cluster.x-k8s.io/v1alpha3
```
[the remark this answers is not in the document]
lentzi90: It won't work until we have support for v1alpha4. `CAPI_VERSION` is already set to v1alpha4 so we must wait for the uplift until we can start using this variable. :slightly_frowning_face:  See discussion above.

vm-setup/roles/v1aX_integration_test/filter_plugins/k8s_fields.py
lentzi90: Interesting! Will check, thanks!

lentzi90: Unfortunately it doesn't work. :slightly_frowning_face: I guess it ignores the auto generated properties when doing the comparison, but not when doing the actual update. This is how I tested it:

```yaml
- name: Create deployment
      kubernetes.core.k8s:
        api_version: apps/v1
        kind: Deployment
        name: test
        namespace: default
        resource_definition:
          metadata:
            labels:
              app: test
          spec:
            replicas: 1
            selector:
              matchLabels:
                app: test
            template:
              metadata:
                labels:
                  app: test
              spec:
                containers:
                  - name: nginx
                    image: nginx

    - name: Get deployment
      kubernetes.core.k8s_info:
        api_version: apps/v1
        kind: Deployment
        name: test
        namespace: default
      register: deployment

    - debug:
        msg: "{{ deployment }}"

    - name: Scale deployment
      kubernetes.core.k8s_scale:
        api_version: apps/v1
        kind: Deployment
        name: test
        namespace: default
        replicas: 2

    - name: Restore deployment
      kubernetes.core.k8s:
        api_version: apps/v1
        kind: Deployment
        name: test
        namespace: default
        apply: yes
        resource_definition: "{{ deployment.resources[0] }}"
```

In the last step there is always a failure due to status 409 (conflict). I believe this is because the resource version has changed. So the server knows that we are trying to update an object that we haven't seen the latest version of yet.

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -4,72 +4,101 @@
...
+      kind: MachineDeployment
+      name: "{{ CLUSTER_NAME }}"
+      namespace: "{{ NAMESPACE }}"
+      resource_definition:
+        spec:
+          strategy:
+            rollingUpdate:
+              maxSurge: 1
+              maxUnavailable: 1
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
```
furkatgofurov7, line 18: In general, are those scale operations working as intended? Because we have seen it silently ignoring the intended change and setting the default value for `maxSurge` field in node reuse tests which resulted in hiding the bug. 

cc @mboukhalfa

furkatgofurov7 (commented):
Overall looks good to me.

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
mboukhalfa: This should work because in node_reuse I was using `apply: yes ` which ignores those fields.

lentzi90: Yes scaling has worked well here. :slightly_smiling_face:

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -125,91 +152,141 @@
...
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    ansible.builtin.command: clusterctl upgrade apply --management-group capi-system/cluster-api --contract v1alpha3
+    environment: "{{ ironic_env | combine(kubeconfig_env) }}"
+    vars:
+      kubeconfig_env:
+        KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+
+  # TODO: Can we check this somehow instead of just waiting?
+  # Relevant issue: https://github.com/kubernetes-sigs/cluster-api/issues/4474
```
fmuyassarov, line 210: Maybe we could add a loop to wait for the pods in the capi-system to be available ?

fmuyassarov (commented):
Really good work!

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
lentzi90: Do you know if that will be enough? I saw that there is also some things in `capi-webhook-system`, `capi-kubeadm-control-plane-system` and `capi-kubeadm-bootstrap-system`. Since I'm not really sure what we would have to wait for I just left it as a pause for now, but I can certainly add a wait for some/all of these deployments/pods if you know which are important! :slightly_smiling_face:

fmuyassarov: Ah true, there is also webhook namespace. Maybe we could take the namespaces that starts with `capi*`. But we could it in a separate PR, this should not be a blocker.

lentzi90:
/hold
Wait with this until https://github.com/metal3-io/metal3-dev-env/pull/721 is done

jan-est:
LGTM

PR #692, opened 2021-06-14
Cleanup variables for CAPI and CAPM3

lentzi90:
- Remove duplicate keys for CAPI_VERSION and CAPM3_VERSION
- Remove variables CAPI_API_VERSION and CAPM3_API_VERSION that were only
used for upgrade tests, and replace them with CAPI_VERSION and
CAPM3_VERSION.

Note that CAPI_VERSION is already set to `v1alpha4` in the CI, whereas CAPI_API_VERSION was set to `v1alpha3`, so to keep the current behavior the version is now hard-coded. There is [an issue](https://github.com/metal3-io/metal3-dev-env/issues/688) for parameterizing this later when support is in place for `v1alpha4`.

### DIFF
--- a/scripts/feature_tests/upgrade/upgrade_vars.sh
+++ b/scripts/feature_tests/upgrade/upgrade_vars.sh
@@ -18,9 +18,6 @@ export CAPM3_REL_TO_VERSION
 export CAPIRELEASE="v0.3.15"
 export CAPI_REL_TO_VERSION="v0.3.16"
 
-export CAPI_API_VERSION="v1alpha3"
-export CAPM3_API_VERSION="v1alpha4"
-
 export KUBERNETES_VERSION="v1.21.1"
 export UPGRADED_K8S_VERSION="v1.21.2"
 
--- a/vm-setup/roles/v1aX_integration_test/templates/Metal3MachineTemplate.yml
+++ b/vm-setup/roles/v1aX_integration_test/templates/Metal3MachineTemplate.yml
@@ -1,10 +1,10 @@
-apiVersion: infrastructure.cluster.x-k8s.io/{{CAPM3_API_VERSION}}
+apiVersion: infrastructure.cluster.x-k8s.io/{{CAPM3_VERSION}}
 kind: Metal3MachineTemplate
 metadata:
   name: {{ M3MT_NAME }}
   namespace: metal3
   ownerReferences:
-  - apiVersion: cluster.x-k8s.io/{{CAPI_API_VERSION}}
+  - apiVersion: cluster.x-k8s.io/{{CAPI_VERSION}}
     kind: Cluster
     name: {{ CLUSTER_NAME }}
     uid: {{ CLUSTER_UID }}
--- a/vm-setup/roles/v1aX_integration_test/vars/main.yml
+++ b/vm-setup/roles/v1aX_integration_test/vars/main.yml
@@ -91,11 +91,6 @@ CAPM3PATH: "{{M3PATH}}/cluster-api-provider-metal3"
 
 CAPM3_REL_TO_VERSION: "{{ lookup('env', 'CAPM3_REL_TO_VERSION') }}"
 CAPI_REL_TO_VERSION: "{{ lookup('env', 'CAPI_REL_TO_VERSION') }}"
-CAPM3_API_VERSION: "{{ lookup('env', 'CAPM3_API_VERSION') }}"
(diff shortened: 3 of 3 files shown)


scripts/feature_tests/upgrade/upgrade_vars.sh
```diff
@@ -18,8 +18,5 @@ export CAPM3_REL_TO_VERSION
 export CAPIRELEASE="v0.3.15"
 export CAPI_REL_TO_VERSION="v0.3.16"
 
-export CAPI_API_VERSION="v1alpha3"
-export CAPM3_API_VERSION="v1alpha4"
```
smoshiur1237, line 22: This variables are needed for upgrade test to build the environment. We shouldn't delete these variables.

smoshiur1237 (changes requested):

scripts/feature_tests/upgrade/upgrade_vars.sh
smoshiur1237: Now its passing the CI run because we are using by default capi v1a3 and capm3 v1alpha4. But when we introduce the change for capi v1a4 it will break.

lentzi90: So these are specifically only here for testing the upgrade? Could we then maybe rename them to something like `CAPI_UPGRADE_VERSION` or `CAPI_OLD_VERSION`? As it is now I think it is very confusing with both `CAPI_VERSION` and `CAPI_API_VERSION`. :thinking:

smoshiur1237: @Xenwar can you please have a look in this change?
/cc @Xenwar

vm-setup/roles/v1aX_integration_test/templates/Metal3MachineTemplate.yml
```diff
@@ -1,10 +1,10 @@
-apiVersion: infrastructure.cluster.x-k8s.io/{{CAPM3_API_VERSION}}
+apiVersion: infrastructure.cluster.x-k8s.io/{{CAPM3_VERSION}}
```
Xenwar, line 1: ok with renaming

vm-setup/roles/v1aX_integration_test/templates/Metal3MachineTemplate.yml
```diff
@@ -1,10 +1,10 @@
...
 kind: Metal3MachineTemplate
 metadata:
   name: {{M3MT_NAME}}
   namespace: metal3
   ownerReferences:
-  - apiVersion: cluster.x-k8s.io/{{CAPI_API_VERSION}}
+  - apiVersion: cluster.x-k8s.io/v1alpha3
```
Xenwar, line 7: This should be parametrised as well. 
We could run the same test with v1alpha4 and v1alpha5 as well.

Xenwar (changes requested):

scripts/feature_tests/upgrade/upgrade_vars.sh
Xenwar: Added a comment

vm-setup/roles/v1aX_integration_test/templates/Metal3MachineTemplate.yml
lentzi90: Yes I know, but we cannot use `CAPI_VERSION` yet. In the CI it is set to `v1alpha4` and it doesn't work. See https://github.com/metal3-io/metal3-dev-env/issues/688

lentzi90: Ok let's wait for v1alpha4 to land. Then I'll use `CAPI_VERSION` here. :slightly_smiling_face:

scripts/feature_tests/upgrade/upgrade_vars.sh
mboukhalfa: @Xenwar This is supposed to be resolved I think ? I cannot find your comment regarding this

namnx228 (commented):
/lgtm

furkatgofurov7 (approved):
Last time something similar was merged when we have not run all the tests and broke the CI even though we thought changes are not related, so to be sure I am triggering remaining tests. Otherwise I am happy with the change. 
 
/test-features
/approve

furkatgofurov7:
Ops, it did not trigger the tests. Fine, I will trigger it from the master run in CI.

PR #695, opened 2021-06-15
Refactor ironic upgrade to avoid shell

lentzi90:
Note that I removed the scale down of capm3-ironic since this is not really necessary. As far as I can understand, it was there to make it easier to check that only new pods were running (i.e. if there are no old pods you can ignore them). I worked around this by looking at the deployments `updatedReplicas`. This way we can know if there is an old or new pod running.


lentzi90:
Waiting for https://github.com/metal3-io/baremetal-operator/pull/918 to simplify the upgrade steps for ironic. Once that PR is merged, we can skip the scale down and scale up since it will be done automatically.

/hold

lentzi90:
https://github.com/metal3-io/baremetal-operator/pull/918 has been merged so we can move forward here.
/unhold
/test-integration
/test-v1a4-centos-integration

fmuyassarov (approved):

PR #696, opened 2021-06-15
Refactor upgrade tests for k8s

lentzi90:
- Use k8s and k8s_info modules instead of shell
- Use label selectors instead of greping for patterns

Note: This still has the CAPI version hard-coded to v1alpha3 while waiting for v1alpha4 to land.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
@@ -278,8 +278,6 @@
   #   # failed_when: api_status.apis not contains the upgraded m3c resource(s)?
 
   - name: Verify upgraded API resource for Metal3Clusters
-    # spec.names.shortNames[]
-    # status.acceptedNames.shortNames[]
     kubernetes.core.k8s_info:
       api_version: apiextensions.k8s.io/v1
       kind: CustomResourceDefinition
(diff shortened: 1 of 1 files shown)


vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -293,181 +293,212 @@
...
-    environment:
-      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+  - name: Get cluster uid
+    community.kubernetes.k8s_info:
+      api_version: cluster.x-k8s.io/v1alpha3
+      kind: Cluster
+      name: {{ CLUSTER_NAME }}
+      namespace: {{ NAMESPACE }}
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     register: CLSTR_UID
```
Xenwar, line 303: What if you rename `CLSTR_UID` to `CLSTR_RESOURCE`. That way, it will not be confused with `CLUSTER_UID`

lentzi90: Good point, will fix!

lentzi90:
Rerunning upgrade tests a few times to see if reverting to shell + kubectl helps with the (flaky) hanging ansible retries.
/test-upgrade-features

lentzi90:
/hold
Wait with this until #721 is done

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -352,52 +352,61 @@
...
+  # Ref https://github.com/ansible/ansible/issues/30411
+  #
+  # Attempted workarounds:
+  # - Use timeout (https://docs.ansible.com/ansible/latest/reference_appendices/playbooks_keywords.html#term-timeout)
+  #   This prevents hanging but fails the whole task instead of retrying.
+  #
+  # Issues that may be related
+  # - https://github.com/kubernetes-client/python/issues/1402
+  # - https://github.com/kubernetes-client/python/issues/602
   - name: Verify that controlplane nodes using the new node image
```
mboukhalfa, line 410: # This is tricky. The control plane nodes are being replaced one by one.
  # This naturally causes some interruptions in the communication with the API.
  # Unfortunately it seems like the k8s_info module can get stuck in these
  # situations (i.e. it doesn't fail, it just hangs).
  # Ref https://github.com/ansible/ansible/issues/30411
  #
  # Attempted workarounds:
  # - Use timeout (https://docs.ansible.com/ansible/latest/reference_appendices/playbooks_keywords.html#term-timeout)
  #   This prevents hanging but fails the whole task instead of retrying.
  #
  # Issues that may be related
  # - https://github.com/kubernetes-client/python/issues/1402
  # - https://github.com/kubernetes-client/python/issues/602

PR #703, opened 2021-06-21
Add a script to run a command with environment variables

wgslr:
This allows executing a script within the environment resolved by metal3-dev-env, i.e., with variables set to their default values if missing and image URLs derived from IMAGE_OS variable. Currently this will be used for running e2e tests in CAPM3. (see [CAPM3#218](https://github.com/metal3-io/cluster-api-provider-metal3/pull/218)). Additionally, exports a few variables used in the CAPM3 PR.

### DIFF
--- a/lib/network.sh
+++ b/lib/network.sh
@@ -74,6 +74,7 @@ if [[ "$CLUSTER_APIENDPOINT_IP" == *":"* ]]; then
 else
   export CLUSTER_APIENDPOINT_HOST="$CLUSTER_APIENDPOINT_IP"
 fi
+export CLUSTER_APIENDPOINT_PORT=${CLUSTER_APIENDPOINT_PORT:-"6443"}
 
 # Calculate DHCP range
 network_address dhcp_range_start "$PROVISIONING_NETWORK" 10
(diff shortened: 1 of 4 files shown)


lib/network.sh
```diff
@@ -74,6 +74,7 @@ if [[ "$CLUSTER_APIENDPOINT_IP" == *":"* ]]; then
 else
   export CLUSTER_APIENDPOINT_HOST="$CLUSTER_APIENDPOINT_IP"
 fi
+export CLUSTER_APIENDPOINT_PORT=${API_ENDPOINT_PORT:-"6443"}
```
macaptain, line 77: I think the RHS variable `API_ENDPOINT_PORT` isn't set anywhere in the dev env. There's `CLUSTER_APIENDPOINT_IP` though, so I think it would make sense to match the naming on the LHS and RHS, call this `CLUSTER_APIENDPOINT_PORT` on both sides and document the name of this variable in vars.md.

lentzi90 (approved):

lib/network.sh
wgslr: Done

PR #704, opened 2021-06-21
Update notes from CentOS 8 Stream to CentOS 8.3

namnx228:
This PR fix the CentOS version in the README from 8 to 8.3


furkatgofurov7:
/retitle Update notes from CentOS 8 Stream to CentOS 8.3

Changing the PR title since it is confusing, as we already reverted back to CentOS 8.3 and it is just updating leftover notes.

furkatgofurov7 (commented):
/lgtm

PR #706, opened 2021-06-21
Add option to enable the reverse proxy for Ironic-inspector

namnx228:
When TLS is used, add option to enable the reverse proxy for Ironic-inspector


namnx228:
The option will not be added: https://github.com/metal3-io/baremetal-operator/pull/923#issuecomment-865751014.
/close

PR #708, opened 2021-06-22
Verify existence of resources before filtering

lentzi90:
For various reasons it may happen that communication with the k8s API
fails or an empty list of resources is returned. When attempting to
filter this empty result, it naturally causes errors that can fail the
whole task. To avoid this we can check that the module succeeded in
communicating with k8s and that the list of resources is not empty. Now
if the module fails for some reason it can retry instead of stopping due
to an error when trying to filter.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/move.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/move.yml
@@ -68,7 +68,8 @@
     register: target_running_pods
     retries: 20
     delay: 20
-    until: target_running_pods.resources | length == 0
+    until: (target_running_pods is succeeded) and
+           (target_running_pods.resources | length == 0)
 
   # Install Ironic
   - name: Install Ironic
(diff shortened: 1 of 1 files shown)


vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -104,7 +106,9 @@
     register: provisioned_machines
     retries: 50
     delay: 20
-    until: provisioned_machines.resources | filter_phase("running") | length == (NUMBER_OF_BMH | int)
+    until: (provisioned_machines is succeeded) and
+           (provisioned_machines.resources | length > 0) and
```
macaptain, line 110: I don't think we have to check that the length of the resources is > 0. The filter should handle an empty list fine.

lentzi90: Right, that's true! Fixing...

lentzi90: Done

fmuyassarov (commented):
/approve

PR #711, opened 2021-06-22
Switch to DIY callback plugin for Ansible

lentzi90:
The DIY callback plugin is configured to print only the status of k8s resources instead of everything (but otherwise work as before). Ansible verbosity is also turned down to default.

The goal is to reduce the amount of useless logs produced by the CI but still show relevant/interesting parts.

My reasoning is as follows. Note that all of this is very much open for discussion. This is just my point of view.

1. We get a lot of output that is not very useful because we run ansible with `-vvv`. If a task fails, it will usually anyway print extra information about the error, so we skip `-vvv` and should still get the most important information.
2. The `k8s_info` module that we are now using a bit more prints everything (including all the managed fields and what not) when running with `-vvv`. This is too much. Without `-vvv` it doesn't print anything though, and that is too little. It becomes hard to tell what is going on for example when a task is using `until` to wait for some node/machine/bmh.
  Solution: use the DIY callback and configure it to print only the `status` field of all resources (+ name and kind). This gives us the most valuable information while still reducing the amount of logs quite a lot. (Note that we will still get the full output if a task fails.)
3. (This could be left as a follow up.) Ansible [blocks can be used for error handling](https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html#handling-errors-with-blocks). We could benefit from this by gathering extra information about the environment in case a task fails. This works well in cases where it is known already what to look for. For example, we may choose to gather information about all Machines, BareMetalHosts, Pods and Nodes if a task fails and then print all of it. Note that this is in addition to the above. We still get the full output from the failed task, but after it we can add more.
4. (Extra) We could also consider disabling the echo of some/all bash commands. They are quite repetitive and in case of failure we would still get the line where it failed. The main drawback that I see with this is that it may be less obvious what value some environment variables have.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -44,7 +44,7 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
   -e "metal3_dir=$SCRIPTDIR" \
   -e "virthost=$HOSTNAME" \
   -i vm-setup/inventory.ini \
-  -b -vvv vm-setup/install-package-playbook.yml
+  -b vm-setup/install-package-playbook.yml
 
 # shellcheck disable=SC1091
 source lib/network.sh
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -28,7 +28,7 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -e "nodes_file=$NODES_FILE" \
     -e "node_hostname_format=$NODE_HOSTNAME_FORMAT" \
     -i vm-setup/inventory.ini \
-    -b -vvv vm-setup/setup-playbook.yml
+    -b vm-setup/setup-playbook.yml
 
 # Usually virt-manager/virt-install creates this: https://www.redhat.com/archives/libvir-list/2008-August/msg00179.html
 if ! sudo virsh pool-uuid default > /dev/null 2>&1 ; then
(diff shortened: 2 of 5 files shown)


vm-setup/roles/v1aX_integration_test/tasks/main.yml
```diff
@@ -40,9 +40,34 @@
...
+        - api_version: v1
+          kind: Node
+        - api_version: v1
+          kind: Pod
+    - name: Print debug info
+      debug:
+        msg: "{{ resources }}"
+    - name: Fail due to earlier error
+      ansible.builtin.fail:
+        msg: Failed due to earlier errors
```
lentzi90, line 70: This is an example of what we could do to gather more information in case some task fails. The `rescue` part is run if there is a failure in the `block`.

fmuyassarov (commented):
I personally like that we less output instead of having too much stdout which we don't really read and it just takes extra space and makes it complicated to read the logs.
I'm +1 for this.

smoshiur1237:
Significant improvement with less logs and will be easy to read the logs. Good work.

lentzi90:
/test-integration
/test-v1a4-centos-integration

Alright, I think it is ready to go now. Removed the example part. Please review/approve if you like it :slightly_smiling_face:

kashifest (commented):
Thanks for taking care of this.
/approve

PR #223, opened 2020-02-14
Add CentOS 8 support

hardys:
This depends on https://github.com/metal3-io/metal3-dev-env/pull/215

PR #712, opened 2021-06-23
Uplift kustomize to v4.1.3

wgslr:
This is needed because only the newer kustomize versions ignore patches for missing targets. This is important for example in CAPM3, where a recent PR [#223](https://github.com/metal3-io/cluster-api-provider-metal3/pull/223) broke e2e tests when using the old kustomize version.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -206,7 +206,7 @@ else
 fi
 
 # Kustomize version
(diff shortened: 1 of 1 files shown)


lentzi90 (approved):

wgslr:
Duplicate of #664

PR #716, opened 2021-07-01
Check NUM_NODES before creating VMs

### LINKED ISSUE #715: Upgrade test is failing with default NUM_NODES
The number of control plane nodes and worker nodes is set directly in the Makefile:
https://github.com/metal3-io/metal3-dev-env/blob/c508a7a536f884fbf2fb2376e0058339f4349199/scripts/feature_tests/upgrade/Makefile#L2-L3
However, these values do not work with the default value for `NUM_NODES`, which is 2. (Se [here](https://github.com/metal3-io/metal3-dev-env/blob/c508a7a536f884fbf2fb2376e0058339f4349199/lib/common.sh#L144).)

The issue I have with this is that it is easy to forget to set `NUM_NODES=4` before running `make feature_tests_upgrade` and it takes a long time before the issue is discovered. (It creates the VMs, initializes the source cluster, installs everything in the source cluster, creates the target cluster, and only then does it fail when the expected number of nodes doesn't join the target cluster.)

I think we should fix this in one of two ways:

1. Make the test detect the problem early and fail so that we don't waste time waiting for the inevitable.
2. Automatically create a suitable number of nodes when running tests that requires it. Since we already set the other variables for upgrade tests, I don't see why we wouldn't also be able to set `NUM_NODES`.

Potential pitfalls:

- The `NUM_OF_MASTER_REPLICAS` and `NUM_OF_WORKER_REPLICAS` are set *after* the VMs are created currently. We would need to move it to before that if we want to set `NUM_NODES` at the same time.

lentzi90:
Fixes #715 

Previously we only set `NUM_OF_MASTER_REPLICAS=3` and `NUM_OF_WORKER_REPLICAS=1` for upgrade tests, but `NUM_NODES` was left at default (2) unless overridden. This moves the variables to `upgrade_vars.sh` from the upgrade Makefile and sets `NUM_NODES` in addition. The reason for moving the variables is that `upgrade_vars.sh` is sourced before creating the VMs, whereas the Makefile is executed after this.

Edit: Based on the discussion below this was changed so that we do *not* set `NUM_NODES`, but instead we check that the value is reasonable directly in `common.sh`. This allows us to error out immediately and notify the user if there is a problem. It also avoids and new issues with `make clean`.


lentzi90:
I found out now that there is one potential issue with this. The cleanup (`make clean`) also relies on `NUM_NODES` to determine the number of VMs to remove. This means that if I now run `make feature_tests_upgrade` without setting `NUM_NODES` I'll get 4 VMs, and when I later run `make clean` it will only remove 2 of them.

I'm not sure how much of a problem this really is, but I would be happy for suggestions on how to resolve it!

namnx228:
> I found out now that there is one potential issue with this. The cleanup (`make clean`) also relies on `NUM_NODES` to determine the number of VMs to remove. This means that if I now run `make feature_tests_upgrade` without setting `NUM_NODES` I'll get 4 VMs, and when I later run `make clean` it will only remove 2 of them.
> 
> I'm not sure how much of a problem this really is, but I would be happy for suggestions on how to resolve it!

Should we add `NUM_NODES` to the `upgrade-vars.sh` also?

namnx228:
> I found out now that there is one potential issue with this. The cleanup (`make clean`) also relies on `NUM_NODES` to determine the number of VMs to remove. This means that if I now run `make feature_tests_upgrade` without setting `NUM_NODES` I'll get 4 VMs, and when I later run `make clean` it will only remove 2 of them.
> 
> I'm not sure how much of a problem this really is, but I would be happy for suggestions on how to resolve it!

I think when `make clean` is called, we can determine how many VMs we have by something like; `sudo virsh list --all | grep node -c` and assign the result to `NUM_NODES`.

scripts/feature_tests/upgrade/upgrade_vars.sh
```diff
@@ -23,3 +23,7 @@ export CAPM3_API_VERSION="v1alpha4"
 
 export KUBERNETES_VERSION="v1.21.1"
 export UPGRADED_K8S_VERSION="v1.21.2"
+
+export NUM_NODES="4"
```
macaptain, line 27: One problem with putting `NUM_NODES` here is that it will prevent overriding this value from the Jenkins jobs:

https://github.com/metal3-io/project-infra/blob/60aafbe4bcf82f2233c6c7bd3be1335dbd5deadd/jenkins/scripts/integration_test.sh#L37

(I'm not sure this is such a useful feature, especially since it seems the replica numbers are hardcoded elsewhere, but it is there.)

macaptain (commented):
I think this would be great to fix, as I've stubbed my toe on this problem before as well, and it is annoying.

The simplest thing may be a check that `NUM_OF_MASTER_REPLICAS + NUM_OF_WORKER_REPLICAS <= NUM_NODES` somewhere early on in Ansible, which would sidestep the `make clean` problem as well. There may be a smarter way to get the build to 'fix itself' if `NUM_NODES` is too low for the test suite.

scripts/feature_tests/upgrade/upgrade_vars.sh
lentzi90: That is a good point. My reasoning was that since we already hardcode `NUM_OF_MASTER_REPLICAS` and `NUM_OF_WORKER_REPLICAS` we might as well do the same with `NUM_NODES`. But maybe it would be better to make them all configurable? Not quite sure where or how though. :thinking:

lentzi90:
> I think when make clean is called, we can determine how many VMs we have by something like; sudo virsh list --all | grep node -c and assign the result to NUM_NODES.

This sounds simple and effective, I like it! Another option I'm considering is using the `NODES_FILE`. This should point to a file where all ironic nodes are defined. :slightly_smiling_face: 

> The simplest thing may be a check that NUM_OF_MASTER_REPLICAS + NUM_OF_WORKER_REPLICAS <= NUM_NODES somewhere early on in Ansible, which would sidestep the make clean problem as well.

I tried this first actually. The problem is that the playbook that creates the VMs doesn't really know if it is for an upgrade test or something else. So it would still take quite some time before it can fail. Now that I moved the variables, it would be possible to do it earlier in the `upgrade_vars.sh` though. So instead of setting `NUM_NODES` we would just check the value and error out if it is too low?

scripts/feature_tests/upgrade/upgrade_vars.sh
macaptain: I'm all in favour of having things be less configurable and 'just work'.

Usually I would set these variables in `<username>_config.sh`, I didn't realise before that they were hardcoded in the case of upgrade tests.

I am sure there is a sophisticated way to do this, but 'fail fast' is a good pattern in Ansible that means you don't have to guess the user's intent about what they want to do in their environment.

lentzi90:
Alright I think I have a better solution! Instead of hardcoding `NUM_NODES` for upgrade tests, we just check that the values make sense in `common.sh`. This avoids any (new) issues with cleanup and fails really fast (~1 sec) if `NUM_NODES` is not properly configured.

scripts/feature_tests/upgrade/upgrade_vars.sh
lentzi90: Agreed! Updated now to fail fast instead of hardcoding `NUM_NODES`

namnx228 (approved):

fmuyassarov (approved):
/approve

PR #717, opened 2021-07-02
WIP: Test registration failure with host busy after pivoting

furkatgofurov7:


furkatgofurov7:
This was to test the pivoting workflow in m3-dev-env and see if we hit the same issue described in [baremetal-operator#930](https://github.com/metal3-io/baremetal-operator/issues/930). Once pivoting is done, worker node was shutdown using reboot annotation and it powered off the node successfully and BMO was able to remove the annotation and power on the node as expected. 

/close

PR #722, opened 2021-07-07
Upgrade Ansible k8s modules to 2.1.1 

mboukhalfa:
This PR upgrade Ansible k8s module to `kubernetes.core 2.1.1` that requires pip3 package` kubernetes>=12.0.0`
Openshift was removed since it is a requirement for the old k8s `version 1.2.1 `and it depends on `kubernetes==11.0.0` which also overrides the new version needed `>=12.0.0`

### DIFF
--- a/vm-setup/roles/packages_installation/defaults/main.yml
+++ b/vm-setup/roles/packages_installation/defaults/main.yml
@@ -43,7 +43,7 @@ packages:
         - libssl-dev
     pip3:
      - python-apt
-     - openshift==0.11.2
+     - kubernetes==17.17.0
      - pyYAML
      - virtualbmc
      - lxml
@@ -86,6 +86,7 @@ packages:
          - httpd-tools
         pip3:
           - jmespath
+          - kubernetes==17.17.0
 DAEMON_JSON_PATH: "{{ metal3_dir }}/vm-setup/roles/packages_installation/files"
 CONTAINER_RUNTIME: "{{ lookup('env', 'CONTAINER_RUNTIME') }}"
 OS_VERSION_ID: "{{ lookup('env', 'OS_VERSION_ID') }}"
--- a/vm-setup/requirements.yml
+++ b/vm-setup/requirements.yml
@@ -1,3 +1,8 @@
-- src: https://github.com/fubarhouse/ansible-role-golang.git
-  version: master
-  name: fubarhouse.golang
+roles:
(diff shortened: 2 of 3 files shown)


lentzi90 (commented):
Great!
I think we may need to update tasks to use `kubernetes.core` instead of `community.kubernetes` also. But let's see how the tests go! :slightly_smiling_face:

mboukhalfa:
> Great!
> I think we may need to update tasks to use `kubernetes.core` instead of `community.kubernetes` also. But let's see how the tests go! slightly_smiling_face

Indeed it did not work with latest 17.17.0 getting back to 12.0.1

lentzi90 (approved):
LGTM

vm-setup/roles/packages_installation/defaults/main.yml
```diff
@@ -43,7 +43,7 @@ packages:
         - libssl-dev
     pip3:
      - python-apt
-     - openshift==0.11.2
+     - kubernetes==17.17.0
```
furkatgofurov7, line 46: is this an ansible k8s module version?

vm-setup/roles/packages_installation/defaults/main.yml
```diff
@@ -43,7 +43,7 @@ packages:
         - libssl-dev
     pip3:
      - python-apt
-     - openshift==0.11.2
```
furkatgofurov7, line 46: why do we need to remove this?

furkatgofurov7 (commented):
Some questions inline and triggering upgrade job since it was missing:

vm-setup/roles/packages_installation/defaults/main.yml
mboukhalfa: Ansible `k8s 2.1.0 `requires pip package `kubernetes>=12.0.0 `

furkatgofurov7: Can you please update the PR description with that note and if possible reference to source info, otherwise it in unclear why do we need that package. Thanks.

vm-setup/roles/packages_installation/defaults/main.yml
mboukhalfa: `openshift` was a requirement for the old k8s version `1.2.1` and it depends on `kubernetes==11.0.0`  which overrides the new version needed >=12.0.0

furkatgofurov7 (commented):
/lgtm

PR #729, opened 2021-07-09
Make Ironic credential available in CAPM3 e2e tests.

namnx228:
Currently, the ironic credential detail is not accessible from the e2e test. This PR aims to fix it.

### DIFF
--- a/scripts/run_command.sh
+++ b/scripts/run_command.sh
@@ -21,5 +21,11 @@ source "${METAL3_DIR}/lib/network.sh"
 # shellcheck disable=SC1090
 # shellcheck disable=SC1091
 source "${METAL3_DIR}/lib/images.sh"
+# shellcheck disable=SC1090
+# shellcheck disable=SC1091
+source "${METAL3_DIR}/lib/ironic_basic_auth.sh"
+# shellcheck disable=SC1090
+# shellcheck disable=SC1091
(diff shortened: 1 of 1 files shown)


scripts/run_command.sh
```diff
@@ -21,5 +21,8 @@ source "${METAL3_DIR}/lib/network.sh"
 # shellcheck disable=SC1090
 # shellcheck disable=SC1091
 source "${METAL3_DIR}/lib/images.sh"
+# shellcheck disable=SC1090
+# shellcheck disable=SC1091
+source "${METAL3_DIR}/lib/ironic_basic_auth.sh"
```
furkatgofurov7, line 26: Only basic auth, don't you need `ironic_tls_setup.sh` too?

furkatgofurov7:
> /test-v1a5-e2e

We can trigger e2e tests only in PRs' towards CAPM3 master branch.

scripts/run_command.sh
namnx228: Good suggestion! I check and see that calling again the `ironic_tls_setup.sh` doesn't regenerate the certificates, but can ensure that we can access the CA certificate from the e2e test. We may need it in the future.

furkatgofurov7 (approved):
/approve

PR #731, opened 2021-07-13
Use kubernetes.core module for Ansible

lentzi90:
We have removed the dependency for the old community.kubernetes module
already so we need to make sure all tasks use the new kubernetes.core
module instead.

It is not strictly needed to use the full name (i.e. `kubernetes.core.k8s` not `k8s`), however it is recommended to use the full name and it makes it more obvious what module it used. The most important thing is to get rid of all references to `community.kubernetes` and replace them with `kubernetes.core`.

This should fix the error seen [here](https://jenkins.nordix.org/view/Airship/job/airship_master_feature_tests_upgrade_ubuntu/416/console)

Edit: Added another commit to fix [this](https://jenkins.nordix.org/view/Airship/job/airship_master_feature_tests_ubuntu/557/console). The status field can be missing right when the cluster is created, so we have to set a default value to avoid issues with undefined values.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
@@ -23,7 +23,7 @@
           (IMAGE_OS == "")
 
   - name: Check if cluster deprovisioning started.
-    k8s_info:
+    kubernetes.core.k8s_info:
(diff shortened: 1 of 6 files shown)


furkatgofurov7:
Let's merge this to unblock the CI. Thanks for the fix @lentzi90 !

PR #733, opened 2021-07-14
Decrease delay when waiting for reboot

lentzi90:
When the delay is large there is a risk that the VM have time to start
again before we see the "shutdown" state.

This is an attempt to fix the flaky issue we are seeing [here](https://jenkins.nordix.org/view/Airship/job/airship_master_feature_tests_ubuntu/559/console).

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
@@ -41,8 +41,8 @@
       command: list_vms
       state: shutdown
     register: shutdown_vms
-    retries: 50
-    delay: 10
+    retries: 170
+    delay: 3
     until: WORKER_VM in shutdown_vms.list_vms
     become: yes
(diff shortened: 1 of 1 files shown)


vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
```diff
@@ -42,7 +42,7 @@
       state: shutdown
     register: shutdown_vms
     retries: 50
```
mboukhalfa, line 44: As we have decreased the delay I think we need to increase the retries to maintain the same time needed to wait ? 
```suggestion
    retries: 166
```

fmuyassarov: otherwise that many retries may not be enough too :)

lentzi90: Thanks, fixed!

PR #737, opened 2021-07-21
Fix k8s_info hanging task in remediation

mboukhalfa:
This PR fixes a task that hangs sometimes in Ansible by reverting the module k8s_info to shell

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/remediation.yml
@@ -168,20 +168,13 @@
       - "{{ MASTER_BMH_2 }}"
 
   - name: Wait until powered on master nodes become Ready
-    k8s_info:
-      api_version: v1
-      kind: nodes
-      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    register: nodes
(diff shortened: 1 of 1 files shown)


fmuyassarov (commented):
Not blocking it but I'm just wondering if we have any idea why it is hanging specifically on this specific `k8s_info` task

mboukhalfa:
> Not blocking it but I'm just wondering if we have any idea why it is hanging specifically on this specific `k8s_info` task

We have faced many tasks with k8s_info and until loop that hang forever either before any try or after some tries. So far we are revert those tasks to shells and trying to bring a way to reproduce the same issue outside our dev-env

fmuyassarov:
> > Not blocking it but I'm just wondering if we have any idea why it is hanging specifically on this specific `k8s_info` task
> 
> We have faced many tasks with k8s_info and until loop that hang forever either before any try or after some tries. So far we revert those tasks to shells and trying to come up with a simple way to reproduce the same issue outside our dev-env

Do we need to update all the `k8s_info` tasks to shell?

furkatgofurov7 (commented):
TBH, I do not see much of a benefit from reverting this back to shell since it is passing in CI as is now and even shell might hang and that can't be avoided under certain environment flakiness in CityCloud which is beyond our control.

PR #738, opened 2021-07-23
Decrease delay when waiting for node to be NotReady

mboukhalfa:
The task that checks when the node becomes `NotReady` may fail since the delay needed for the node to change from `notReady` to ready is close to the delay of task. This may allow the node to become ready before the task checks the `notReady` status.

This PR try to avoid these issues seen [here](https://jenkins.nordix.org/view/Airship/job/airship_master_feature_tests_ubuntu/575/consoleFull).


furkatgofurov7 (commented):
/lgtm

PR #743, opened 2021-07-27
Add GO package cleanup to host cleanup script

Rozzii:


host_cleanup.sh
```diff
@@ -68,4 +68,7 @@ fi
 # Clean up any serial logs
 sudo rm -rf /var/log/libvirt/qemu/*serial0.log*
 
+# Clean up Metal3 GO packages
+rm -rf "${HOME}/go/src/github.com/metal3-io"
```
macaptain, line 72: I expect you want to use `$M3PATH` instead of `${HOME}/go`, since this path depends on what `$GOPATH` is.

And I wonder if someone is running metal3-dev-env locally, won't this delete their work-in-progress code?

Thinking on this more, perhaps the best solution is on the `make` side instead of the `make clean` side. It would be far safer if, when we [clone_repos](https://github.com/metal3-io/metal3-dev-env/blob/master/03_launch_mgmt_cluster.sh#L52) in the 03 script, we ensure that the repos are up to date with what's been specified in `BMOBRANCH` and `CAPM3BRANCH` environment variables, instead of removing these directories when they may contain WIP changes.

Rozzii:
FORCE_REPO_UPDATE environment variable is good enough for this purpose so I will close the PR.

PR #746, opened 2021-07-27
Ignore minikube flaky issue

mboukhalfa:
This PR adds a loop to ignore Minikube flaky issue

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -414,13 +414,19 @@ differs(){
 function init_minikube() {
     #If the vm exists, it has already been initialized
     if [[ "$(sudo virsh list --name --all)" != *"minikube"* ]]; then
-      # Restart libvirtd.service as suggested here 
-      # https://github.com/kubernetes/minikube/issues/3566
-      sudo systemctl restart libvirtd.service
-      # Even if it fails to start minikube here, lets ignore the error
-      # It will be retried again in 03 script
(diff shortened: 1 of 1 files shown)


lib/common.sh
```diff
@@ -414,13 +414,19 @@ differs(){
...
     if [[ "$(sudo virsh list --name --all)" != *"minikube"* ]]; then
-      # Restart libvirtd.service as suggested here 
-      # https://github.com/kubernetes/minikube/issues/3566
-      sudo systemctl restart libvirtd.service
-      # Even if it fails to start minikube here, lets ignore the error
-      # It will be retried again in 03 script
-      sudo su -l -c "minikube start --insecure-registry ${REGISTRY}" "$USER" || true
-      sudo su -l -c "minikube stop" "$USER" || true
+      # Loop to ignore minikube issues
+      while /bin/true; do
```
macaptain, line 418: This is an infinite loop if it errors forever. I don't think this will be a problem. The Jenkins build has a timeout on it, and running locally, you'll see that something goes wrong. If we need to put a timeout here, I think we can add it later.

macaptain:
/lgtm

Maybe a better title for the PR is 'Retry on minikube flaky issue', what do you think?

furkatgofurov7 (commented):
Let's watch CI after we merge this and see if we need some more work on this issue, thanks @mboukhalfa for taking care of this. 

/approve

PR #749, opened 2021-07-29
Install BMO after cluster init is completed in the target cluster

ardaguclu:
Baremetal Operator needs resources like certificates come from cluster init.
This PR moves this test's order just after cluster api resources are created in the target cluster.


furkatgofurov7 (commented):
/lgtm

PR #760, opened 2021-08-06
Don't try to delete non-existing httpd container

fmuyassarov:
Don't try to delete httpd container which is merged into the Ironic-api.


namnx228 (approved):
/lgtm

PR #754, opened 2021-07-30
Update feature test notes

furkatgofurov7:
Since we are running feature test with latest API versions of CAPM3/CAPI, this patch updates the notes accordingly.

PR #764, opened 2021-08-12
Parameterize resource API versions

fmuyassarov:
Parameterize resource API versions since we run two different API versions of CAPI and CAPM3 resources.
Based on https://github.com/metal3-io/metal3-dev-env/pull/754


kashifest (approved):

namnx228 (approved):
/lgtm

PR #765, opened 2021-08-16
Don't restore BMO configmap until we deploy it in the target cluster

fmuyassarov:
While deploying BMO in the source cluster, we are modifying IRONIC related env vars that are in [ironic.env](https://github.com/metal3-io/baremetal-operator/blob/master/config/default/ironic.env) file. Right after deploying the BMO, we are restoring changes in [ironic.env](https://github.com/metal3-io/baremetal-operator/blob/master/config/default/ironic.env). 

However, when deploying BMO in the target cluster we are relaying on the default env vars from [config/default/ironic.env](https://github.com/metal3-io/baremetal-operator/blob/master/config/default/ironic.env) which doesn't include corrected env vars. So, basically we are missing the step of env vars modification during the BMO deployment in the target cluster.

There are two ways to fix it:
1. Redo the modifications of env vars just before BMO deployment in target cluster and then restore it again
2. Don't restore the modified ironic.env in 03_* script (during BMO deployment in the source cluster) and do it only after we deploy the BMO in the target cluster. 

For now, I've chosen the option-2 and if you think that option-1 is better then please let me know.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -110,9 +110,6 @@ EOF
   # Deploy. Args: <deploy-BMO> <deploy-Ironic> <deploy-TLS> <deploy-Basic-Auth> <deploy-Keepalived>
   "${BMOPATH}/tools/deploy.sh" true false "${IRONIC_TLS_SETUP}" "${IRONIC_BASIC_AUTH}" true
 
-  # Restore original files
-  mv "${BMOPATH}/config/default/ironic.env.orig" "${BMOPATH}/config/default/ironic.env"
-
   # If BMO should run locally, scale down the deployment and run BMO
   if [ "${BMO_RUN_LOCAL}" == "true" ]; then
     if [ "${IRONIC_TLS_SETUP}" == "true" ]; then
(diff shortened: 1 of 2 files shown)


kashifest (commented):
One question, since `[this](https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/tasks/move.yml#L93)` is using ironic_bmo_configmap from the source cluster , why can't we use the same configmap file for bmo as well?

kashifest (approved):

namnx228:
I have a question regarding to this PR

namnx228:
@fmuyassarov you mentioned that you implement option 1 which is to reapply the modification to the env file, but I don't see this modification. What doing here is just restore the original env file? Are you missing something, or am I understanding something wrongly?

fmuyassarov:
> @fmuyassarov you mentioned that you implement option 1 which reapplies the modification to the env file, but I don't see this modification. What doing here is just restore the original env file? Are you missing something, or am I understanding something wrongly?
> PS: I feel like this PR is implementing the option 2, not the option 1, actually!

Ah sorry, it should be option-2. Fixed in the description

fmuyassarov:
> One question, since [this](https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/tasks/move.yml#L93) is using ironic_bmo_configmap from the source cluster , why can't we use the same configmap file for bmo as well?

As discussed, we can and we probably should. But I will deal with it on a separate patch

namnx228 (approved):
/lgtm

PR #756, opened 2021-08-04
Separate k8s version variable for upgrade test

Sunnatillo:
This PR separates k8s version variables which are used in upgrade test and anywhere else.
It cleans up KUBERNETES_VERSION and UPGRADED_K8S_VERSION variables from  feature_test_vars.sh

PR #766, opened 2021-08-17
Fix node reuse tests

fmuyassarov:
Use one patch version higher of K8S version for `UPGRADED_K8S_VERSION` variable in feature tests. Currently, both `UPGRADED_K8S_VERSION` and `KUBERNETES_VERSION` env vars are set to 1.21.2.

Co-authored-by: Mohammed BOUKHLAFA @mboukhalfa


lib/common.sh
```diff
@@ -200,7 +200,7 @@ export DEFAULT_HOSTS_MEMORY=${DEFAULT_HOSTS_MEMORY:-4096}
 export CLUSTER_NAME=${CLUSTER_NAME:-"test1"}
 export CLUSTER_APIENDPOINT_IP=${CLUSTER_APIENDPOINT_IP:-"192.168.111.249"}
 export KUBERNETES_VERSION=${KUBERNETES_VERSION:-"v1.21.2"}
-export UPGRADED_K8S_VERSION=${UPGRADED_K8S_VERSION:-"v1.21.2"}
+export UPGRADED_K8S_VERSION=${UPGRADED_K8S_VERSION:-"v1.21.3"}
```
mboukhalfa, line 203: I have checked the history of this variables they always have the same default version however I think we had specific exports in  `scripts/feature_tests/feature_test_vars.sh` however they have been removed in the PR #756 
I am not sure if it would be better to restore the old declarations in `feature_test_vars.sh` especially that we have also declarations in `upgrade_test_vars.sh` I mean it is more explicit except if @Sunnatillo has another reason for removing those vars from the file `feature_test_vars.sh`

fmuyassarov: You are right. Those vars were removed from `scripts/feature_tests/feature_test_vars.sh` because we thought all the tests are exporting it separately with their desired value and we missed that node reuse scripts are relaying on them. Since during the feature tests it is only node reuse test that consumes `UPGRADED_K8S_VERSION` var, we could export it within [node_reuse.sh](https://github.com/metal3-io/metal3-dev-env/blob/master/scripts/feature_tests/node_reuse/node_reuse.sh), which will in-turn be inherited by [Ansible vars](https://github.com/metal3-io/metal3-dev-env/blob/780ad2eb12d8238d1b101e5b37dcce4c238e6d3e/vm-setup/roles/v1aX_integration_test/vars/main.yml#L23). As such, there we would not be a need to have this var in lib/common.sh. Because at the end of day, it is Ansible scripts consuming it, and whichever shell script (`node_reuse.sh`, `upgrade.sh` and etc) calls those Ansible playbooks will need to export desired value so that it is inherited by Ansible playbooks.

kashifest (approved):

mboukhalfa:
\LGTM

jan-est:
LGTM

PR #773, opened 2021-08-24
Fix flaky reboot task in remediation

macaptain:
In the remediation tests, if the rebooted node is never spotted going into 'NotReady' by Ansible (i.e. the node recovers quicker than expected), a task loops until the build times out.

https://jenkins.nordix.org/view/Airship/job/airship_master_feature_tests_ubuntu/622/console

This change fixes the flake. Instead of checking for 'NotReady', we confirm that the node was really rebooted by checking the virtual machine goes into shutdown, and then wait for the virtual machine to go into running. This confirms that the machine was rebooted. Then we check that the node has become Ready after the reboot. The reboot tasks have been moved into a separate file with variables matching the names in the similar power_cycle.yaml tasks for consistency.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/power_cycle.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/power_cycle.yml
@@ -15,13 +15,13 @@
     k8s_info:
       api_version: v1
       kind: nodes
-      name: "{{K8S_NODE}}"
+      name: "{{ K8S_NODE }}"
       kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     register: nodes
     retries: 150
     delay: 3
     vars:
-      q: "[? metadata.name == '{{K8S_NODE}}' && status.conditions[? type=='Ready' && status!='True']]"
+      q: "[? metadata.name == '{{ K8S_NODE }}' && status.conditions[? type=='Ready' && status!='True']]"
     until:
       - nodes is succeeded
       - nodes.resources | json_query(q) | length > 0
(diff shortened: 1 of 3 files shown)


mboukhalfa:
Then can we remove also `Wait until powered off "{{ K8S_NODE }}" becomes NotReady` and the pause tasks from `power_cycle.yaml `if the vm check is enough?

I am thinking if we can remove the repeated code to have the `single node reboot` does also the `two master nodes reboot` since it is almost the same steps probably we can use loop list https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#comparing-loop-and-with 
And probably also merging reboot and powercyple with some conditions in distinct tasks `when: reboot` and `when: shutdown`

macaptain:
> Then can we remove also `Wait until powered off "{{ K8S_NODE }}" becomes NotReady` and the pause tasks from `power_cycle.yaml `if the vm check is enough?
> 
> I am thinking if we can remove the repeated code to have the `single node reboot` does also the `two master nodes reboot` since it is almost the same steps probably we can use loop list https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#comparing-loop-and-with
> And probably also merging reboot and powercyple with some conditions in distinct tasks `when: reboot` and `when: shutdown`

There is definitely room for more refactoring here. I think the power cycle one doesn't flake, because we don't power the VM back on until after we've checked the NotReady status. The reboot one can race ahead of the check and break the build.

mboukhalfa (approved):

kashifest (commented):
/lgtm
@fmuyassarov wdyt?

fmuyassarov (approved):
Looks good

PR #775, opened 2021-08-25
Add debug info for check_bm_hosts on failure

macaptain:
My integration test failed on check_bm_hosts without enough information to see what state the bare metal host was in (https://jenkins.nordix.org/job/airship_metal3io_metal3_dev_env_v1a4_integration_test_centos/1000/console). 

Neither the Jenkins log nor the artifact logs contain the full state of the BMH in this failure case. This change adds debug information so that if check_bm_hosts fails, the JSON for the BareMetalHost resource is printed.


macaptain:
Example of build with new output: https://jenkins.nordix.org/job/airship_metal3io_metal3_dev_env_v1a4_integration_test_centos/1003/console

PR #753, opened 2021-07-30
Clean when deleting minikube

mboukhalfa:
This PR adds options `--all --purge ` to `minikube delete` to delete all profiles and the '.minikube' folder. 

This will avoid any potential issue when retrying to create minikube such as : `Message='operation failed: network 'mk-minikube' already exists with uuid c79ec658-7a5d-4453-9836-0dfe2e62188b'` seen [here](https://jenkins.nordix.org/view/Airship/job/airship_master_feature_tests_centos/458/consoleFull).

PR #778, opened 2021-08-25
Fix minikube domain interfaces issue

mboukhalfa:
After merging #753 that adds `--purge` to `minikube delete` this option deletes also the set user's configuration specifically the `driver kvm2`. 
This permits minikube to automatically select the driver. 

Although minikube `automatically selected the kvm2 driver`, it tried `with alternate driver podman` after a failure to `Startup with kvm2 driver`  as we can check in [CI log](https://jenkins.nordix.org/view/Airship/job/airship_master_feature_tests_centos/487/console) this leads to the `error: failed to get domain 'minikube'`.

This PR adds minikube configs after each cleaning and also moves the `init_minikube` function from the common file to 01_prepare_host.sh since it is used only in the that script.


mboukhalfa:
/assign @kashifest 
cc @furkatgofurov7

01_prepare_host.sh
```diff
@@ -188,7 +188,48 @@ function configure_minikube() {
...
+function init_minikube() {
+    #If the vm exists, it has already been initialized
+    if [[ "$(sudo virsh list --name --all)" != *"minikube"* ]]; then
+      # Loop to ignore minikube issues
+      while /bin/true; do
+        minikube_error=0
+        # Restart libvirtd.service as suggested here 
+        # https://github.com/kubernetes/minikube/issues/3566
+        sudo systemctl restart libvirtd.service
+        configure_minikube
```
macaptain, line 204: The `configure_minikube` function is two lines so it would be clearer to copy the lines here. This isn't a blocker for merging at all.

macaptain (approved):
Looks good, nice fix.

PR #779, opened 2021-08-26
✨ Enhance Ansible retry logs

mboukhalfa:
This PR optimizes Ansible k8s_info logs by printing only the first retry details and omit repeating the same output details for every retry.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -2,11 +2,13 @@
 callback_whitelist=ansible.posix.profile_tasks
 stdout_callback=community.general.diy
 
+[callback_profile_tasks]
+task_output_limit = 3
+
 [callback_diy]
 ; Note that the action can be both k8s_info and community.kubernetes.k8s_info
-runner_retry_msg='
-  FAILED: {{ ansible_callback_diy.task.name }}
-  {{ ("FAILED: " + ansible_callback_diy.task.name) | length * "-" }}
+runner_retry_msg='{% if ansible_callback_diy.result.output.attempts == 1 %}
+  FAILED retry: {{ ansible_callback_diy.task.name }}
   {% if ansible_callback_diy.task.action is search("k8s_info") %}
   {% for r in ansible_callback_diy.result.output.resources %}
     kind: {{ r.kind }}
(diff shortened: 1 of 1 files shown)


ansible.cfg
```diff
@@ -4,14 +4,13 @@ stdout_callback=community.general.diy
 
 [callback_diy]
 ; Note that the action can be both k8s_info and community.kubernetes.k8s_info
-runner_retry_msg='
+runner_retry_msg="{% if ansible_callback_diy.result.output.attempts == 1 %}
```
macaptain, line 7: If we reach the retry limit, I'd also like to know what the state was. Can we also print more details if `attempts == retries?`

mboukhalfa: At the end of the retries I expect to end either with OK or FAIL an in both cases I think  we are printing the state. Do you think there is a case when it should pass silently after the number of retries without OK or FAIL?

macaptain: Does it pretty print on failure? Or when it fails, does it print the state as JSON?

mboukhalfa: I think we still have the same failure message as in this example https://jenkins.nordix.org/view/Airship/job/airship_master_feature_tests_centos/490/consoleFull
Probably we can filter the failure message issue to be printed but I think failures need full message as in the example.

macaptain: For me, the YAML is much easier to read. I can see exactly which machine failed to provision in that example, and then I can look for it in the logs. So I think your change is a good improvement because we don't need to print over and over again when nothing has changed. But I would also like the easily readable summary at the end of the retries if something has gone wrong as well.

namnx228, line 7: Do we miss the end double quote?

ansible.cfg
```diff
@@ -4,14 +4,13 @@ stdout_callback=community.general.diy
...
   FAILED: {{ ansible_callback_diy.task.name }}
-  {{ ("FAILED: " + ansible_callback_diy.task.name) | length * "-" }}
-  {% if ansible_callback_diy.task.action is search("k8s_info") %}
+  {% if ansible_callback_diy.task.action is search('k8s_info')%}
```
namnx228, line 9:
```suggestion
  {% if ansible_callback_diy.task.action is search('k8s_info') %}
```

namnx228 (commented):
LGTM, just some sm

ansible.cfg
mboukhalfa: at the end of line 23
```
 RETRYING: {{ ansible_callback_diy.task.name }} {{ ansible_callback_diy.result.output.attempts }}/{{ ansible_callback_diy.task.retries }}"
```

namnx228 (commented):
/lgtm

fmuyassarov:
This is really awesome @mboukhalfa . Thanks
Can you please rebase though? 

/approve

PR #780, opened 2021-08-27
Refactor CP scale down to avoid hanging conns

macaptain:
Swap around control plane scale down commands and waiting to ensure that
we aren't running kubectl or k8s_info against a cluster where the API
servers / control plane may not yet be ready. This should avoid running
into cases where kubectl commands error or k8s_info module hangs after
scaling down the target cluster's control plane.

The untaint commands are refactored so they don't hide unexpected errors.


macaptain:
The key part of this PR is merged in #793. I'll look to refine the `kubectl taint` failure condition in a separate PR.

PR #784, opened 2021-09-06
Add Ironic endpoints checks

fmuyassarov:
This patch adds two basic checks to verify that unauthenticated user can't reach Ironic endpoint.
[`/nodes`](https://docs.openstack.org/api-ref/baremetal/?expanded=list-nodes-detail#list-nodes)

### DIFF
--- a/04_verify.sh
+++ b/04_verify.sh
@@ -298,5 +298,20 @@ for container in ${EXPTD_CONTAINERS}; do
 done
 
 
+IRONIC_NODES_ENDPOINT="${IRONIC_URL}nodes"
+status="$(curl -sk -o /dev/null -I -w "%{http_code}" "${IRONIC_NODES_ENDPOINT}")"
+if [[ $status == 200 ]]; then
+    echo "⚠️  ⚠️  ⚠️   WARNING: Ironic endpoint is exposed for unauthenticated users"
+    exit 1
+elif [[ $status == 401 ]]; then
+    echo "OK - Ironic endpoint is secured"
+else
+    echo "FAIL- got $status from ${IRONIC_NODES_ENDPOINT}, expected 401"
+    echo ""
+    exit "$status"
+fi
+echo ""
+
 echo -e "\nNumber of failures : $FAILS"
 exit "${FAILS}"
+
--- a/vm-setup/roles/v1aX_integration_test/tasks/move.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/move.yml
@@ -194,3 +194,13 @@
     delay: 20
     until: (bmh is succeeded) and
            (bmh.resources | filter_provisioning("provisioned") | length == (NUMBER_OF_BMH | int))
+
+  # Normally as non authenticated user we should
+  # fail here(get 401) to reach Ironic.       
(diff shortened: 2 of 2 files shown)


fmuyassarov:
/hold
I will specifying specific status code

furkatgofurov7 (commented):
LGTM

fmuyassarov:
I added same check to Ansible playbook so that we do the same check on the Ironic that is running on the target cluster.

04_verify.sh
```diff
@@ -297,6 +297,22 @@ for container in ${EXPTD_CONTAINERS}; do
   iterate check_container "$container"
 done
 
+AUTHENTICATED_ENDPOINTS="${IRONIC_URL}nodes ${IRONIC_URL}nodes/detail"
+for endpoint in ${AUTHENTICATED_ENDPOINTS}; do
+    status="$(curl -sk -o /dev/null -I -w "%{http_code}" "$endpoint")"
+    if [ "$status" == 401 ]; then
+        echo "OK - Ironic $endpoint endpoint is secured"
+    elif [[ "$status" == 200 || "$status" == 405 ]]; then
```
dtantsur, line 305: 405 is not really possible for these two

04_verify.sh
```diff
@@ -297,6 +297,22 @@ for container in ${EXPTD_CONTAINERS}; do
   iterate check_container "$container"
 done
 
+AUTHENTICATED_ENDPOINTS="${IRONIC_URL}nodes ${IRONIC_URL}nodes/detail"
```
dtantsur, line 300: Checking nodes is likely enough.

04_verify.sh
```diff
@@ -297,6 +297,22 @@ for container in ${EXPTD_CONTAINERS}; do
...
+    elif [[ "$status" == 200 || "$status" == 405 ]]; then
+        echo "WARNING: Ironic endpoint is exposed for unauthenticated users"
+        exit 1
+    else
+        exit "$status"
+    fi
+done
+process_status $?
+echo ""
+
```
dtantsur, line 314: Let's maybe check that v1 v1/lookup and v1/heartbeat are not authenticated?

04_verify.sh
fmuyassarov: although [doc](https://docs.openstack.org/api-ref/baremetal/?expanded=agent-lookup-detail,agent-heartbeat-detail#agent-heartbeat) says return code could be  200, 400, 404, I'm getting 405
```
[airshipci@fmuyassarov metal3-dev-env]$ curl -k  https://172.22.0.2:6385/v1/heartbeat/
{"error_message": "{\"code\": 405, \"title\": \"Method Not Allowed\", \"description\": \"\"}"}[airshipci@fmuyassarov metal3-dev-env]$ 

```

fmuyassarov:
@dtantsur  I did some local checks, and I decided to keep only `/nodes` path, because  `/heartbeat/` requires POST method, and since we are not checking `/heartbeat/` I removed /lookup too. I think if for any reason heartbeat and lookup are not reachable we will anyways fail to provision, only difference is we will catch the issue little bit late. So, I think `nodes` is enough for now. WDYT?

PR #785, opened 2021-09-08
Fix bug in node reuse where BMHs are not found

macaptain:
The BMHs aren't found when generating templates in node reuse because we're pivoted. To fix this, instead we only generate the templates when provisioning.


kashifest (approved):

furkatgofurov7 (commented):
/lgtm

PR #786, opened 2021-09-08
Add various metrics services to metal3-dev-env

jan-est:
This PR adds [Prometheus](https://prometheus.io/), [node-exporter](https://prometheus.io/docs/guides/node-exporter/#monitoring-linux-host-metrics-with-the-node-exporter) and [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) deployments to metal3-dev-env.

Prometheus data will be fetched and stored after our `feature tests` for debugging purposes.

### DIFF
--- a/vm-setup/monitoring/node-exporter/service.yaml
+++ b/vm-setup/monitoring/node-exporter/service.yaml
@@ -1,0 +1,19 @@
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: node-exporter
+  namespace: monitoring
+  annotations:
+      prometheus.io/scrape: 'true'
+      prometheus.io/port:   '9100'
+spec:
+  selector:
+      app.kubernetes.io/component: exporter
+      app.kubernetes.io/name: node-exporter
+  ports:
+  - name: node-exporter
+    protocol: TCP
+    port: 9100
+    targetPort: 9100
+   
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -494,6 +494,17 @@ if [ "${EPHEMERAL_CLUSTER}" != "tilt" ]; then
     launch_ironic
 fi
 
+if [[ "${PROMETHEUS_MONITORING}" == true ]]; then
+  kubectl create namespace monitoring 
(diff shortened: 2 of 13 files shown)


vm-setup/monitoring/node-exporter/service.yaml
```diff
@@ -0,0 +1,18 @@
...
+      prometheus.io/port:   '9100'
+spec:
+  selector:
+      app.kubernetes.io/component: exporter
+      app.kubernetes.io/name: node-exporter
+  ports:
+  - name: node-exporter
+    protocol: TCP
+    port: 9100
+    targetPort: 9100
```
Rozzii, line 18: This is missing the new line on EOF

vm-setup/monitoring/prometheus/monitoring-namespace.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Namespace
+apiVersion: v1
+metadata:
+  name: monitoring
+  labels:
+    name: monitoring
```
Rozzii, line 6: New line is missing at EOF

Rozzii (changes requested):
I think other than the missing newlines it looks good.

vm-setup/monitoring/prometheus/monitoring-namespace.yaml
jan-est: I did remove the whole file, since we are deploying namespace [here](https://github.com/metal3-io/metal3-dev-env/blob/cb68a92ee148ded537727cec9a207c1064b5f962/03_launch_mgmt_cluster.sh#L501)

Rozzii (approved):

furkatgofurov7:
I expect, we should be able to see Prometheus metrics after running feature tests on this PR

/test-features

furkatgofurov7:
@macaptain do we need this PR anymore, I think we dropped supporting it in other repositories, isn't it?

macaptain:
> @macaptain do we need this PR anymore, I think we dropped supporting it in other repositories, isn't it?

Agreed, let's close this PR.

/close

PR #787, opened 2021-09-09
BugFix: Download upgraded image and add variables to M3M template.

smoshiur1237:
There is a bug in upgrade workflow where upgraded version of image needs to be downloaded from artifactory and pass the upgraded variables in Metal3MachineTemplate.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/download_image.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/download_image.yml
@@ -18,7 +18,7 @@
 
           - name: Verify specific centos image containing newer version of cloud-init is downloaded
             get_url:
-              url: "{{ IMAGE_LOCATION_CENTOS }}"
+              url: "{{ IMAGE_LOCATION_CENTOS }}/{{ IMAGE_NAME_CENTOS }}"
               dest: "{{ IRONIC_IMAGE_DIR }}/{{ IMAGE_NAME_CENTOS }}"
               mode: 0664
 
@@ -63,7 +63,7 @@
 
           - name: Verify specific Ubuntu image containing newer version of cloud-init is downloaded
             get_url:
-              url: "{{ IMAGE_LOCATION }}"
+              url: "{{ IMAGE_LOCATION }}/{{ IMAGE_NAME }}"
               dest: "{{ IRONIC_IMAGE_DIR }}/{{ IMAGE_NAME }}"
               mode: 0664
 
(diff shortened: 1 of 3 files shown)


vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -441,6 +450,9 @@
       CLUSTER_UID: "{{ clusters.resources[0].metadata.uid }}"
       M3MT_NAME: "{{CLUSTER_NAME}}-new-controlplane-image"
       DATA_TEMPLATE_NAME: "{{CLUSTER_NAME}}-controlplane-template"
+      RAW_IMAGE_NAME: "UBUNTU_20.04_NODE_IMAGE_K8S_{{UPGRADED_K8S_VERSION}}-raw.img"
+      IMAGE_URL: "http://172.22.0.1/images/{{ RAW_IMAGE_NAME }}"
+      IMAGE_CHECKSUM: "http://172.22.0.1/images/{{ RAW_IMAGE_NAME }}.{{ IMAGE_CHECKSUM_TYPE }}sum"
```
kashifest, line 455: Shouldn't this have the same set of vars as above?

smoshiur1237: Other two variables are not needed in  template file. Those two are necessary during downloading the image.

vm-setup/roles/v1aX_integration_test/tasks/download_image.yml
```diff
@@ -18,7 +18,7 @@
 
           - name: Verify specific centos image containing newer version of cloud-init is downloaded
             get_url:
-              url: "{{ IMAGE_LOCATION_CENTOS }}"
+              url: "{{ IMAGE_LOCATION_CENTOS }}/{{ IMAGE_NAME_CENTOS }}"
```
kashifest, line 21: I think this `IMAGE_LOCATION_CENTOS `was also wrong @smoshiur1237 , I think we should use `IMAGE_LOCATION` and `IMAGE_NAME` only here as well. @fmuyassarov can you confirm?

vm-setup/roles/v1aX_integration_test/tasks/download_image.yml
```diff
@@ -18,7 +18,7 @@
...
               dest: "{{ IRONIC_IMAGE_DIR }}/{{ IMAGE_NAME_CENTOS }}"
```
kashifest, line 22: If you change that @smoshiur1237 , would you also change all the occurrences of  `IMAGE_NAME_CENTOS` like this

kashifest (commented):
Actually the whole download_image needs refactoring, I think we don't need separate tasks for centos and ubuntu  and its also limiting download of other images defined in lib/common.sh. I ll open a refactor PR for it. Leave it as it is now @smoshiur1237

vm-setup/roles/v1aX_integration_test/tasks/download_image.yml
smoshiur1237: Yes the download file needs optimization. It can have the same name for both OS.

kashifest (approved):

PR #788, opened 2021-09-09
chdir into BMOPATH when using deploy.sh

honza:
This enables us to use the provided Makefile.

I'm hoping this will fix the integration test issues in https://github.com/metal3-io/baremetal-operator/pull/968


furkatgofurov7 (commented):
/lgtm

PR #789, opened 2021-09-13
Add ansible verbosity to default level

kashifest:
This PR reinstates ansible verbosity to default level for tasks which are run through `v1aX_integration_test.yml` and cleanup task


smoshiur1237:
Looks good to me.

PR #793, opened 2021-09-14
Fix node reuse test with upgraded image

smoshiur1237:
The node reuse test is failing during the upgrade of the new kubernetes version  because of image upgrading is necessary to have a proper upgrade. Otherwise that leads to us cloud-init failure because of binary conflict. This PR will add the image upgrading part and will solve the issue.


vm-setup/roles/v1aX_integration_test/templates/Metal3MachineTemplate.yml
```diff
@@ -9,6 +9,7 @@ metadata:
     name: {{ CLUSTER_NAME }}
     uid: {{ CLUSTER_UID }}
 spec:
+  nodeReuse: true
```
jan-est, line 12: Could we parameterize this and set value as "true" on  node_reuse.sh? We could have NODE_REUSE_VAL set to false by default?

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -61,6 +118,9 @@
       definition:
         spec:
           version: "{{ UPGRADED_K8S_VERSION }}"
+          machineTemplate:
+            infrastructureRef:
+              name: "{{CLUSTER_NAME}}-new-controlplane-image"
           rolloutStrategy:
             rollingUpdate:
               maxSurge: 0
```
jan-est, line 126: Do we need to hard code maxSurge value here? We have MAX_SURGE_VALUE environment variable to set this value when creating the controlplane. https://github.com/metal3-io/metal3-dev-env/blob/3423a5039f4cb939c3a8d5cf2d32b1b194172cd0/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane.yaml#L9

vm-setup/roles/v1aX_integration_test/templates/Metal3MachineTemplate.yml
smoshiur1237: Good point, will do that in separate PR. Now my priority is to fix the node reuse test and then will refactor other things.

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
smoshiur1237: Will do it separate refactoring PR.

smoshiur1237:
/test-integration
test-v1a4-centos-integration

scripts/feature_tests/feature_test_vars.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
 
 export NUM_OF_MASTER_REPLICAS="3"
-export NUM_OF_WORKER_REPLICAS="1"
\ No newline at end of file
+export NUM_OF_WORKER_REPLICAS="1"
+export KUBERNETES_VERSION="v1.21.2"
```
macaptain, line 5: Does this mean that v1.21.2 becomes the Kubernetes version for all feature tests (instead of 1.22.0)?

kashifest: Yes , thats gonna change pretty soon, once this PR lands and  after a few days of CI sanity I ll do changes for 1.22.0 -> 1.22.1

scripts/feature_tests/node_reuse/node_reuse.sh
```diff
@@ -12,4 +12,11 @@ export FROM_K8S_VERSION="v1.21.2"
 export KUBERNETES_VERSION=${FROM_K8S_VERSION}
 export UPGRADED_K8S_VERSION="v1.22.0"
 
+if [[ "${IMAGE_OS}" == "Ubuntu" ]]; then
+  export UPGRADED_IMAGE_NAME="UBUNTU_20.04_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}.qcow2"
+  export UPGRADED_RAW_IMAGE_NAME="UBUNTU_20.04_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}-raw.img"
+else
+  export UPGRADED_IMAGE_NAME="CENTOS_8_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}.qcow2"
+  export UPGRADED_RAW_IMAGE_NAME="CENTOS_8_NODE_IMAGE_K8S_${UPGRADED_K8S_VERSION}-raw.img"
+fi
```
kashifest, line 21: Aren't we using same vars for upgrade? Shall we keep it somewhere in lib/images?

smoshiur1237: No, I didn't, as we were testing only in ubuntu. I have created this var only for node reuse tests.

PR #794, opened 2021-09-15
WIP:Add upgraded kubernetes version to v1.22.1 for node reuse.

smoshiur1237:


smoshiur1237:
The test is out of scope in current workflow and is not necessary. So this PR can be closed. It will be taken care with the following [PR](https://github.com/metal3-io/metal3-dev-env/pull/793)

PR #800, opened 2021-09-22
Pull container images only if missing

macaptain:
In the 01 (setup) script, try to pull images only if the image is not already found.

This should help us avoid hitting the DockerHub API rate limit.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -87,7 +87,7 @@ else
       curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/"${KIND_VERSION}"/kind-"$(uname)"-amd64
       chmod +x ./kind
       sudo mv kind /usr/local/bin/.
-      sudo "${CONTAINER_RUNTIME}" pull kindest/node:"${KIND_NODE_IMAGE_VERSION}"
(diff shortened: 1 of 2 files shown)


kashifest (approved):

01_prepare_host.sh
```diff
@@ -170,10 +170,20 @@ if [ ! -f "${IMAGE_NAME}" ] ; then
...
+# If a given container with tag doesn't exist locally, pull it. 
+# Otherwise, do nothing.
+# Helps conserve number of API calls to DockerHub to avoid hitting rate limit.
+function container_image_pull_if_missing() {
+  local IMAGE="$1"
+  if ! sudo "${CONTAINER_RUNTIME}" | grep -q "$IMAGE"; then
+    sudo "${CONTAINER_RUNTIME}" pull "$IMAGE"
+  fi
+}
+
```
kashifest, line 182: I am thinking shall we keep this function in lib/common.sh, if other scripts might need it? Not sure if they need it though.

macaptain: I've moved the function to lib/common.sh.

smoshiur1237: Pulling image is used only in 01 script.

PR #801, opened 2021-09-22
Bugfix pull image if missing bash function

macaptain:
A bug in the bash function for pulling an image only if it's missing meant that existing images weren't found. This should solve the problem.

I've also renamed the function so it starts with a verb.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -424,9 +424,9 @@ differs(){
 # Inputs:
 # - Full name of a Docker/podman/crictl image including tag
 #
-function container_image_pull_if_missing() {
+function pull_container_image_if_missing() {
   local IMAGE="$1"
-  if ! sudo "${CONTAINER_RUNTIME}" | grep -q "$IMAGE"; then
+  if [[ -z $(sudo "${CONTAINER_RUNTIME}" image ls "$IMAGE" | tail -n +2) ]]; then
     sudo "${CONTAINER_RUNTIME}" pull "$IMAGE"
   fi
 }
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -87,7 +87,7 @@ else
       curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/"${KIND_VERSION}"/kind-"$(uname)"-amd64
       chmod +x ./kind
       sudo mv kind /usr/local/bin/.
-      container_image_pull_if_missing kindest/node:"${KIND_NODE_IMAGE_VERSION}"
+      pull_container_image_if_missing kindest/node:"${KIND_NODE_IMAGE_VERSION}"
   fi
   if [ "${EPHEMERAL_CLUSTER}" == "tilt" ]; then
     curl -fsSL https://raw.githubusercontent.com/tilt-dev/tilt/master/scripts/install.sh | bash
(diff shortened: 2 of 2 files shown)


macaptain:
I've tested this is working now:

```
$ sudo docker images | grep kindest/node
kindest/node                                                 v1.22.0   8bc140d2ba34   6 weeks ago    933MB
kindest/node                                                 v1.21.2   bd88cf3babd7   2 months ago   928MB
$ pull_container_image_if_missing kindest/node:v1.21.2
$ pull_container_image_if_missing kindest/node:v1.21.3
Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

lib/common.sh
```diff
@@ -424,9 +424,9 @@ differs(){
 # Inputs:
 # - Full name of a Docker/podman/crictl image including tag
 #
-function container_image_pull_if_missing() {
+function pull_container_image_if_missing() {
   local IMAGE="$1"
-  if ! sudo "${CONTAINER_RUNTIME}" | grep -q "$IMAGE"; then
+  if [[ -z $(sudo "${CONTAINER_RUNTIME}" image ls "$IMAGE" | tail +2) ]]; then
```
Xenwar, line 429:
```suggestion
  if [[ -z $(sudo "${CONTAINER_RUNTIME}" image ls "$IMAGE" -q) ]]; then
```

macaptain: `-q` lists the image IDs:

```
$ sudo docker image ls -q                                                                                       
6d49f8ffbce9                                                                                                                                                                                                                                                        
182c83f40238                                                                                                                                                                                                                                                        
ff66cc3c4d73                                                                                                                            
...             
```

It will be hard to match against these.

The purpose of the `tail +2` command is to skip the header:

```
$ sudo docker image ls kindest/node
REPOSITORY     TAG       IMAGE ID       CREATED        SIZE
kindest/node   v1.22.0   8bc140d2ba34   6 weeks ago    933MB
kindest/node   v1.21.2   bd88cf3babd7   2 months ago   928MB
```
```
$ sudo docker image ls kindest/node | tail +2
kindest/node   v1.22.0   8bc140d2ba34   6 weeks ago    933MB
kindest/node   v1.21.2   bd88cf3babd7   2 months ago   928MB
```

The version of `podman` in our environment has `-n` option which skips the header, but our version of `docker` doesn't have this, so I have to use the tail command.

macaptain: This could work! I think it is less obvious that the purpose of the `-q` is to skip the header though. Maybe the `| tail +2` is clearer?

macaptain: `podman` has `-n` which skips the header, but our version of Docker doesn't have this sadly.

Xenwar: oh, I though `-q` would work for all container run time. Please do disregard my comment.

macaptain: `-q` will work for all the runtimes - the purpose is to show image IDs and it happens to skip the header, so it would work, I just wonder if it's clearer with the tail command that we want to skip the header.

Xenwar: I get it `-q` does not add any value in this case. In fact, hides some useful details.

kashifest:
v1a4 is hitting the docker pull rate limit bug again

fmuyassarov:
It could be that something is different in podman, but I'm curious to try once more
/test-v1a4-centos-integration

macaptain:
These Docker image are available in a CentOS image:
```
$ sudo podman image ls
REPOSITORY                              TAG         IMAGE ID      CREATED      SIZE
docker.io/library/centos                centos8     5d0da3dc9764  8 days ago   239 MB
registry.hub.docker.com/library/golang  1.16        828b0508ae64  13 days ago  941 MB
docker.io/library/registry              latest      b2cb11db9d3d  3 weeks ago  26.8 MB
```

PR #803, opened 2021-09-22
Add Docker registry to Renovate

macaptain:
Pin the Docker registry image to a specific version and upgrade it with Renovate.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -157,7 +157,7 @@ export NODE_DRAIN_TIMEOUT=${NODE_DRAIN_TIMEOUT:-"0s"}
 export MAX_SURGE_VALUE="${MAX_SURGE_VALUE:-"1"}"
 
 # Docker registry for local images
-export DOCKER_REGISTRY_IMAGE=${DOCKER_REGISTRY_IMAGE:-"docker.io/registry:latest"}
+export DOCKER_REGISTRY_IMAGE=${DOCKER_REGISTRY_IMAGE:-"registry:2.7.1"}
 
 # VBMC and Redfish images
 export VBMC_IMAGE=${VBMC_IMAGE:-"quay.io/metal3-io/vbmc"}
--- a/renovate.json
+++ b/renovate.json
@@ -16,7 +16,13 @@
       "packageNames": ["kindest/node"],
       "automerge": false,
       "commitMessageTopic": "kindest/node"
-    },    
+    },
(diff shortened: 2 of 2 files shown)


namnx228 (approved):
/lgtm

lib/common.sh
```diff
@@ -157,7 +157,7 @@ export NODE_DRAIN_TIMEOUT=${NODE_DRAIN_TIMEOUT:-"0s"}
 export MAX_SURGE_VALUE="${MAX_SURGE_VALUE:-"1"}"
 
 # Docker registry for local images
-export DOCKER_REGISTRY_IMAGE=${DOCKER_REGISTRY_IMAGE:-"docker.io/registry:latest"}
+export DOCKER_REGISTRY_IMAGE=${DOCKER_REGISTRY_IMAGE:-"registry:2.7.1"}
```
fmuyassarov, line 160: does it mean that we need to keep this version in airship-dev-tools too since we are prepulling it in there?

macaptain: I thought the 01_ script got run in the image building process, so this would cover it, but it seems there is more to it than that. I'll take a look. (https://github.com/Nordix/airship-dev-tools/blob/master/ci/scripts/image_scripts/source_cluster_container_images.sh)

fmuyassarov: although we run `01_script` in airship-dev-tools (only on Ubuntu for now, since Centos has a bug we spotted recently) the pulling of the registry image happens in 02 script https://github.com/metal3-io/metal3-dev-env/blob/445c19f4d8163a27d568674e36fedead73261d7e/02_configure_host.sh#L133. And we don't run 02 script during image building. For that reason, we added pulling command in airship-dev-tools for some of the images including registry.

macaptain: I think the pulling of the image will also happen in `01`, for the reason that there is an environment variable for the Registry image with name ending in 'IMAGE':
https://github.com/metal3-io/metal3-dev-env/blob/0bdd8ddce13b699ff5d11728c6b20f6c14a8ae70/lib/common.sh#L160

https://github.com/metal3-io/metal3-dev-env/blob/0bdd8ddce13b699ff5d11728c6b20f6c14a8ae70/01_prepare_host.sh#L174

So we should be fine.

fmuyassarov: Ah you are right.

PR #804, opened 2021-09-23
Use kubectl instead of k8s module for applying

macaptain:
The k8s module does no validation on the yaml when applying! It's like running kubectl with --validate=false. Furthermore, even adding the validate block to the module won't validate custom resources. This is quite unexpected, and meant that we were applying invalid template (nodeDrainTimeout was skipped from our definitions.)

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/templates/cluster-template-workers.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/cluster-template-workers.yaml
@@ -19,12 +19,6 @@ spec:
         cluster.x-k8s.io/cluster-name: ${ CLUSTER_NAME }
         nodepool: nodepool-0
     spec:
-{% if CAPI_VERSION == "v1alpha3" %}
-      nodeDrainTimeout: ${ NODE_DRAIN_TIMEOUT }
-{% else %}
-      machineTemplate:
-        nodeDrainTimeout: ${ NODE_DRAIN_TIMEOUT }
-{% endif %}
       clusterName: ${ CLUSTER_NAME }
       version: ${ KUBERNETES_VERSION }
       bootstrap:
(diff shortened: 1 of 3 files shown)


macaptain:
Now catching any invalid templates: https://jenkins.nordix.org/job/airship_metal3io_metal3_dev_env_v1a5_integration_test_ubuntu/444/console

And together with the template fix.

vm-setup/roles/v1aX_integration_test/templates/cluster-template-workers.yaml
```diff
@@ -19,12 +19,6 @@ spec:
         cluster.x-k8s.io/cluster-name: ${ CLUSTER_NAME }
         nodepool: nodepool-0
     spec:
-{% if CAPI_VERSION == "v1alpha3" %}
-      nodeDrainTimeout: ${ NODE_DRAIN_TIMEOUT }
```
fmuyassarov, line 23: 👍

vm-setup/roles/v1aX_integration_test/vars/main.yml
```diff
@@ -108,6 +108,7 @@ provision_actions:
     - "ci_test_provision"
     - "provision_cluster"
     - "provision_controlplane"
+    - "provision_worker"
```
fmuyassarov, line 111: why do we need it here?

macaptain: Otherwise the templates don't get updated when we provision workers. It's an omission from a previous change.

macaptain: https://github.com/metal3-io/metal3-dev-env/pull/785/files missed one of the provisioning actions for when we should generate templates.

PR #808, opened 2021-09-28
Use shell instead of k8s_info when KCP is unstable

macaptain:
In the remediation tests, there are two tasks where the control plane is scaling down or having machines powered off. It's possible for the k8s_info module to hang on these places, so replacing them with shell will fix a flake where the build times out.


kashifest (approved):

smoshiur1237 (approved):
/lgtm

PR #809, opened 2021-09-28
Add a bash script that only generates the cluster templates

namnx228:
This PR adds a script to generate the cluster template only. It can be used by the CAPM3 e2e test framework


elfosardo (approved):

kashifest (approved):

smoshiur1237 (approved):
Looks good to me

PR #810, opened 2021-09-28
Handle 'available' state

zaneb:
We are in the (long) process of renaming 'ready' to 'available'. We need
to be able to handle seeing either state.


fmuyassarov (approved):

furkatgofurov7 (commented):
This is touching node reuse, we have to run feature tests as well on this PR. 
/test-features

PR #812, opened 2021-09-29
Do not verify TLS in sushy-tools, real hardware doesn't do it

dtantsur:
It is only this year when the Redfish standard added a feature to enable
host verification for virtual media. Until we support it, disabling
the host validation is the right thing to do.

### DIFF
--- a/vm-setup/roles/virtbmc/defaults/main.yml
+++ b/vm-setup/roles/virtbmc/defaults/main.yml
@@ -1,7 +1,8 @@
 # Can be set to "teardown" to destroy a previous configuration
 virtbmc_action: setup
 sushy_ignore_boot_device: False
-sushy_vmedia_verify_ssl: True
+# This matches how real hardware behaves
+sushy_vmedia_verify_ssl: False
 
 # BMC credentials
 vbmc_username: "admin"


elfosardo (approved):

vm-setup/roles/virtbmc/defaults/main.yml
```diff
@@ -1,7 +1,8 @@
 # Can be set to "teardown" to destroy a previous configuration
 virtbmc_action: setup
 sushy_ignore_boot_device: False
-sushy_vmedia_verify_ssl: True
+# This matches how real hardware behaves
+sushy_vmedia_verify_ssl: False
```
fmuyassarov, line 5: I can see that `SUSHY_EMULATOR_VMEDIA_VERIFY_SSL` is now [defaulting](https://github.com/openstack/sushy-tools/blob/569a02c1ccca1a6824fc0b9542ac2966bcaf997a/doc/source/admin/emulator.conf#L100 ) to `False`  to _match the actual bare metal hardware_, but I'm just curious why real hardware would not do host verification before pulling the image?

fmuyassarov (commented):
/approve

vm-setup/roles/virtbmc/defaults/main.yml
dtantsur: Because until very recently there was no way to provide custom certificates. And without custom certificates this feature cannot be reasonably used.

smoshiur1237 (approved):
/lgtm

PR #814, opened 2021-09-30
Fix minikube download for v1a4

kashifest:
For v1a4 tests minikube version is 1.22.0 but since we pre-run 01 with default values in the image, we have different version than expected for v1a4 tests. This commit is trying to fix that issue.


elfosardo (approved):

smoshiur1237 (approved):
/lgtm

PR #819, opened 2021-10-04
simplify bmh state checking conditions

mboukhalfa:
This PR simplify the generic conditions used to check the BMH state since we can have only one BMH deprovisioning / provisioning at a time.


vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -175,9 +175,8 @@
     delay: 2
     until:
       - bmhs is succeeded
-      - (saved_bmhs.resources | filter_provisioning("deprovisioning") | map(attribute='metadata.name')) |
-        intersect(bmhs.resources | filter_provisioning("available,ready") | map(attribute='metadata.name')) |
-        length ==  (NUMBER_OF_BMH_MD | int)
+      - saved_bmhs.resources | filter_provisioning("deprovisioning") | json_query('metadata.name') == bmhs.resources | 
+        filter_provisioning("available,ready") | json_query('metadata.name')
```
furkatgofurov7, line 179: Can you please update the name of the task too because it is misleading, in fact it is always only one BMH not BMH**s** that is deprovisioning.

furkatgofurov7 (changes requested):

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -175,11 +175,10 @@
     delay: 2
     until:
       - bmhs is succeeded
-      - (saved_bmhs.resources | filter_provisioning("deprovisioning") | map(attribute='metadata.name')) |
-        intersect(bmhs.resources | filter_provisioning("available,ready") | map(attribute='metadata.name')) |
-        length ==  (NUMBER_OF_BMH_MD | int)
```
furkatgofurov7, line 180: why are we omitting `length` check btw?

furkatgofurov7: You are right, I overlooked the previous step, so we will always have only and one BMH in  `saved_bmhs` from the previous step which will be deprovisioning and becoming ready and registered in that var, and picking a first node from that ends up being the same BMH again, alright, fine with the logic above.

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -190,9 +190,9 @@
     delay: 2
     until:
       - bmhs is succeeded
-      - (saved_bmhs.resources | filter_provisioning("deprovisioning") | map(attribute='metadata.name')) |
-        intersect(bmhs.resources | filter_provisioning("provisioning") | map(attribute='metadata.name')) |
-        length == (NUMBER_OF_BMH_MD | int)
+      - bmhs.resources | filter_provisioning("provisioning") | length == 1
```
furkatgofurov7, line 193: not a blocker, better to parameterize the code base rather than hardcoding.

smoshiur1237 (approved):
/lgtm

PR #820, opened 2021-10-04
WIP: Set static IP variables for ironic nodes

imain:
We need to be able to set static IP variables so we can configure
the network data for each host.

Co-Authored-By: Caleb Boylan <cboylan@redhat.com>


imain:
We piggy backed on the DHCP variables here.  The other option would be to create a set of variables around static IPs, and then have another vm_setup_vars.yaml to configure static IPs on a network.  Requires more plumbing but certainly doable.

imain:
https://github.com/openshift-metal3/dev-scripts/pull/1297

hardys:
Looks like we're going with https://github.com/metal3-io/metal3-dev-env/pull/879 instead

PR #834, opened 2021-10-17
Uplift kubernetes version to 1.22.2

kashifest:
upgrade checks 1.22.1->1.22.2


vm-setup/roles/v1aX_integration_test/vars/main.yml
```diff
@@ -19,8 +19,8 @@ NODE_DRAIN_TIMEOUT: "{{ lookup('env', 'NODE_DRAIN_TIMEOUT') | default('0s', true
 BMOPATH: "{{ lookup('env', 'BMOPATH') }}"
 IRONIC_DATA_DIR: "{{ lookup('env', 'IRONIC_DATA_DIR') }}"
 KUBECONFIG_PATH: "/home/ubuntu/.kube/config"
-KUBERNETES_VERSION: "{{ lookup('env', 'KUBERNETES_VERSION') | default('v1.21.2', true) }}"
-UPGRADED_K8S_VERSION: "{{ lookup('env', 'UPGRADED_K8S_VERSION') | default('v1.22.0', true) }}"
+KUBERNETES_VERSION: "{{ lookup('env', 'KUBERNETES_VERSION') | default('v1.22.2', true) }}"
+UPGRADED_K8S_VERSION: "{{ lookup('env', 'UPGRADED_K8S_VERSION') | default('v1.22.2', true) }}"
```
mboukhalfa, line 23:
```suggestion
KUBERNETES_VERSION: "{{ lookup('env', 'KUBERNETES_VERSION') | default('v1.22.1', true) }}"
UPGRADED_K8S_VERSION: "{{ lookup('env', 'UPGRADED_K8S_VERSION') | default('v1.22.2', true) }}"
```
I think we should upgrade from an old to a newer version ( `UPGRADED_K8S_VERSION` > `KUBERNETES_VERSION` )

config_example.sh
```diff
@@ -121,8 +121,8 @@
 #
 #export CAPM3_VERSION=v1alpha5
 
-#export KUBERNETES_VERSION="v1.22.0"
-#export UPGRADED_K8S_VERSION="v1.22.0"
+#export KUBERNETES_VERSION="v1.22.2"
```
mboukhalfa, line 124:
```suggestion
#export KUBERNETES_VERSION="v1.22.1"
```

kashifest:
@mboukhalfa normally we run v1a4 and v1a5 jobs with the highest k8s version which in this case should be 1.22.2. For feature test and upgrade test, we install the source cluster with one patch version less i.e. 1.22.1 and then upgrade to latest which is 1.22.2. Feature test and upgrade test have their own vars file where these values should be set. All the other places we should default to highest IMO.

vars.md
```diff
@@ -48,7 +48,7 @@ assured that they are persisted.
 | POD_CIDR | Pod CIDR | "x.x.x.x/x" | "192.168.0.0/18" |
 | NODE_HOSTNAME_FORMAT | Node hostname format. This is a format string that must contain exactly one %d format field that will be replaced with an integer representing the number of the node. | "node-%d" |
 | KUBERNETES_VERSION | Kubernetes version | "x.x.x" | "1.21.0" |
```
furkatgofurov7, line 50: this needs an update (v1.22.2)

PR #840, opened 2021-10-27
Clarify BMC_DRIVER usage to include redfish-virtualmedia

hardys:
redfish-virtualmedia is also a valid option for this variable


fmuyassarov (commented):
/lgtm

@hardys Prow is down, once it is up again, I will give lgtm to it once more.

PR #841, opened 2021-10-27
Make BOOT_MODE configurable

hardys:
Currently this is hard-coded to default to uefi for ipv6, and bios for
ipv4, but there are times when you may wish to override these defaults

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -241,7 +241,7 @@ function make_bm_hosts() {
       -password "$password" \
       -user "$user" \
       -boot-mac "$mac" \
-      -boot-mode "legacy" \
+      -boot-mode "${BOOT_MODE}" \
       "$name"
   done
 }
--- a/config_example.sh
+++ b/config_example.sh
@@ -115,6 +115,12 @@
 #
 #export BMC_DRIVER="mixed"
 
+#
+# Set libvirt firmware and BMC bootMode
+# Choose "legacy" (bios), "UEFI", or "UEFISecureBoot"
+# Defaults to legacy for ipv4, UEFI for ipv6
+# export BOOT_MODE="UEFI"
+
 # Select the Cluster API provider Metal3 version
 # Accepted values : v1alpha4, v1alpha5, v1beta1
 # default: v1alpha4
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -22,6 +22,7 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -e "virthost=$HOSTNAME" \
     -e "platform=$NODES_PLATFORM" \
     -e "libvirt_firmware=$LIBVIRT_FIRMWARE" \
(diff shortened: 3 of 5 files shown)


03_launch_mgmt_cluster.sh
```diff
@@ -234,13 +234,18 @@ EOF
 # Create the BMH CRs
 #
 function make_bm_hosts() {
+  if [ "${LIBVIRT_FIRMWARE}" = "bios" ]; then
+    BOOT_MODE="legacy"
+  else
+    BOOT_MODE="UEFI"
+  fi
```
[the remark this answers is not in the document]
hardys: @fmuyassarov good point, we'd need to also set `libvirt_secure_boot` ref https://github.com/metal3-io/metal3-dev-env/commit/1241e2954026068fd2e88232178e64e0b8c0c317

But perhaps it would be a more intuitive user experience if we just exposed `BOOT_MODE` so that the value aligns with the BMH `bootMode` API, then automatically select the required firmware based on that setting.

We might also want to consider making UEFI the default so that it aligns with the BMH defaults?

hardys:
@fmuyassarov I adjusted this to instead expose BOOT_MODE, and also added support for UEFISecureBoot

What do you think about switching the default to UEFI so this aligns with the BMO defaults? (and upstream Ironic I think /cc @dtantsur)

IIRC the reason we enable UEFI for ipv6 is because we needed that for ipv6 PXE boot to work, but it probably doesn't make sense long term to default to legacy BIOS boot for ipv4, both could just use UEFI?

fmuyassarov:
> IIRC the reason we enable UEFI for ipv6 is because we needed that for ipv6 PXE boot to work, but it probably doesn't make sense long term to default to legacy BIOS boot for ipv4, both could just use UEFI?

I don't have strong opinion on this to be honest, not sure if there is some other need to default it to legacy. So, I'm +1 for defaulting to UEFI for in IPv4 as well.
/approve

hardys:
> > IIRC the reason we enable UEFI for ipv6 is because we needed that for ipv6 PXE boot to work, but it probably doesn't make sense long term to default to legacy BIOS boot for ipv4, both could just use UEFI?
> 
> I don't have strong opinion on this to be honest, not sure if there is some other need to default it to legacy. So, I'm +1 for defaulting to UEFI in IPv4 as well. /approve

Ok thanks, switching the default should probably be a separate PR, so I can propose that as a follow-up :)

config_example.sh
```diff
@@ -115,6 +115,12 @@
 #
 #export BMC_DRIVER="mixed"
 
+#
+# Set libvirt firmware and BMC bootMode
+# Choose "legacy" (bios), "UEFI", or "UEFISecureBoot"
+# Defaults to legacy for ipv4, UEFI for ipv6
+# export BOOT_MODE="UEFI"
```
furkatgofurov7, line 122: This should be defaulted to "legacy" based on what we have set as a default for this var in vars.md or the other way around?

Otherwise lgtm.

kashifest: I think it defaults to legacy already for ipv4 as line 121 states @furkatgofurov7 ?
/approve

PR #843, opened 2021-10-29
Remove undesired bash vars and options

mboukhalfa:
Refactor `KUBERNETES_VERSION` to avoid renovate conflicts and remove `set -u` that unset vars while sourced before the make.


fmuyassarov (approved):

macaptain (changes requested):
`set -u` is a safety feature which ensures that all the Bash variables we use are defined. I think instead of removing it, we should ensure all the environment variables we need are defined and sourced somewhere in the environment.

hardys:
> `set -u` is a safety feature which ensures that all the Bash variables we use are defined. I think instead of removing it, we should ensure all the environment variables we need are defined and sourced somewhere in the environment.

I agree, it'd be good to understand more about why it was necessary to remove it - can we just default whatever is undefined instead?

mboukhalfa:
> > `set -u` is a safety feature which ensures that all the Bash variables we use are defined. I think instead of removing it, we should ensure all the environment variables we need are defined and sourced somewhere in the environment.
> 
> I agree, it'd be good to understand more about why it was necessary to remove it - can we just default whatever is undefined instead?

Trying to source  `scripts/feature_tests/upgrade/upgrade_vars.s` in common script. However as common repeatedly sourced in different places I suppose that many vars will be undefined e.g the first time [common sourced](https://github.com/metal3-io/metal3-dev-env/blob/7887985267b257e756dad3f4e2d9470d232239e9/01_prepare_host.sh#L7) in script 01 GOPATH  will be undefined since Go not installed yet

furkatgofurov7:
@mboukhalfa do you think this PR is still needed, if so please rebase and re-run the tests otherwise we could close it. Thanks!

PR #845, opened 2021-10-29
Add quay.io as the registry

kashifest:
Because of recent change regarding container registry parameterization a bug  was introduced in metal3-dev-env which resulted in the local images not being checked in case of PRs in CAPM3, BMO, IPAM etc repositories. The reason being these commands are trying to find `registry.nordix.org/quay-io-proxy/metal3-io/baremetal-operator ` as image in manager.yml but in manager.yaml of all repositories we have `quay.io`. This resulted in sed being a no-op and local images were not tested anymore. This quick fix is falling back to quay.io until proper fix lands.


furkatgofurov7 (approved):
/approve
/lgtm

Merging it since this is a serious bug and we need a quick fix.

PR #848, opened 2021-11-01
Allow creating VMs with emulated TPM device

mkowalski:


fmuyassarov (commented):
do we need to expose the variable to a user as environment variable so that it is possible to tune the value without modifying the Ansible defaults?

hardys:
> do we need to expose the variable to a user as environment variable so that it is possible to tune the value without modifying the Ansible defaults?

Good point - if we add a lookup for that here, https://github.com/openshift-metal3/dev-scripts/pull/1306 is no longer needed

01_prepare_host.sh
```diff
@@ -201,7 +201,7 @@ function init_minikube() {
       # Loop to ignore minikube issues
       while /bin/true; do
         minikube_error=0
-        # Restart libvirtd.service as suggested here 
+        # Restart libvirtd.service as suggested here
```
hardys, line 204: These whitespace changes are unrelated to the rest of the patch, may be worth fixing if you're doing other changes

vm-setup/roles/common/defaults/main.yml
```diff
@@ -13,6 +13,7 @@ default_memory: 8192
 default_vcpu: 2
 num_nodes: 2
 extradisks: false
+tpm_emulator: false
```
hardys, line 16: So here, as @fmuyassarov says we could instead do e.g:

```
tpm_emulator: "{{ lookup('env', 'VM_TPM_EMULATOR')|default(false, true) }}"
```

01_prepare_host.sh
mkowalski: Yeah, let's not do it here

vm-setup/roles/common/defaults/main.yml
mkowalski: Oh, that indeed makes much more sense, thanks!

fmuyassarov (commented):
@mkowalski This looks better. We usually add new vars with a short description in https://github.com/metal3-io/metal3-dev-env/blob/master/vars.md and it would be nice if do the same here too. Otherwise LGTM

fmuyassarov:
/lgtm
/test-v1a5-centos-integration
/test-v1b1-integration 
Thank you!

PR #849, opened 2021-11-02
Add support for DEPLOY_ISO_URL

hardys:
This allows local testing of the BMO functionality added via
https://github.com/metal3-io/baremetal-operator/pull/907

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -110,6 +110,10 @@ IRONIC_ENDPOINT=${IRONIC_URL}
 IRONIC_INSPECTOR_ENDPOINT=${IRONIC_INSPECTOR_URL}
 EOF
 
+  if [ -n "${DEPLOY_ISO_URL}" ]; then
+    echo "DEPLOY_ISO_URL=${DEPLOY_ISO_URL}" | sudo tee -a "${BMOPATH}/config/default/ironic.env"
+  fi
+
   # Deploy. Args: <deploy-BMO> <deploy-Ironic> <deploy-TLS> <deploy-Basic-Auth> <deploy-Keepalived>
   "${BMOPATH}/tools/deploy.sh" true false "${IRONIC_TLS_SETUP}" "${IRONIC_BASIC_AUTH}" true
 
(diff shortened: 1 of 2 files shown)


dtantsur (approved):

03_launch_mgmt_cluster.sh
```diff
@@ -110,6 +110,10 @@ IRONIC_ENDPOINT=${IRONIC_URL}
 IRONIC_INSPECTOR_ENDPOINT=${IRONIC_INSPECTOR_URL}
 EOF
 
+  if [ -n "${DEPLOY_ISO_URL}" ]; then
+    echo "DEPLOY_ISO_URL=${DEPLOY_ISO_URL}" | sudo tee -a "${BMOPATH}/config/default/ironic.env"
+  fi
+
```
fmuyassarov, line 116: why not to append this var in ? https://github.com/metal3-io/metal3-dev-env/blob/08950cc47e6460de941f21229808fade04c7468e/03_launch_mgmt_cluster.sh#L107-L110

03_launch_mgmt_cluster.sh
```diff
@@ -195,6 +199,10 @@ IRONIC_FAST_TRACK=true
 RESTART_CONTAINER_CERTIFICATE_UPDATED="${RESTART_CONTAINER_CERTIFICATE_UPDATED}"
 EOF
 
+  if [ -n "${DEPLOY_ISO_URL}" ]; then
+    echo "DEPLOY_ISO_URL=${DEPLOY_ISO_URL}" | sudo tee -a "$IRONIC_DATA_DIR/ironic_bmo_configmap.env"
+  fi
+
```
fmuyassarov, line 205: same here, we could probably reuse the tee we already do in https://github.com/metal3-io/metal3-dev-env/blob/08950cc47e6460de941f21229808fade04c7468e/03_launch_mgmt_cluster.sh#L183-L184

03_launch_mgmt_cluster.sh
hardys: @fmuyassarov I made it conditional because unlike the kernel/ramdisk DEPLOY_ISO_URL isn't set by default - if we're OK with just appending an empty value then I can just add it to the list

hardys: FYI it's needed for testing virtualmedia via https://github.com/metal3-io/ironic-agent-image/ - so it's not required by default but a user may want to export it when testing that path (and in future perhaps we may want to add fedora-coreos support to metal3-dev-env, but that's not my current aim, I'm just trying to make developer testing easier)

fmuyassarov: Makes sense. I missed that we don't have anything by default for it. So, better to avoid empty value.

hardys:
Looking for an lgtm on this, @kashifest @furkatgofurov7 or @macaptain perhaps you can take a look please?

PR #853, opened 2021-11-05
Add krew installation to 01 script

jan-est:
This PR will add [krew](https://krew.sigs.k8s.io/) installation as part of the 01 script. 

Krew is the plugin manager for `kubectl` command-line tool. Krew will be used to install `promdump`. [Promdump](https://github.com/ihcsim/promdump) is used to dump the head and persistent blocks of Prometheus.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -63,6 +63,19 @@ fi
 # shellcheck disable=SC1091
 source lib/releases.sh
 
+## Install krew
+if ! kubectl krew > /dev/null 2>&1; then
+  cd "$(mktemp -d)" &&
+  OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
+  ARCH="$(uname -m | sed -e 's/x86_64/amd64/' -e 's/\(arm\)\(64\)\?.*/\1\2/' -e 's/aarch64$/arm64/')" &&
+  KREW="krew-${OS}_${ARCH}" &&
+  curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/${KREW}.tar.gz" &&
+  tar zxvf "${KREW}.tar.gz" &&
+  rm -f "${KREW}.tar.gz" &&
+  ./"${KREW}" install krew
+  echo export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH" >> ~/.bashrc
+fi
+
 # Allow local non-root-user access to libvirt
 # Restart libvirtd service to get the new group membership loaded
 if ! id "$USER" | grep -q libvirt; then


01_prepare_host.sh
```diff
@@ -63,6 +63,16 @@ fi
 # shellcheck disable=SC1091
 source lib/releases.sh
 
+## Install krew
+set -x; cd "$(mktemp -d)" &&
```
macaptain, line 67: `set -x` is already done at the top of the script.

01_prepare_host.sh
```diff
@@ -63,6 +63,16 @@ fi
...
+OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
+ARCH="$(uname -m | sed -e 's/x86_64/amd64/' -e 's/\(arm\)\(64\)\?.*/\1\2/' -e 's/aarch64$/arm64/')" &&
+KREW="krew-${OS}_${ARCH}" &&
+curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/${KREW}.tar.gz" &&
+tar zxvf "${KREW}.tar.gz" &&
+./"${KREW}" install krew
+echo export PATH="${KREW_ROOT:-$HOME/.krew}/bin:$PATH" >> ~/.bashrc
```
macaptain, line 74: Everytime this script runs it's going to add a line to .bashrc. This installation is a one-off. So we need to guard if krew is already installed.

01_prepare_host.sh
```diff
@@ -63,6 +63,18 @@ fi
 # shellcheck disable=SC1091
 source lib/releases.sh
 
+## Install krew
+if ! command -v kubectl krew 2>/dev/null; then
```
macaptain, line 67: This is true if `kubectl` is installed, regardless of whether krew is installed.

```
$ kubectl krew
Error: unknown command "krew" for "kubectl"
Run 'kubectl --help' for usage.
```

```
$ if command -v kubectl krew 2>/dev/null; then echo "krew is installed"; fi
/usr/bin/kubectl
krew is installed
```

Probably instead of using `command`, you want to test a `kubectl krew` command which has exit code 0 when krew is installed.

01_prepare_host.sh
```diff
@@ -63,6 +63,18 @@ fi
...
+  cd "$(mktemp -d)" &&
+  OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
+  ARCH="$(uname -m | sed -e 's/x86_64/amd64/' -e 's/\(arm\)\(64\)\?.*/\1\2/' -e 's/aarch64$/arm64/')" &&
+  KREW="krew-${OS}_${ARCH}" &&
+  curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/${KREW}.tar.gz" &&
+  tar zxvf "${KREW}.tar.gz" &&
```
macaptain, line 73: Should this tarball be cleaned up on successful installation?

elfosardo (approved):

fmuyassarov:
I think it is going to be useful to have Prometheus set up, so that we have a clearer picture with the cluster status.
/approve

PR #855, opened 2021-11-08
🐛 Refactor node reuse testing

furkatgofurov7:
This patch fixes the issue in node reuse testing workflow in features tests workflow where:
- after testing KCP node reuse use case, we do scale down KCP to 1 and and it triggers de provisioning of BMHs to start testing MD use case, thus all BMH that were used before would have node reuse labels on them. So while testing MD case, we do not remove them and it will always chooses the same host since other BMHs have already KCP name set in node reuse label. To fix it, we remove old labels from all BMHs and then test MD use case to check if node reuse feature is working.
- also removing marking/unmarking of BMHs with unhealthy annotation, since now when MD maxSurge field is updated with 0 it deletes the existing machine first and then recreates a new one, resulting in unneeded unhealthy annotation setting/unsetting.
- removes unnecessary creation of multiple new M3MTs and workarounds like setting node reuse field to "False" at the end of the tests.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
@@ -49,58 +49,7 @@
       IMAGE_URL: "http://172.22.0.1/images/{{ RAW_IMAGE_NAME }}"
       IMAGE_CHECKSUM: "http://172.22.0.1/images/{{ RAW_IMAGE_NAME }}.{{ IMAGE_CHECKSUM_TYPE }}sum"
 
-  - name: Get cluster uid
-    kubernetes.core.k8s_info:
-      api_version: cluster.x-k8s.io/{{ CAPI_VERSION }}
-      kind: Cluster
-      name: "{{ CLUSTER_NAME }}"
-      namespace: "{{ NAMESPACE }}"
-      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    register: clusters
-
-  - name: Create controlplane Metal3MachineTemplates
-    kubernetes.core.k8s:
-      state: present
-      template: Metal3MachineTemplate.yml
-      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    vars:
-      CLUSTER_UID: "{{ clusters.resources[0].metadata.uid }}"
-      M3MT_NAME: "{{CLUSTER_NAME}}-new-controlplane-image"
-      DATA_TEMPLATE_NAME: "{{CLUSTER_NAME}}-controlplane-template"
-      RAW_IMAGE_NAME: "{{ UPGRADED_RAW_IMAGE_NAME }}"
-      IMAGE_URL: "http://172.22.0.1/images/{{ RAW_IMAGE_NAME }}"
-      IMAGE_CHECKSUM: "http://172.22.0.1/images/{{ RAW_IMAGE_NAME }}.{{ IMAGE_CHECKSUM_TYPE }}sum"
-      NODE_REUSE_STATUS: "true"
-
-  - name: Create worker Metal3MachineTemplates
-    kubernetes.core.k8s:
-      state: present
-      template: Metal3MachineTemplate.yml
-      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    vars:
-      CLUSTER_UID: "{{ clusters.resources[0].metadata.uid }}"
-      M3MT_NAME: "{{CLUSTER_NAME}}-new-workers"
-      DATA_TEMPLATE_NAME: "{{CLUSTER_NAME}}-workers-template"
-      RAW_IMAGE_NAME: "{{ UPGRADED_RAW_IMAGE_NAME }}"
-      IMAGE_URL: "http://172.22.0.1/images/{{ RAW_IMAGE_NAME }}"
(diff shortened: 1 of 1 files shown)


smoshiur1237 (approved):
/lgtm

Rozzii:
let met test this locally, I want to make sure it solves our original issue

mboukhalfa:
> let met test this locally, I want to make sure it solves our original issue

As this passed on CAPM3 v1b1 I guess would be better to open a PR towards v1a5 either here on CAPM3

furkatgofurov7:
> As this passed on CAPM3 v1b1 I guess would be better to open a PR towards v1a5 either here on CAPM3

This passed with v1a5 CAPM3 (5 days ago)  and we moved to v1beta1 in feature tests just yesterday.

Rozzii:
can I modify this with the simplified deletion?

Rozzii:
I have made a test PR based on this that tries to recreate a the original issue: https://github.com/metal3-io/metal3-dev-env/pull/867

01_prepare_host.sh
```diff
@@ -6,6 +6,43 @@ source lib/logging.sh
...
+function clone_repos() {
+  mkdir -p "${M3PATH}"
+  clone_repo "${CAPM3REPO}" "${CAPM3BRANCH}" "${CAPM3PATH}"
+  pushd "${CAPM3PATH}"
+  git checkout "${CAPM3BRANCH}"
+  popd
+}
+
+clone_repos
+
```
mboukhalfa, line 45: is not this duplicated in script 03

mboukhalfa (commented):
LGTM

01_prepare_host.sh
furkatgofurov7: yeah, this is for testing will be removed

furkatgofurov7:
Passed, triggering again to run full tests.
/test-features

furkatgofurov7:
This follows similar change as in [capm3#390](https://github.com/metal3-io/cluster-api-provider-metal3/pull/390 ) but in feature tests.
/assign @fmuyassarov 
/cc @kashifest @Xenwar @namnx228

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -179,7 +132,7 @@
         bmhs.resources | filter_provisioning("available,ready") | map(attribute='metadata.name')
         )
 
-  - name: Check if the deprovisioned BMH is re-used for the next provisioning.
+  - name: Check node reuse logic ensuring the just deprovisioned BMH is re-used for the next provisioning.
```
fmuyassarov, line 135: nit: there is seems to be a typo in this text

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -419,23 +382,8 @@
...
-        apiVersion: metal3.io/v1alpha1
-        kind: BareMetalHost
-        metadata:
-          name: "{{ item }}"
-          annotations:
-            capi.metal3.io/unhealthy: null
-    loop: "{{ saved_bmhs.resources | map(attribute='metadata.name') }}"
-
-  - name: Check if the previous deprovisioned BMH is re-used for the next provisioning.
+  - name: Check node reuse logic ensuring the just deprovisioned BMH is re-used for the next provisioning.
```
fmuyassarov, line 386: same here

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -110,8 +59,15 @@
       definition:
         spec:
           nodeReuse: true
+          template:
+            spec:
+              image:
+                checksum: "http://172.22.0.1/images/{{ UPGRADED_RAW_IMAGE_NAME }}.{{ IMAGE_CHECKSUM_TYPE }}sum"
+                checksumType: "{{IMAGE_CHECKSUM_TYPE}}"
+                format: "{{IMAGE_FORMAT}}"
+                url: "http://172.22.0.1/images/{{ UPGRADED_RAW_IMAGE_NAME }}"
```
fmuyassarov, line 68: dataTemplate is not necessary?

fmuyassarov (commented):
looks good, just a couple of nits

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
furkatgofurov7: no, it uses same dt since we are not generating a new m3mt

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
furkatgofurov7: as is also should be clear, but if you think not, I can change it, wanted to avoid re-running feature tests for name change, let me know your opinion

furkatgofurov7 (commented):
@fmuyassarov thanks for review, PTAL again

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
fmuyassarov: OK sure, "Check node reuse logic ensuring the just deprovisioned" part looked a bit odd to me. But that's not something to worry, let's move on

furkatgofurov7: Ok, I have fixed it now, should be good now

fmuyassarov: This looked cleaner. Thanks

kashifest (approved):

PR #858, opened 2021-11-11
Fix BMO CRDs labelling

kashifest:
`cluster.x-k8s.io/provider="metal3"` label is not necessary for BMO.


Rozzii (approved):

fmuyassarov:
@kashifest can you please add a note why it is not necessary. I believe we will come back to it one day :)
/approve

furkatgofurov7 (commented):
/lgtm

PR #859, opened 2021-11-15
WIP Fix ownership issues

Rozzii:
Ownership related errors were reported when bootstrapping the dev-env
this commit aims to fix the ownership issues.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -135,11 +135,11 @@ EOF
     fi
     if [ "${IRONIC_BASIC_AUTH}" == "true" ]; then
       sudo mkdir -p /opt/metal3/auth/ironic
-      sudo chown "$USER":"$USER" /opt/metal3/auth/ironic
+      sudo chown --recursive "$USER":"$USER" /opt/metal3/auth/ironic
       cp "${IRONIC_AUTH_DIR}ironic-username" /opt/metal3/auth/ironic/username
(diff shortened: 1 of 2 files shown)


03_launch_mgmt_cluster.sh
```diff
@@ -135,11 +135,11 @@ EOF
     fi
     if [ "${IRONIC_BASIC_AUTH}" == "true" ]; then
       sudo mkdir -p /opt/metal3/auth/ironic
-      sudo chown "$USER":"$USER" /opt/metal3/auth/ironic
+      sudo chown --recursive "$USER":"$USER" /opt/metal3/auth/ironic
```
hardys, line 138: It'd be interesting to understand why this is needed, since the chown should happen immediately after directory creation - did the script get run as root, then later as some other user?

We might also consider just removing the `sudo` on directory creation vs mkdir-then-chown

Rozzii: I think here it is simply done like this because the non root user has no writing rights to /opt by default so that could have been the original reason why it is sudo mkdir then the sudo chown  is there to change ownership thus in the future the current user has no need for sudo at least this was my understanding.

So in conclusion it is all the result of working in "/opt".

Rozzii: But I think this --recursive wont be the solution to the permission errors this should have no effect .

Rozzii:
I will close this PR because for now I think the issue I was facing is not an upstream problem.

PR #863, opened 2021-11-16
Print commands as they are executed during fetch_manifests.sh

fmuyassarov:
For some reason, when we fetch custom resources during 05 script,
metal3Machine objects don't include the status. This addition is to
see how YAMLs of those all the CRs is being obtained.

### DIFF
--- a/scripts/fetch_manifests.sh
+++ b/scripts/fetch_manifests.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+set -x
 mkdir -p /tmp/manifests
 
 manifests=(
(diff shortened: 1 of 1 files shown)


scripts/fetch_manifests.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+set -x
```
hardys, line 3: If you also want to log the manifests you might want to replace the ` > /tmp/manifests` below with `tee` ?

fmuyassarov: Good idea! Added it

PR #864, opened 2021-11-16
Migrate upgrade tests to v1beta1 APIs + Point mariadb container spec to the separate image.

namnx228:
This PR modifies the upgrade test scripts to make it compatible with new api versions of CAPI and CAPM3 v1beta1.
Also, this PR fixes the issue with upgrade where mariadb container needs to be deployed by a separate mariadb image

### DIFF
--- a/scripts/feature_tests/upgrade/upgrade.sh
+++ b/scripts/feature_tests/upgrade/upgrade.sh
@@ -9,4 +9,4 @@ source "upgrade_vars.sh"
 
 export ACTION="upgrading"
 
-"${METAL3_DIR}"/scripts/run.sh
+"${METAL3_DIR}"/scripts/run.sh 
--- a/scripts/feature_tests/upgrade/upgrade_vars.sh
+++ b/scripts/feature_tests/upgrade/upgrade_vars.sh
@@ -4,28 +4,13 @@ set -eux
 # Folder created for specific capi release when running
 # ${CLUSTER_API_REPO}/cmd/clusterctl/hack/create-local-repository.py
 
-export CAPIRELEASE_HARDCODED="v0.4.99"
+export CAPIRELEASE_HARDCODED="v1.0.99"
 
-function get_latest_capm3_release() {
-    clusterctl upgrade plan | grep infrastructure-metal3 | awk 'NR == 1 {print $5}'
-}
+export CAPIRELEASE="v1.0.0"
+export CAPI_REL_TO_VERSION="v1.0.1"
 
-# CAPM3 release version which we upgrade from.
-export CAPM3RELEASE="v0.5.0"
-CAPM3_REL_TO_VERSION="$(get_latest_capm3_release)" || true
-# CAPM3 release version which we upgrade to.
-export CAPM3_REL_TO_VERSION
-
-# Fetch latest release version of CAPI from the output of clusterctl command.
-function get_latest_capi_release() {
-    clusterctl upgrade plan | grep cluster-api | awk 'NR == 1 {print $5}'
-}
-
-# CAPI release version which we upgrade from.
-export CAPIRELEASE="v0.4.1"
-CAPI_REL_TO_VERSION="$(get_latest_capi_release)" || true
-# CAPI release version which we upgrade to.
-export CAPI_REL_TO_VERSION
+export CAPM3RELEASE="v1.0.0"
+export CAPM3_REL_TO_VERSION="v1.0.0"
 export FROM_K8S_VERSION="v1.22.2"
 export KUBERNETES_VERSION=${FROM_K8S_VERSION}
 export UPGRADED_K8S_VERSION="v1.22.3"
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -xe
-
 METAL3_DIR="$(dirname "$(readlink -f "${0}")")/.."
 
 ACTION="${ACTION:-""}"
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -xe
 
 METAL3_DIR="$(dirname "$(readlink -f "${0}")")/.."
 
--- a/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
@@ -14,7 +14,7 @@
 
   - name: Cleanup - remove existing next versions of controlplane components CRDs
     vars:
-      working_dir: "{{HOME}}/.cluster-api/dev-repository/"
+      working_dir: "{{HOME}}/.cluster-api/dev-repository"
     file:
       state: absent
       path: "{{item}}"
@@ -185,7 +185,7 @@
         replace: "upgm3c"
 
   - name: Perform upgrade on the target cluster
-    ansible.builtin.command: clusterctl upgrade apply --contract v1alpha4
+    ansible.builtin.command: clusterctl upgrade apply --contract v1beta1
     environment:
       KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
 
@@ -202,7 +202,7 @@
       KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
 
   - name: Perform upgrade on the source cluster
-    ansible.builtin.command: clusterctl upgrade apply --contract v1alpha4
+    ansible.builtin.command: clusterctl upgrade apply --contract v1beta1
 
   # TODO: Can we check this somehow instead of just waiting?
   # Relevant issue: https://github.com/kubernetes-sigs/cluster-api/issues/4474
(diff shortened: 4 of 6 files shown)


scripts/feature_tests/upgrade/upgrade_vars.sh
```diff
@@ -11,8 +11,17 @@ function get_latest_capm3_release() {
...
 # CAPM3 release version which we upgrade from.
-export CAPM3RELEASE="v0.5.0"
-CAPM3_REL_TO_VERSION="$(get_latest_capm3_release)" || true
+export CAPM3RELEASE="v0.5.2"
+# The if below is a workaround until the v0.5.3 or later version is released
+if get_latest_capm3_release | grep -q 'Already'; then
+    export CAPM3_REL_TO_VERSION=${CAPM3RELEASE}
+else
+    # CAPM3 release version which we upgrade to.
+    export CAPM3_REL_TO_VERSION
```
furkatgofurov7, line 20: This is not correct, if we find new release we have to fetch that release version and set it properly
```suggestion
    # CAPM3 release version which we upgrade to.
    CAPM3_REL_TO_VERSION="$(get_latest_capm3_release)" || true
    export CAPM3_REL_TO_VERSION
```

scripts/feature_tests/upgrade/upgrade_vars.sh
```diff
@@ -11,8 +11,17 @@ function get_latest_capm3_release() {
...
+fi
+# TODO: When remove the workaround, uncomment the command below
+# CAPM3_REL_TO_VERSION="$(get_latest_capm3_release)" || true
```
furkatgofurov7, line 23: can be removed since covered in the condition above

scripts/feature_tests/upgrade/upgrade_vars.sh
```diff
@@ -11,18 +11,23 @@ function get_latest_capm3_release() {
 }
 
 # CAPM3 release version which we upgrade from.
-export CAPM3RELEASE="v0.5.0"
-CAPM3_REL_TO_VERSION="$(get_latest_capm3_release)" || true
-# CAPM3 release version which we upgrade to.
-export CAPM3_REL_TO_VERSION
+export CAPM3RELEASE="v0.5.2"
+
+if get_latest_capm3_release | grep -q 'Already'; then
```
fmuyassarov, line 16: does it mean that we are already in the latest release of the provider?  If yes, why not skip verifying it and go directly `get_latest_capm3_release()`
basically,
```
CAPM3_REL_TO_VERSION="$(get_latest_capm3_release)" || true
export CAPM3_REL_TO_VERSION
```

fmuyassarov (commented):
looks good, just one question

scripts/feature_tests/upgrade/upgrade_vars.sh
namnx228: If I am not wrong, if the CAPM3RELEASE is already the latest version, `get_latest_capm3_release` will not return the latest version, but the word "Already", which is not the thing we can assign to `CAPM3_REL_TO_VERSION`.

fmuyassarov: thanks, didn't know that. Why don't we do it for `CAPI_REL_TO_VERSION`  though?

furkatgofurov7: @fmuyassarov the reason for checking  `if get_latest_capm3_release | grep -q 'Already';` is, when we do not have yet release to upgrade, we set CAPM3_REL_TO_VERSION to CAPM3RELEASE which is no-op. Once we have 2 releases of CAPM3, we can set  `CAPM3RELEASE=1.0.0` and `CAPM3_REL_TO_VERSION=1.0.1` i.e. We do not need it for `CAPI_REL_TO_VERSION` because there is already 2 releases available, so `CAPIRELEASE=1.0.0` and `CAPI_REL_TO_VERSION` would be 1.0.1 based on  `clusterctl upgrade plan | grep cluster-api | awk 'NR == 1 {print $5}'`

scripts/feature_tests/upgrade/upgrade_vars.sh
```diff
@@ -11,18 +11,23 @@ function get_latest_capm3_release() {
 }
 
 # CAPM3 release version which we upgrade from.
-export CAPM3RELEASE="v0.5.0"
-CAPM3_REL_TO_VERSION="$(get_latest_capm3_release)" || true
-# CAPM3 release version which we upgrade to.
-export CAPM3_REL_TO_VERSION
+export CAPM3RELEASE="v0.5.2"
```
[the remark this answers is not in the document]
mboukhalfa: I think there is no v1.0.0 release on CAPM3 repo. So it needs to be created before using it here

scripts/feature_tests/upgrade/upgrade_vars.sh
fmuyassarov: this is a workaround for now, got it

scripts/run.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -xe
+set -x
```
Xenwar, line 3: do we also need set `+x `at the end ?

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -259,7 +259,7 @@
       name: metal3clusters.infrastructure.cluster.x-k8s.io
       kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     register: crd
-    failed_when: '"upgm3c" not in crd.resources[0].spec.names.shortNames'
+    failed_when: '"upgm3c" not in crd.resources[0].spec.names.shortNames and CAPM3RELEASE != CAPM3_REL_TO_VERSION'
```
Xenwar, line 262: I think the version comparison should be  a `when` condition by itself  so that the assertion runs only when the versions are not the same

Rozzii (approved):

scripts/feature_tests/upgrade/upgrade.sh
```diff
@@ -9,4 +9,4 @@ source "upgrade_vars.sh"
 
 export ACTION="upgrading"
 
-"${METAL3_DIR}"/scripts/run.sh
+"${METAL3_DIR}"/scripts/run.sh || true
```
lentzi90, line 12: Will not this mean that it will just continue even if upgrade fails and we may not notice it?

namnx228: Ouch!!! That was used for debugging. I forgot to remove it. Good catch, @lentzi90

scripts/run.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -xe
```
kashifest, line 3: why removing this?

furkatgofurov7 (approved):
Thanks for taking care of it @namnx228

furkatgofurov7:
/retitle Migrate upgrade tests to v1beta1 APIs

kashifest:
I have a question. Didn't we agree that we will test upgrade from v1a4 capi to v1b1 CAPI @furkatgofurov7 ? Whats the point of checking patch version upgrade anymore?

furkatgofurov7:
> I have a question. Didn't we agree that we will test upgrade from v1a4 capi to v1b1 CAPI @furkatgofurov7 ?

Did we? I am not sure, but we agreed to have that in e2e tests for sure.

> Whats the point of checking patch version upgrade anymore?
Well, since we started we had tested patch version afaik, so I think we followed that pattern here too. Also, it is not only v1a4-->v1b1, we have to support v1a5-->v1b1 too, at least that's what we are trying to cover in e2e tests.

kashifest:
> Did we? I am not sure, but we agreed to have that in e2e tests for sure.

Yes as far as I remember.
 

> Well, since we started we had tested patch version afaik, so I think we followed that pattern here too. Also, it is not only v1a4-->v1b1, we have to support v1a5-->v1b1 too, at least that's what we are trying to cover in e2e tests.

by v1a4 I mean CAPI so thats v1a5-> v1b1 since that would be the most challenging one. e2e tests are still flaky and I think since ansible tests are mostly used currently, it should be tested here as well.

furkatgofurov7:
> > Did we? I am not sure, but we agreed to have that in e2e tests for sure.
> 
> Yes as far as I remember.
> 
> > Well, since we started we had tested patch version afaik, so I think we followed that pattern here too. Also, it is not only v1a4-->v1b1, we have to support v1a5-->v1b1 too, at least that's what we are trying to cover in e2e tests.
> 
> by v1a4 I mean CAPI so thats v1a5-> v1b1 since that would be the most challenging one. e2e tests are still flaky and I think since ansible tests are mostly used currently, it should be tested here as well.

If that is the case, this patch needs a change

kashifest:
Lets take this patch in to get the upgrade tests passing at least with patch version upgrade. However, the end goal should be to upgrade CAPI v1a4 api to CAPI v1b1 api since thats what most of the users would want to test. lets tackle this in a followup PR.
/lgtm

kashifest:
> This PR modifies the upgrade test scripts to make it compatible with new api versions of CAPI and CAPM3 v1beta1.

Can you please also add mariadb in the title and the description
/hold
/lgtm

namnx228:
> > This PR modifies the upgrade test scripts to make it compatible with new api versions of CAPI and CAPM3 v1beta1.
> 
> Can you please also add mariadb in the title and the description /hold /lgtm

@kashifest  Done. I will unhold this PR. 
/hold cancel

PR #866, opened 2021-11-17
Don't start fetching CRs until M3Machine status is rebuilt

fmuyassarov:
Currently we don't have status of Metal3Machine objects in the
tar of manifests we get at the end of the run. And the reason is
we call fetch_manifests.sh after we pivot back and it starts very
quick even before Metal3Machine controller rebuilds the status
after metal3Machine objects are pivoted back. This patch holds on
running fetch_manifests.sh script until the status of metal3Machine
is available.
Note: when clusterctl move, status is not moved, and it is job of
controllers to rebuild the status.

### DIFF
--- a/05_test.sh
+++ b/05_test.sh
@@ -11,6 +11,13 @@ export ACTION="ci_test_provision"
 
 "${METAL3_DIR}"/scripts/run.sh
 
+# wait until status of Metal3Machine is rebuilt
+while [ -z "${status}" ]
+do
+    status=$(kubectl get m3m -n "${NAMESPACE}" -o=jsonpath="{.items[*]['status.ready']}")
+    sleep 1s
+done
+
 "${METAL3_DIR}"/scripts/fetch_manifests.sh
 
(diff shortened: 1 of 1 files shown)


kashifest (approved):

05_test.sh
```diff
@@ -11,6 +11,12 @@ export ACTION="ci_test_provision"
 
 "${METAL3_DIR}"/scripts/run.sh
 
+# wait until status of Metal3Machine is rebuilt
+while [ -z "${status}" ]
+do
+    status=$(kubectl get m3m -n "${NAMESPACE}" -o=jsonpath="{.items[*]['status.ready']}")
```
mboukhalfa, line 17: Don't we need to sleep sometime in between retries ?

fmuyassarov: yep that could be easier, but I didn't know what would be the best value for the sleep

fmuyassarov: I know that controller is fast enough to rebuilt the status, perhaps ~10-15 secs would be enough ?, not sure

fmuyassarov: the problem with this approach is, it relies only on m3m, and what if there is another object that is waiting for its status to be rebuilt too. Then this won't be enough. But controller does reconciliation in parallel though :)

mboukhalfa: then probably sleep for 1s after retry should be ok

fmuyassarov: agree

PR #873, opened 2021-11-24
Pin calico version to v3.21.0

fmuyassarov:
This will avoid patching calico version in airship-dev-tools
repo, where we pre pull these images. As such, we won't need
to merge renovate patches, then build node images, and finally
build source cluster images every time new minor/patch release
of Calico is out. This will ensure that we do those steps only when
necessary (i.e. when we really need to bump image version for some
reason).

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/verify.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/verify.yml
@@ -94,6 +94,20 @@
       kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     register: install_cni
 
+  - name: Wait (maximum 3 mins) until Calico pods start running
+    kubernetes.core.k8s_info:
+      api_version: v1
+      kind: Pod
+      namespace: kube-system
+      kubeconfig: /tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml
+      field_selectors:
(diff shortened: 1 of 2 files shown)


vm-setup/roles/v1aX_integration_test/tasks/verify.yml
```diff
@@ -58,13 +58,19 @@
       block: "{{ kubeconfig_secret.resources[0].data.value | b64decode }}"
 
   # Install Calico
-  - name: Download Calico manifest
+  - name: Download Calico v3.21.x manifests
     get_url:
-      url: https://docs.projectcalico.org/manifests/calico.yaml
+      url: https://docs.projectcalico.org/archive/v3.21/manifests/calico.yaml
```
macaptain, line 63: I think it would be nice if this Calico version were an Ansible 'defaults' or 'var'. Then we could also point Renovate at it.

macaptain (commented):
Great to have this pinned at last!

vm-setup/roles/v1aX_integration_test/tasks/verify.yml
fmuyassarov: Agree on adding default value as ansible var.

kashifest (approved):

fmuyassarov:
Sorry for spamming :) missed registration.

PR #876, opened 2021-11-26
Fetch container logs before pivoting when it is minikube cluster

fmuyassarov:
When host is CentOS (minikube cluster), we skip /tmp/podman dir creation, because
we don't fetch any container logs as we do for Ubuntu host (kind cluster). For that
reason, copying of logs which don't exist fails in https://github.com/metal3-io/project-infra/blob/83a5967f2c97886664adeaf714f79d72be47834b/jenkins/scripts/files/run_fetch_logs.sh#L60.
```
Warning: Permanently added '31.132.58.190' (ECDSA) to the list of known hosts.
cp: cannot stat '/tmp/podman/*': No such file or directory
```
Adding container logs fetching for minikube cluster based setup, and this will also
avoid failure with `cp`.
Also added registry, sushy-tools, vbmc containers to the list.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/move.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/move.yml
@@ -2,54 +2,65 @@
   - name: Define number of BMH's
     set_fact:
       NUMBER_OF_BMH: "{{ NUM_OF_MASTER_REPLICAS|int +  NUM_OF_WORKER_REPLICAS|int }}"
+      general_containers:
+          - httpd-infra
+          - registry
+          - sushy-tools
+          - vbmc
+      ironic_containers:
+          - ironic-api
+          - ironic-conductor
+          - ironic-inspector
+          - ironic-endpoint-keepalived
+          - ironic-log-watch
+          - dnsmasq
+          - mariadb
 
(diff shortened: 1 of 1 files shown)


vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -3,53 +3,84 @@
...
+            - ironic-log-watch
+            - dnsmasq
+            - httpd-infra
+            - mariadb
+            - registry
+            - sushy-tools
+            - vbmc
+
+      - name: Fetch container logs before pivoting (kind cluster)
+        shell: "sudo {{ CONTAINER_RUNTIME }} logs {{ item }} > /tmp/{{ CONTAINER_RUNTIME }}/{{ item }}/stdout.log 2> /tmp/{{ CONTAINER_RUNTIME }}/{{ item }}/stderr.log"
```
Xenwar, line 27: Is `2>` in this case is getting the error encountered during log collection process ? or trying to separate logs into stdout and stderr ?

fmuyassarov: the latter option

kashifest (approved):

PR #877, opened 2021-11-29
WIP: Fallback to pulling container images from quay if Harbor is down

furkatgofurov7:
Nordix harbor where we pull container images from using proxy cache to be prepared for quay outage, is itself down due to some unrelated issues of Nordix infra itself, i.e:
`Error response from daemon: Get "https://registry.nordix.org/v2/": x509: certificate has expired or is not yet valid: current time 2021-11-29T06:14:29Z is after 2021-11-29T02:28:37Z `

To avoid this and not fail in the middle of tests in the future, we can fall back to pulling container images from quay itself since the chances where both registries are down at the same time is low and if that is the case even we have nothing to try . This patch does not add a retry logic for pulling images since this does not make much sense to try pulling N times when default registry is down, so it just falls back to latter option where it pulls them from quay.


furkatgofurov7:
This is not needed.

/close

PR #879, opened 2021-12-06
Add flag to allow disabling of DHCP

imain:
In order to verify static IP functionality in the installer, we need to allow disabling of DHCP in our virtual networks.

Co-Authored-By: Caleb Boylan <cboylan@redhat.com>

### DIFF
--- a/lib/network.sh
+++ b/lib/network.sh
@@ -118,6 +118,8 @@ else
     exit 1
 fi
 
+export DHCP_ENABLED=${DHCP_ENABLED:-"true"}
+
 if [[ "${EPHEMERAL_CLUSTER}" == "minikube" ]] && [[ -n "${EXTERNAL_SUBNET_V6}" ]]; then
     network_address MINIKUBE_BMNET_V6_IP "${EXTERNAL_SUBNET_V6}" 9
 fi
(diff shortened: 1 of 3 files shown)


vm-setup/roles/common/defaults/main.yml
```diff
@@ -58,6 +59,7 @@ external_network:
     forward_mode: "{% if manage_baremetal == 'y' %}nat{% else %}bridge{% endif %}"
     address_v4: "{{ baremetal_network_cidr_v4|nthhost(1)|default('', true) }}"
     netmask_v4: "{{ baremetal_network_cidr_v4|ipaddr('netmask') }}"
+    dhcp_enabled: "{{ virsh_dhcp_enabled }}"
```
hardys, line 62: We need to either default this in the j2 template, or add the new key to https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/common/tasks/extra_networks_tasks.yml

lib/network.sh
```diff
@@ -118,6 +118,8 @@ else
     exit 1
 fi
 
+export DHCP_ENABLED=${DHCP_ENABLED:-"true"}
```
hardys, line 121: It'd be good to update https://github.com/metal3-io/metal3-dev-env/blob/master/vars.md

hardys:
/ok-to-test

A few comments but overall looks good :)

hardys:
After a closer look I pushed https://github.com/metal3-io/metal3-dev-env/pull/902 which I think may be a cleaner solution - we then don't need any new flag and can just define the network like:

```    
    dhcp_range_v4: "{{ [ baremetal_network_cidr_v4|nthhost(20), baremetal_network_cidr_v4|nthhost(60) ] if lookup('env', 'STATIC_IPS') == '' else [] }}"
    dhcp_range_v6: "{{ [ baremetal_network_cidr_v6|nthhost(20), baremetal_network_cidr_v6|nthhost(60) ] if lookup('env', 'STATIC_IPS') == '' else [] }}"
```

hardys:
https://github.com/metal3-io/metal3-dev-env/pull/902 landed so closing this

PR #533, opened 2020-10-28
Don't clone Cluster API

fmuyassarov:
Don't clone Cluster-api since we are not building clusterctl from CAPI code anymore.

PR #882, opened 2021-12-08
Include CAPI repo to the list of cloned repos

Xenwar:
Include CAPI repo to the list of cloned repos. required in Tilt based environment setup.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -58,6 +58,7 @@ function clone_repos() {
   clone_repo "${BMOREPO}" "${BMOBRANCH}" "${BMOPATH}" "${BMOCOMMIT}"
   clone_repo "${CAPM3REPO}" "${CAPM3BRANCH}" "${CAPM3PATH}"
   clone_repo "${IPAMREPO}" "${IPAMBRANCH}" "${IPAMPATH}"
+  clone_repo "${CAPIREPO}" "${CAPIBRANCH}" "${CAPIPATH}"
 }
 
 # ------------------------------------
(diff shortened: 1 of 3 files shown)


namnx228 (commented):
/lgtm

furkatgofurov7:
@Xenwar just a suggestion: I think we can just revert https://github.com/metal3-io/metal3-dev-env/pull/533 it is more cleaner IMO and this is what you are trying to achieve here, WDYT?

Xenwar:
> @Xenwar just a suggestion: I think we can just revert #533 it is more cleaner IMO and this is what you are trying to achieve here, WDYT?

Thanks. That is exactly what I am trying to achieve.

Xenwar:
update pr with content of #533. Not sure if there are better ways of achieving that.

furkatgofurov7 (approved):

furkatgofurov7:
cc @namnx228

kashifest:
Shall we clone CAPI all the time or shall we do it only when `EPHEMERAL_CLUSTER=tilt` ?

Xenwar:
> Shall we clone CAPI all the time or shall we do it only when `EPHEMERAL_CLUSTER=tilt` ?

I would suggest all the time since cloning it is not that expensive. On the other hand, making it conditional just make the code bigger.

kashifest:
> > Shall we clone CAPI all the time or shall we do it only when `EPHEMERAL_CLUSTER=tilt` ?
> 
> I would suggest all the time since cloning it is not that expensive. On the other hand, making it conditional just make the code bigger.

Ok. 
/lgtm

PR #835, opened 2021-10-19
⚠️ Parameterize container image registry var

furkatgofurov7:
This patch should help us:
- to overcome a problem of quay.io inaccessibility and when we are not able to pull container images from that registry. 
- to download container images faster since it's local infra

[Proxy Cache](https://goharbor.io/docs/2.1.0/administration/configure-proxy-cache/) in [Nordix Harbor registry](https://registry.nordix.org/) was configured so where we can pull container images through it and all of them will be cached in [quay-io-proxy](https://registry.nordix.org/harbor/projects/32/repositories) repo in Nordix Harbor making them available always.

Upd:
Related to [project-infra#252](https://github.com/metal3-io/project-infra/pull/252)

PR #887, opened 2021-12-14
Add deprovision target to clean up target cluster

lentzi90:
~~This makes the `host_cleanup.sh` script delete the local registry container. It should eliminate problems with stale images being used when using `pull-if-not-present`.~~

**Edit:** Adds deprovision target to the Makefile that cleans up the target cluster. This is useful when the BMHs are not VMs that can be thrown away. We could use this in the BML pipeline, see https://github.com/metal3-io/project-infra/pull/292

I also made Ironic inject the public ssh key in the ramdisk. If this is not wanted I can drop the commit, but I think it can be useful for debugging.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -208,6 +208,7 @@ IRONIC_INSPECTOR_ENDPOINT=${IRONIC_INSPECTOR_URL}
 CACHEURL=http://${PROVISIONING_URL_HOST}/images
 IRONIC_FAST_TRACK=true
 RESTART_CONTAINER_CERTIFICATE_UPDATED="${RESTART_CONTAINER_CERTIFICATE_UPDATED}"
+IRONIC_RAMDISK_SSH_KEY=${SSH_PUB_KEY_CONTENT}
 EOF
 
   if [ -n "${DEPLOY_ISO_URL}" ]; then
@@ -232,7 +233,7 @@ EOF
 
     # Copy the generated configmap for ironic deployment
     cp "$IRONIC_DATA_DIR/ironic_bmo_configmap.env"  "${BMOPATH}/ironic-deployment/keepalived/ironic_bmo_configmap.env"
-    
+
     # Deploy. Args: <deploy-BMO> <deploy-Ironic> <deploy-TLS> <deploy-Basic-Auth> <deploy-Keepalived>
     "${BMOPATH}/tools/deploy.sh" false true "${IRONIC_TLS_SETUP}" "${IRONIC_BASIC_AUTH}" true
 
@@ -240,7 +241,7 @@ EOF
     mv "${BMOPATH}/ironic-deployment/ironic/ironic.yaml.orig" "${BMOPATH}/ironic-deployment/ironic/ironic.yaml"
     mv "${BMOPATH}/ironic-deployment/keepalived/keepalived_patch.yaml.orig" "${BMOPATH}/ironic-deployment/keepalived/keepalived_patch.yaml"
   fi
-  
+
   # Restore original files
   mv "${BMOPATH}/ironic-deployment/keepalived/ironic_bmo_configmap.env.orig" "${BMOPATH}/ironic-deployment/keepalived/ironic_bmo_configmap.env"
   popd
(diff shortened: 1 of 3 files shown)


host_cleanup.sh
```diff
@@ -31,6 +31,8 @@ if [ "${CAPM3_RUN_LOCAL}" = true ]; then
   fi
 fi
 
+# Clean up the local container registry
+sudo "${CONTAINER_RUNTIME}" rm registry -f || true
```
macaptain, line 35: I think this won't be enough to solve the problem we have in the Bare Metal Lab, because the any images that have been pulled from the internet to Docker's image storage won't be cleaned, and on next run, `pull-if-not-present` will skip the pull, and the same old images will be pushed to the registry again. I expect you may need to do a `docker image prune` here as well.

The history of this is, we once had a lot of CI failures due to hitting Docker's API rate limit, and one of the things I tried to mitigate is to ensure we didn't pull any images during CI runs, but only in CI VM image build. My unwise idea was to not attempt `docker pull` if we already had the image matching the tag we need. Then when we build the image and run the 01 script there on a fresh VM, we pull the images, but in CI run, we have the images already, so no matter how many times you run the build, no `docker pull` is done. Hence `pull-if-not-present`, and this works to a point. Unfortunately, the problem is that a lot of the images are tagged `latest`: then if updates are pushed upstream, the new function doesn't bother to try to pull them, since it sees it already has a matching image named :latest. In long running environments like the BML, we end up with stale 'latest' images locally, which are then kept in the local registry.

The real fix to the problem above is to use a registry which caches the images and get rid of `pull-if-not-present`. Then we can do a pull, get the freshest image if it exists every time and not hit a rate limit. @furkatgofurov7 had success setting this up for Quay, so this may already exist and I think it's the best way to solve both the rate limit issue and to make sure we have the latest images in every environment.

The other solution to fix the Bare Metal Lab is to add image prune and registry teardown to the Bare Metal Lab specific playbook: https://github.com/metal3-io/project-infra/blob/d369c5a96c31d7c8e73b73d45dc7959d4323c346/jenkins/scripts/bare_metal_lab/deploy-lab.yaml#L59

lentzi90: Oh I see. Thank you @macaptain ! I'll investigate if we can use Quay.
I would argue that we should clean the local registry always when running `make clean` anyway, and I'll add pruning to this. Or do you mean that this should only be done in the BML environment? I guess there could be reasons to not prune and remove the local registry all the time, is this the case?
What I'm trying to say is that the BML specific playbook runs `make clean` so I thought it would make most sense to add the cleaning to the script that is run when you do this.

macaptain: I think the reason for not cleaning the registry and pruning the images is just that it's quite a lot to download each time. If you had a fresh image, the `docker push` _should_ update the registry. But as described, the images go stale because we stop pulling them from the internet once we have them the first time.

I think also we don't want to issue a full `docker image prune` in metal3-dev-env - it may delete other images unrelated to metal3. Instead, using the same loop we use to `pull` the images to instead `rm` the images.

In order of preference, I think the best way forward is:

1) Use @furkatgofurov7 's registry proxy in metal3-dev-env and delete `pull-if-not-present`
2) Fix the BML playbook to properly clean the registry and images
3) Selectively delete metal3 images in metal3-dev-env when running `make clean`

furkatgofurov7: @macaptain I missed a context here, which container image we are talking about here to pull from registry?

macaptain: @furkatgofurov7 

https://github.com/metal3-io/metal3-dev-env/pull/835
https://github.com/metal3-io/project-infra/pull/252

We're using this proxy to pull from Quay, but could we use this proxy registry in the same way for images pulled from DockerHub?

furkatgofurov7: @macaptain I think no, since proxy has been configured to cache images only from metal3 quay.io, all the images we want from DockerHub needs to be in quay.

lentzi90: Thanks for the links @macaptain , this is really helpful!
It looks like Nordix already have a [Docker Hub proxy](https://wiki.nordix.org/display/DEV/Docker+Hub+Proxy) that we could probably use.

furkatgofurov7: @lentzi90 @macaptain actually there is already another docker-hub-proxy cache in [registry](https://registry.nordix.org/harbor/projects/25/repositories) for DockerHub images and that should work

lentzi90:
After some consideration and discussions (see https://github.com/metal3-io/project-infra/pull/292) I think this is no longer needed. There are already scripts to run to deprovision and a Makefile target will probably not be used much.
Closing.

PR #892, opened 2021-12-21
Support proxy for Docker Hub

lentzi90:
We already have `CONTAINER_REGISTRY` for setting the "main" registry to use (basically quay.io proxy). With this PR we can also set `DOCKER_HUB_PROXY` to use a different proxy for Docker Hub. The default is `docker.io`, which basically means no change.

For the registry image, we used to just specify `registry`. I changed this to include the Docker Hub proxy and set the path to `library/registry` since this is the "real" image on Docker Hub. It defaults to `library/` if nothing is specified on Docker Hub, but this does not work well with proxies.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -233,7 +233,7 @@ EOF
 
     # Copy the generated configmap for ironic deployment
     cp "$IRONIC_DATA_DIR/ironic_bmo_configmap.env"  "${BMOPATH}/ironic-deployment/keepalived/ironic_bmo_configmap.env"
-    
+
     # Deploy. Args: <deploy-BMO> <deploy-Ironic> <deploy-TLS> <deploy-Basic-Auth> <deploy-Keepalived>
     "${BMOPATH}/tools/deploy.sh" false true "${IRONIC_TLS_SETUP}" "${IRONIC_BASIC_AUTH}" true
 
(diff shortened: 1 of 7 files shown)


Rozzii (approved):

03_launch_mgmt_cluster.sh
```diff
@@ -500,21 +500,15 @@ create_clouds_yaml
...
   patch_clusterctl
   launch_cluster_api_provider_metal3
-fi
 
-if [ "${EPHEMERAL_CLUSTER}" != "tilt" ]; then
     if [ "${CAPM3_VERSION}" != "v1alpha4" ]; then
         launch_baremetal_operator
     fi
     launch_ironic
-fi
```
fmuyassarov, line 515: I think we can improve a bit further 
```shell
if [ "${EPHEMERAL_CLUSTER}" != "tilt" ]; then
  start_management_cluster
  kubectl create namespace metal3

  patch_clusterctl
  launch_cluster_api_provider_metal3

    if [ "${CAPM3_VERSION}" == "v1alpha4" ]; then
        BMO_NAME_PREFIX="${NAMEPREFIX}-baremetal-operator"
    else
        BMO_NAME_PREFIX="${NAMEPREFIX}"
        launch_baremetal_operator
    fi
    launch_ironic

```

fmuyassarov (approved):

03_launch_mgmt_cluster.sh
lentzi90: Good point, thanks! Will update :slightly_smiling_face:

furkatgofurov7 (commented):
/lgtm

PR #894, opened 2021-12-30
Make upgrade test CAPI v1a4 -> v1b1

lentzi90:
The upgrade tests previously just did a patch version upgrade. In order
to properly test changes in the contract we now test upgrades where CAPI
goes from v1alpha4 to v1beta1 and CAPM3 follows from v1alpha5 to
vabeta1.

### DIFF
--- a/scripts/feature_tests/upgrade/upgrade_vars.sh
+++ b/scripts/feature_tests/upgrade/upgrade_vars.sh
@@ -6,11 +6,17 @@ set -eux
 
 export CAPIRELEASE_HARDCODED="v1.0.99"
 
-export CAPIRELEASE="v1.0.0"
+export CAPI_VERSION="v1alpha4"
+export CAPIRELEASE="v0.4.5"
 export CAPI_REL_TO_VERSION="v1.0.1"
 
-export CAPM3RELEASE="v1.0.0"
+export CAPM3_VERSION="v1alpha5"
+export CAPM3RELEASE="v0.5.3"
 export CAPM3_REL_TO_VERSION="v1.0.0"
+export UPGRADED_CAPM3_VERSION="v1beta1"
+
+# Ubuntu is hard coded in the upgrade tests. Make sure we use it throughout.
+export IMAGE_OS="Ubuntu"
 export FROM_K8S_VERSION="v1.22.2"
 export KUBERNETES_VERSION=${FROM_K8S_VERSION}
 export UPGRADED_K8S_VERSION="v1.22.3"
--- a/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
@@ -1,5 +1,12 @@
 ---
 
+  - name: Fetch container images before upgrade
+    shell: |
+      kubectl get pods -A -o jsonpath="{range .items[*].spec.containers[*]}{.image}{'\n'}{end}" |
+      sort | uniq > /tmp/manifests/container_images_before_upgrade.txt
+    environment:
+      KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+
 # ---------- ---------- ---------- ---------- ---------- ---------- ---------- ----------
 #                       Upgrade controlplane components                                 |
 # ---------- ---------- ---------- ---------- ---------- ---------- ---------- ----------
@@ -35,6 +42,16 @@
       dest: /tmp/cluster-api-clone
       version: "{{ CAPI_REL_TO_VERSION }}"
 
+  - name: Checkout upgraded version of capm3
+    ansible.builtin.git:
+      repo: 'https://github.com/metal3-io/cluster-api-provider-metal3.git'
+      dest: "{{ CAPM3PATH }}"
+      version: "{{ upgraded_capm3_branch }}"
+      # No need to clone, it is already available
+      clone: no
+      # The repo is modified as part of deploying capm3 so we must force
+      force: yes
+
   - name: Create clusterctl-settings.json for cluster-api and capm3 repos
     ansible.builtin.template:
       src: "{{ item.src }}"
@@ -66,6 +83,48 @@
       chdir: /tmp/cluster-api-clone/
     environment:
       KUBECONFIG: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+
+  - name: Update kustomizations with images to use
+    ansible.builtin.command: "{{ item }}"
+    args:
+      chdir: "{{ CAPM3PATH }}"
+    environment:
+      MANIFEST_IMG: "{{ upgraded_capm3_image }}"
+      MANIFEST_TAG: "{{ upgraded_capm3_tag }}"
+      MANIFEST_IMG_IPAM: "{{ upgraded_ipam_image }}"
+      MANIFEST_TAG_IPAM: "{{ upgraded_ipam_tag }}"
+    loop:
+      - make set-manifest-image
+      - make set-manifest-image-ipam
+
+  - name: Build IPAM manifests
+    ansible.builtin.command: "{{ item }}"
+    args:
+      chdir: "{{ CAPM3PATH }}"
+    loop:
+      - make hack/tools/bin/kustomize
+      - ./hack/tools/bin/kustomize build "{{ IPAMPATH }}/config/default" --output config/ipam/metal3-ipam-components.yaml
+
+  - name: Update CAPM3 imports to use local IPAM manifest
+    ansible.builtin.replace:
+      path: "{{ CAPM3PATH }}/config/ipam/kustomization.yaml"
+      regexp: https://github\.com/metal3-io/ip-address-manager/releases/download/v.*/ipam-components\.yaml
+      replace: metal3-ipam-components.yaml
+
+  - name: Make capm3 release manifests
+    ansible.builtin.command: make release-manifests
+    args:
+      chdir: "{{ CAPM3PATH }}"
+
+  # Note: release manifests are not impacted by this (they are in .gitignore)
+  - name: Clean CAPM3 repo after building release manifests
+    ansible.builtin.git:
+      repo: 'https://github.com/metal3-io/cluster-api-provider-metal3.git'
+      dest: "{{ CAPM3PATH }}"
+      version: "{{ upgraded_capm3_branch }}"
+      clone: no
+      # Force to discard local changes
+      force: yes
 
   - name: Create folder structure for next version controlplane components
     vars:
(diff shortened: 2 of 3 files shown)


vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -234,32 +243,26 @@
...
-  - name: Verify upgraded API resource for Metal3Clusters
+  - name: Verify upgraded CRDs for Metal3Clusters
     kubernetes.core.k8s_info:
       api_version: apiextensions.k8s.io/v1
       kind: CustomResourceDefinition
       name: metal3clusters.infrastructure.cluster.x-k8s.io
       kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
     register: crd
-    failed_when: '"upgm3c" not in crd.resources[0].spec.names.shortNames'
+    failed_when: UPGRADED_CAPM3_VERSION not in (crd.resources[0].spec.versions | map(attribute='name'))
```
lentzi90, line 265: Now that we can verify the API resources for CAPM3, I think we could actually remove the "Verify upgraded CRDs for Metal3Clusters". This is how we used to do it [before the k8s module](https://github.com/metal3-io/metal3-dev-env/blob/56c32374c4c5713b2d272274e6df44de61aa2b94/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml#L207). Then we checked the CRDs for CAPI and API resources for CAPM3.

I would even suggest that we check only API resources for both CAPI and CAPM3 now that they actually change from v1alphaX to v1betaX so there is a change that we can detect. The reason for checking the CRDs (from what I understand) was that we made changes to them to *simulate* a change but still kept the "contract" version the same. By checking the API version instead of the CRD we would reduce the spam in the log quite a lot (the CRD is 1000s of lines, the API resources are just ~10).
WDYT? I could do a follow up PR with this later to stick to the point here.

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -67,6 +77,39 @@
...
+    ansible.builtin.command: "{{ item }}"
+    args:
+      chdir: "{{ CAPM3PATH }}"
+    loop:
+      - make hack/tools/bin/kustomize
+      - ./hack/tools/bin/kustomize build "{{ IPAMPATH }}/config/default" --output config/ipam/metal3-ipam-components.yaml
+
+  # TODO: update capm3 images (before making manifests)
+  # If we want to use the local registry and/or proxy we must make changes to the
+  # kustomizations before making the manifests.
```
lentzi90, line 90: Note this!
I'm not sure what to do about the images. Previously, when we just updated a patch version, I think the same image was reused (so not really an upgrade to the images, just to the CRDs). If the images actually were updated we used the upstream directly (not the local registry or proxy).

In 03_launch_management_cluster.sh we change the images [like this](https://github.com/metal3-io/metal3-dev-env/blob/620b4e6abef8c0edaedd8dc3cdab2a5b3d0da7bd/03_launch_mgmt_cluster.sh#L324-L351). But we do not have any variables for what images to use after the upgrade so it would take a bit of work to set it up. Don't get me wrong, I think it would be good to do it, I'm just not sure if it is worth to do at this point. The upgrade tests are not run that often so rate limits may not be that important. What do you think?

lentzi90 (commented):
Added some notes for things I'm not sure about.

scripts/feature_tests/upgrade/upgrade_vars.sh
```diff
@@ -6,11 +6,17 @@ set -eux
...
 export CAPI_REL_TO_VERSION="v1.0.1"
 
-export CAPM3RELEASE="v1.0.0"
+export CAPM3_VERSION="v1alpha5"
+export CAPM3RELEASE="v0.5.3"
 export CAPM3_REL_TO_VERSION="v1.0.0"
+export UPGRADED_CAPM3_VERSION="v1beta1"
+
+# Ubuntu is hard coded in the upgrade tests. Make sure we use it throughout.
+export IMAGE_OS="Ubuntu"
```
kashifest, line 19: Hard-coded? Why do we need this value to be set here? Shouldn't it be coming through jjb?

lentzi90: Yes jjb is setting it correctly, I was just thinking that we may as well hard code `IMAGE_OS` for upgrade tests since it is [hard coded here](https://github.com/metal3-io/metal3-dev-env/blob/620b4e6abef8c0edaedd8dc3cdab2a5b3d0da7bd/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml#L334). For the jjb/Jenkins jobs I guess it doesn't matter, but if you want to run them manually it will break if you forget to set this.

kashifest: Aha , makes sense now.

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
fmuyassarov: I'm not that much familiar with upgrade tests, but I believe you are correct that we were re-using the same image.

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -234,32 +243,26 @@
...
-  # we cannot use k8s_cluster_info. A fix is to be included in 2.0.0 of kubernetes.core
-  # - name: Verify upgraded API resource for Metal3Clusters
-  #   kubernetes.core.k8s_cluster_info:
-  #   register: api_status
-  #   # failed_when: api_status.apis not contains the upgraded m3c resource(s)?
+  - name: Verify upgraded API resources for CAPM3
+    kubernetes.core.k8s_cluster_info:
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: api_status
+    failed_when: "'infrastructure.cluster.x-k8s.io/v1beta1' not in api_status.apis"
```
fmuyassarov, line 256: is it possible to make `v1beta1` part as env var `UPGRADED_CAPM3_VERSION` or will it complicate things for future reader ?

fmuyassarov (commented):
/lgtm

scripts/feature_tests/upgrade/upgrade_vars.sh
```diff
@@ -6,11 +6,17 @@ set -eux
...
-export CAPIRELEASE="v1.0.0"
+export CAPI_VERSION="v1alpha4"
+export CAPIRELEASE="v0.4.5"
 export CAPI_REL_TO_VERSION="v1.0.1"
 
-export CAPM3RELEASE="v1.0.0"
+export CAPM3_VERSION="v1alpha5"
+export CAPM3RELEASE="v0.5.3"
 export CAPM3_REL_TO_VERSION="v1.0.0"
+export UPGRADED_CAPM3_VERSION="v1beta1"
```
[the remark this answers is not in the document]
lentzi90: I don't understand. This is similar to the [`CAPM3_VERSION` variable](https://github.com/metal3-io/metal3-dev-env/blob/620b4e6abef8c0edaedd8dc3cdab2a5b3d0da7bd/config_example.sh#L128). I guess we could work with just the `CAPIRELEASE` and `CAPM3RELEASE` variables and translate from that to `v1aX`/`v1bX`?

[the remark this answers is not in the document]
fmuyassarov: Sorry I misunderstood the variables

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
lentzi90: I have now added variables for the upgraded images and made the changes required to use them. It does not use the local registry but it does respect the `CONTAINER_REGISTRY` variable.

lentzi90:
Updated so that container images get proper upgraded versions and made the check for new API/CRDs cleaner.
PTAL!

lentzi90:
Now saving container image names and tags before and after upgrade in the `manifests` folder.
Most variables in upgrade_vars.sh are also overridable now. However, I think this will require some changes in other places to really work. Some variables are set already in jjb and jenkins e.g. [here](https://github.com/metal3-io/project-infra/blob/85c058a41ae3099af4208bff00f80441d8902986/jenkins/jobs/integration_tests.pipeline#L57). Would appreciate some input on this, like what/where should we change first?

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -1,5 +1,12 @@
 ---
 
+  - name: Fetch container images before upgrade
+    shell: |
+      kubectl get pods -A -o jsonpath="{range .items[*].spec.containers[*]}{.image}{'\n'}{end}" |
+      sort | uniq > /tmp/manifests/container_images_before_upgrade.txt
```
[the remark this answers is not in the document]
lentzi90: It is collected together with the manifests and logs so you can find it in the tarball in Jenkins.

vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml
```diff
@@ -127,62 +186,29 @@
...
-        regexp: 'description: KubeadmConfig'
-        replace: "description: upgradedKubeadmConfig"
-      - path: "{{working_dir}}/dev-repository//control-plane-kubeadm/{{CAPI_REL_TO_VERSION}}/control-plane-components.yaml"
-        regexp: 'description: KubeadmControlPlane'
-        replace: "description: upgradedKubeadmControlPlane"
-      - path: "{{working_dir}}/overrides/infrastructure-metal3/{{CAPM3_REL_TO_VERSION}}/infrastructure-components.yaml"
-        regexp: '\bm3c\b'
-        replace: "upgm3c"
+    - src: "{{ dev_repository }}/cluster-api/{{CAPIRELEASE}}/core-components.yaml"
+      dest: "{{ dev_repository }}/cluster-api/{{CAPI_REL_TO_VERSION}}/core-components.yaml"
```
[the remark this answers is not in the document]
lentzi90: This is actually exactly the same as before, I just changed the syntax slightly when I updated the CAPM3 part (last 2 items in the loop).
[Here is how it looked before](https://github.com/metal3-io/metal3-dev-env/blob/620b4e6abef8c0edaedd8dc3cdab2a5b3d0da7bd/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml#L128-L164).

So what I wanted to change here was just to make CAPM3 use the new rendered manifests. That is, change [this](https://github.com/metal3-io/metal3-dev-env/blob/620b4e6abef8c0edaedd8dc3cdab2a5b3d0da7bd/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml#L157-L164) to [this](https://github.com/Nordix/metal3-dev-env/blob/a013d4e3776743616c04bf1994e507f213f2881f/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml#L208-L211). The rest is the same as before.

I didn't quite understand why we do this part for CAPI, but I noticed that we are not really copying the old `CAPIRELEASE` to `CAPI_REL_TO_VERSION`. What we do is, first we copy `CAPIRELEASE_HARDCODED` to `CAPIRELEASE`, then we copy `CAPIRELEASE` to `CAPI_REL_TO_VERSION`. See [this](https://github.com/metal3-io/metal3-dev-env/blob/620b4e6abef8c0edaedd8dc3cdab2a5b3d0da7bd/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml#L87-L115) and then a few lines later [this](https://github.com/metal3-io/metal3-dev-env/blob/master/vm-setup/roles/v1aX_integration_test/tasks/upgrade.yml#L128). I *think* the idea is that we want the latest CAPI, which is the `CAPIRELEASE_HARDCODED`, but I don't understand why we copy it 2 times...

kashifest (approved):

PR #899, opened 2022-01-11
Add ironic inspector to image vars

lentzi90:
This is to make it use the local registry or proxy. The other images are
already defined in this way, the inspector was just missing from the
list.


lentzi90:
Replaced by https://github.com/metal3-io/baremetal-operator/pull/1066

PR #902, opened 2022-01-12
Handle zero length dhcp_range variables

hardys:
In this case, we treat it the same as the lists being undefined since
it implies DHCP is disabled.

This means it's possible to do a conditional in a vars file e.g:
```
dhcp_range_v4: "{{ [ <IP>, <IP>] if lookup('env', 'STATIC_IPS') == '' else [] }}"
```
Which is more convenient that conditionally generating the map keys.


hardys:
Note this is an alternative to https://github.com/metal3-io/metal3-dev-env/pull/879 - we can abandon that PR if this one merges

furkatgofurov7 (approved):
/approve

PR #903, opened 2022-01-13
Add the new epel-next repository for CentOS Stream

elfosardo:
EPEL Next is an additional repository that contains packages built
specifically against CentOS Stream to allow version requirements for
software already present in CentOS Stream but not in RHEL.
For more info see [1]

[1] https://docs.fedoraproject.org/en-US/epel/epel-about-next/

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -21,9 +21,10 @@ if [[ $OS == ubuntu ]]; then
   elif [[ ${DISTRO} == "ubuntu20" ]]; then
     sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.8 1
   fi
-
-else
-  sudo dnf update -y && sudo dnf install -y epel-release
+elif [[ $OS == "centos" || $OS == "rhel" ]]; then
+  sudo dnf upgrade -y
+  sudo dnf config-manager --set-enabled powertools
+  sudo dnf install -y epel-release epel-next-release
   sudo dnf -y install python3-pip
   sudo alternatives --set python /usr/bin/python3
 fi


01_prepare_host.sh
```diff
@@ -21,9 +21,10 @@ if [[ $OS == ubuntu ]]; then
   elif [[ ${DISTRO} == "ubuntu20" ]]; then
     sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.8 1
   fi
-
-else
-  sudo dnf update -y && sudo dnf install -y epel-release
+elif [[ $OS == centos ]]; then
+  sudo dnf upgrade -y
+  sudo dnf config-manager --set-enabled powertools
+  sudo dnf install -y epel-release epel-next-release
```
furkatgofurov7, line 27: @elfosardo these changes LGTM, but could you share a bit more on why we would need to add the new epel-next repository for CentOS, on the PR descriptions if possible. Thanks!

elfosardo: done

furkatgofurov7 (approved):

01_prepare_host.sh
```diff
@@ -21,9 +21,10 @@ if [[ $OS == ubuntu ]]; then
   elif [[ ${DISTRO} == "ubuntu20" ]]; then
     sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.8 1
   fi
-
-else
-  sudo dnf update -y && sudo dnf install -y epel-release
+elif [[ $OS == centos ]]; then
```
dtantsur, line 24: This used to apply to RHEL as well, now we  have no logic for RHEL?

elfosardo:
@dtantsur can you have another look please ?

furkatgofurov7 (commented):
/lgtm

PR #905, opened 2022-01-14
Ensure that machines are provisioned and running before repivot

lentzi90:
Since we are doing this check in multiple places, I decided to factor
out the steps into a separate file that can be included where needed.
When doing this we must ensure that the correct kubeconfig is used by
passing it as a variable.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
@@ -2,7 +2,6 @@
   - set_fact:
       NUMBER_OF_BMH_KCP: "{{ NUM_OF_CONTROLPLANE_REPLICAS|int }}"
       NUMBER_OF_BMH_MD: "{{ NUM_OF_WORKER_REPLICAS|int }}"
-      NUMBER_OF_ALL_BMH: "{{ NUM_OF_WORKER_REPLICAS|int + NUM_OF_CONTROLPLANE_REPLICAS|int }}"
 
   - name: Untaint all nodes.
     shell: |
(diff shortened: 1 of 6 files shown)


vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
```diff
@@ -441,28 +441,7 @@
...
+  - name: Verify that all machines are provisioned and running.
+    include_tasks: verify_provisioned.yml
+    vars:
       kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
-    register: machines
-    retries: 200
-    delay: 20
-    until:
-      - machines is succeeded
-      - machines.resources | filter_phase("running") | length == (NUMBER_OF_ALL_BMH | int)
```
furkatgofurov7, line 468: **NUMBER_OF_ALL_BMH** defined at the beginning of this file can be safely removed I believe, since it is covered by a more consistent var called **NUMBER_OF_BMH**?

vm-setup/roles/v1aX_integration_test/tasks/verify_provisioned.yml
```diff
@@ -0,0 +1,63 @@
+---
```
furkatgofurov7, line 1: Hmm, this can be a bit confusing, since we are checking different resources for different states:

- cluster ==> "provisioned"
- bmh ==> "provisioned"
- m3m ==> "ready"
- machine ==> "runnning"

While I do not have an ideal naming convention for this file, maybe we could rename it to "verify_desired_state.yml" or "verify_resources_states" depnding how do you feel about it.

furkatgofurov7 (commented):
Overall looks like a good optimization, thanks! Few suggestions inline:

vm-setup/roles/v1aX_integration_test/tasks/node_reuse.yml
lentzi90: Yes I think you are correct. Not sure why we have both. I'll update!

vm-setup/roles/v1aX_integration_test/tasks/verify_provisioned.yml
lentzi90: True, I like `verify_resources_states.yml`. Fixing...

furkatgofurov7 (approved):

PR #910, opened 2022-01-20
🌱 Add logic to verify passwordless sudo

Rozzii:
This fix is implemented because on a freshly cloned or fully cleaned metal3-dev-env
it is possible to bootstrap a cluster without the passwordless sudo
configuration but at subsequent executions the script will stop running
and one of the components will throw an error to notify the user that it needs
passwordless sudo.

In order to avoid running the dev-env with incorrect sudo configuration this
commit implements a verification step that will stop the execution in case
the sudo configuratio is incorrect.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -2,6 +2,12 @@
 
 [[ ":$PATH:" != *":/usr/local/go/bin:"* ]] && PATH="$PATH:/usr/local/go/bin"
 
+# Verify that passwordless sudo is configured correctly
+if ! sudo -nl | grep -q '(ALL) NOPASSWD: ALL';then
+    echo "ERROR: metal3-dev-env requires passwordless sudo configuration!"
+    exit 1
+fi
+
 eval "$(go env)"
 export GOPATH
 


lib/common.sh
```diff
@@ -2,6 +2,12 @@
 
 [[ ":$PATH:" != *":/usr/local/go/bin:"* ]] && PATH="$PATH:/usr/local/go/bin"
 
+# Verify that passwordless sudo is configured correctly
+if ! sudo -l | grep -q '(ALL) NOPASSWD: ALL';then
```
macaptain, line 6: We need `sudo -nl`, otherwise this prompt will hang if the user doesn't have passwordless sudo permissions.

macaptain (changes requested):

lib/common.sh
Rozzii: Thanks, fixed.

furkatgofurov7 (approved):
/retitle 🌱 Add logic to verify passwordless sudo

furkatgofurov7:
/retitle 🌱 Add logic to verify passwordless sudo

PR #911, opened 2022-01-21
🐛 Fix BMO and Ironic image updates

### LINKED ISSUE #908: Image replacement for BMO and Ironic not working
The function for [updating the BMO and Ironic images](https://github.com/metal3-io/metal3-dev-env/blob/c9f982a50116fe9724992381265eb27f1d7789f8/03_launch_mgmt_cluster.sh#L72) is not working properly.

With the [default value](https://github.com/metal3-io/metal3-dev-env/blob/c9f982a50116fe9724992381265eb27f1d7789f8/lib/common.sh#L205), it successfully updates the patch to use the image in the local registry. However, if another value is used, e.g. `quay.io/metal3-io/baremetal-operator:capm3-v0.5.4`, no replacement is done.

The log message hints at the issue:

```
Image quay.io/metal3-io/baremetal-operator:capm3-v0.5.4 will be replaced with local image 192.168.111.1:5000/localimages/baremetal-operator:capm3-v0.5.4 in image kustomization
```

It is assuming that the image specified in `BAREMETAL_OPERATOR_IMAGE` is used in the kustomization and can be substituted directly. If the image doesn't match, no substitution will be made.

**Suggested solution:**

Take a look at how this is [done for CAPM3](https://github.com/metal3-io/metal3-dev-env/blob/c9f982a50116fe9724992381265eb27f1d7789f8/03_launch_mgmt_cluster.sh#L324) and do it in a similar way for BMO and Ironic.

lentzi90:
Fixes #908. Needs https://github.com/metal3-io/baremetal-operator/pull/1074 before it can work.

The function used to update the container images for BMO and Ironic only
worked for replacing the original image with the one in the local
registry. (See #908. ) This fix makes it possible to use a different images and more
importantly tags, so that we can start BMO with a specific tag and then
upgade to latest.

This commit also unifies the image replacement so that it works similar
for CAPM3, BMO and Ironic.

Finally, backup and restore of some files before and after changes has
been removed. The reason for this is twofold:

1. All changed files are anyway restored every time the
03_launch_mgmt_cluster.sh script is run, since it does a fresh clone
every time, so manually doing these backups and restorations is not
necessary.
2. Restoring the files directly when they have been used hides the true
state for the user and makes it harder to understand what is happening.


lentzi90:
/hold
This cannot be merged before https://github.com/metal3-io/baremetal-operator/pull/1074

03_launch_mgmt_cluster.sh
```diff
@@ -225,25 +184,18 @@ EOF
...
+    update_component_image Keepalived "${IRONIC_KEEPALIVED_LOCAL_IMAGE:-$IRONIC_KEEPALIVED_IMAGE}"
+    update_component_image IPA-downloader "${IPA_DOWNLOADER_LOCAL_IMAGE:-$IPA_DOWNLOADER_IMAGE}"
 
     # Copy the generated configmap for ironic deployment
     cp "$IRONIC_DATA_DIR/ironic_bmo_configmap.env"  "${BMOPATH}/ironic-deployment/keepalived/ironic_bmo_configmap.env"
 
     # Deploy. Args: <deploy-BMO> <deploy-Ironic> <deploy-TLS> <deploy-Basic-Auth> <deploy-Keepalived>
     "${BMOPATH}/tools/deploy.sh" false true "${IRONIC_TLS_SETUP}" "${IRONIC_BASIC_AUTH}" true
-
-    # Restore original files
```
Xenwar, line 240: We are no longer restoring original files ?
is it so that re-running deployment script will override it anyway or there is another step to do the restoration.

lentzi90: Yes, I removed this because of 2 reasons:
1. In [03_launch_mgmt_cluster.sh we remove and clone again](https://github.com/metal3-io/metal3-dev-env/blob/master/03_launch_mgmt_cluster.sh#L38-L40) every time unless `FORCE_REPO_UPDATE` is set to false (default is true). So the cleaning will happen anyway the next time we run it.
2. I think it is easier to understand what is happening if we don't restore it immediately. It was very confusing to me at first that I could check in the manifests and see one thing and then in the cluster it would be something different. Now we will be able to see what modifications were made more easily. (And it is easy to restore manually also with `git reset --hard`.)

Maybe a third reason would also be that we sometimes need these changes also after the script is done. For example when pivoting. It becomes hard then to understand what is going on when the files are changed in the script and then restored in the playbook.

Xenwar: Got it, thanks.

Xenwar:
approved: when tests are passed (pending another PR)

smoshiur1237 (approved):
lgtm

lentzi90:
> do we need to run feature and upgrade tests as well. otherwise lgtm

Good point! At least upgrade tests could be affected.

/test-features-upgrade

03_launch_mgmt_cluster.sh
```diff
@@ -319,7 +266,7 @@ function update_capm3_imports(){
 }
 
 #
-# Update the images for the CAPM3 deployment file to use local ones
+# Update the images for CAPM3 and BMO deployment file to use local ones
 #
 function update_component_image(){
```
mboukhalfa, line 271: since what this function is changing the registry not the image, should not we change the name of this function to something like `set_localregistry()` so it is more about registry not image

03_launch_mgmt_cluster.sh
```diff
@@ -319,7 +266,7 @@ function update_capm3_imports(){
 }
 
 #
-# Update the images for the CAPM3 deployment file to use local ones
+# Update the images for CAPM3 and BMO deployment file to use local ones
```
mboukhalfa, line 269: # Update CAPM3 and BMO manifest to use the the local registry for images

03_launch_mgmt_cluster.sh
```diff
@@ -319,7 +266,7 @@ function update_capm3_imports(){
...
 #
 function update_component_image(){
   IMPORT=$1
```
mboukhalfa, line 272: I would suggest to use COMPONENT instead and add a comment

03_launch_mgmt_cluster.sh
```diff
@@ -333,21 +280,51 @@ function update_component_image(){
     TMP_IMAGE_TAG="latest"
   fi
```
mboukhalfa, line 282: What if we used another function `extract_image_name_and_tag`

03_launch_mgmt_cluster.sh
lentzi90: You are correct in that it used to change only the registry. But this PR makes it actually change everything, so we can set a new image tag if we want. This will be used later to do upgrade tests for BMO.

03_launch_mgmt_cluster.sh
lentzi90: It is not just for registry. For example, we use this to get an old release of CAPM3 for upgrade tests.

03_launch_mgmt_cluster.sh
lentzi90: Sure I agree that would make sense. I just didn't want to change more things than I had to. It should be easier to see what actually changed in this way. But I can do a follow up where I change the names of things.

03_launch_mgmt_cluster.sh
lentzi90: Same as above, I didn't want to change more than necessary since it makes it harder to understand what actually changed.

03_launch_mgmt_cluster.sh
```diff
@@ -66,55 +66,17 @@ function clone_repos() {
...
-# Create the BMO deployment (used for v1a4 only)
+# Create the BMO deployment (not used for CAPM3 v1a4 since BMO is bundeled there)
 #
 function launch_baremetal_operator() {
   pushd "${BMOPATH}"
 
   # Deploy BMO using deploy.sh script
 
   # Update container images to use local ones
-  cp "${BMOPATH}/config/manager/manager.yaml" "${BMOPATH}/config/manager/manager.yaml.orig"
```
kashifest, line 113: why are we  removing this and lines similar to these ? The idea was to leave the repos intact aftern the image manipulations we are doing.

03_launch_mgmt_cluster.sh
```diff
@@ -172,8 +134,6 @@ function update_images(){
     IMAGE_NAME="${IMAGE##*/}"
     LOCAL_IMAGE="${REGISTRY}/localimages/$IMAGE_NAME"
     OLD_IMAGE_VAR="${IMAGE_VAR%_LOCAL_IMAGE}_IMAGE"
-    # Strip the tag for image replacement
-    OLD_IMAGE="${!OLD_IMAGE_VAR%:*}"
```
kashifest, line 176: Not clear why are we removing this?

03_launch_mgmt_cluster.sh
```diff
@@ -225,25 +184,18 @@ EOF
...
 
-    # Update container images to use local ones
-    cp "${BMOPATH}/ironic-deployment/ironic/ironic.yaml" "${BMOPATH}/ironic-deployment/ironic/ironic.yaml.orig"
-    cp "${BMOPATH}/ironic-deployment/keepalived/keepalived_patch.yaml" "${BMOPATH}/ironic-deployment/keepalived/keepalived_patch.yaml.orig"
-    update_kustomization_images "${BMOPATH}/ironic-deployment/ironic/ironic.yaml"
-    update_kustomization_images "${BMOPATH}/ironic-deployment/keepalived/keepalived_patch.yaml"
+    update_component_image Ironic "${IRONIC_LOCAL_IMAGE:-$IRONIC_IMAGE}"
+    update_component_image Mariadb "${MARIADB_LOCAL_IMAGE:-$MARIADB_IMAGE}"
+    update_component_image Keepalived "${IRONIC_KEEPALIVED_LOCAL_IMAGE:-$IRONIC_KEEPALIVED_IMAGE}"
+    update_component_image IPA-downloader "${IPA_DOWNLOADER_LOCAL_IMAGE:-$IPA_DOWNLOADER_IMAGE}"
```
kashifest, line 190: I dont think assigning the vars in a function call is a good idea. Would you please fix the value of the image before calling the fucntion.

03_launch_mgmt_cluster.sh
lentzi90: Basically I think it is unnecessary and makes it harder to understand what is happening. I added in the description this:

Backup and restore of some files before and after changes has
been removed. The reason for this is twofold:

1. All changed files are anyway restored every time the
    03_launch_mgmt_cluster.sh script is run, since it does a fresh clone
    every time, so manually doing these backups and restorations is not
    necessary.
2. Restoring the files directly when they have been used hides the true
    state for the user and makes it harder to understand what is happening.

If we still want to restore the repo I would much prefer to do it with `git reset --hard` at the end to avoid any mistakes.

03_launch_mgmt_cluster.sh
lentzi90: This was used to avoid touching the tag of the image. Now I want to replace also the tag so there is no point in stripping it.
So previously it did something like this:
```
my-registry.org/some/image:tag -> local.registry/localimages/some/image:tag
```
Now it will replace all of it (based on variables like `BARE_METAL_OPERATOR_IMAGE`):
```
my-registry.org/some/image:tag -> local.registry/localimages/some/custom-image:custom-tag
```

03_launch_mgmt_cluster.sh
lentzi90: I'm not sure I understand. Something like this?
```bash
TMP_IMAGE="${IRONIC_LOCAL_IMAGE:-$IRONIC_IMAGE}"
update_component_image Ironic "${TMP_IMAGE}"
```
Or maybe with if/else?
```bash
if [ -n "${CAPM3_LOCAL_IMAGE}" ]; then
  update_component_image CAPM3 "${CAPM3_LOCAL_IMAGE}"
else
  update_component_image CAPM3 "${CAPM3_IMAGE}"
fi
```
I actually had this if/else solution first but thought it was a bit cluttered. But I can put it back if you think it is better?

03_launch_mgmt_cluster.sh
kashifest: Well, if you want to know what image is the current deployment running, you should see the deployment in cluster, not the original files. Secondly, this change is not related to the fix you are doing, so I would rather keep it as it is for now and see if your changes are working properly or not first. We can take this part later if necessary IMO.

03_launch_mgmt_cluster.sh
kashifest: ok I got it. It looks fine then

03_launch_mgmt_cluster.sh
kashifest: For the sake of clarity yes please.

lentzi90: Done. Note that the files needed to backup/restore changed since we use the new make targets in BMO now. I think it also shows why it would be good to change this: it is a dependency between metal3-dev-env and BMO/CAPM3. I'll create a follow up PR for fixing this.

03_launch_mgmt_cluster.sh
lentzi90: Done

03_launch_mgmt_cluster.sh
lentzi90: Updated

kashifest (approved):

03_launch_mgmt_cluster.sh
kashifest: Didn't get the later part of the comment. Can you elaborate?

lentzi90: Basically we edit files in BMO/CAPM3 from metal3-dev-env, so there is a direct dependency between them. Like if we for example move or rename one of the deployment files in BMO, it would break these scripts since they depend on those files directly. Of course there will always be some relation between them, like using make targets, but this is a bit too much in my opinion and I have some ideas for how to reduce it.

kashifest: Allright, not sure if I got it fully but perhaps once you have a PR, it will be more visible . Good work in this one.

03_launch_mgmt_cluster.sh
```diff
@@ -225,11 +191,29 @@ EOF
...
-    cp "${BMOPATH}/ironic-deployment/keepalived/keepalived_patch.yaml" "${BMOPATH}/ironic-deployment/keepalived/keepalived_patch.yaml.orig"
-    update_kustomization_images "${BMOPATH}/ironic-deployment/ironic/ironic.yaml"
-    update_kustomization_images "${BMOPATH}/ironic-deployment/keepalived/keepalived_patch.yaml"
+    cp "${BMOPATH}/ironic-deployment/default/kustomization.yaml" "${BMOPATH}/ironic-deployment/default/kustomization.yaml.orig"
+    cp "${BMOPATH}/ironic-deployment/keepalived/kustomization.yaml" "${BMOPATH}/ironic-deployment/keepalived/kustomization.yaml.orig"
+
+    if [ -n "${IRONIC_LOCAL_IMAGE}" ]; then
+      update_component_image Ironic "${IRONIC_LOCAL_IMAGE}"
+    else
+      update_component_image Ironic "${IRONIC_IMAGE}"
```
fmuyassarov, line 200: do we really needs else part? I assume if we drop it, we will anyways use the latest images or the image as defaulted in `IRONIC_IMAGE`?

lentzi90: We need both. The `*_LOCAL_IMAGE` variables are for images that are built locally while the other are for "the rest". In CI we use them to set the registry for example so we can use nordix harbor instead of docker hub or quay. For some tests we also want specific tags, like when upgrading from an older version.

lentzi90:
(Seem to be hitting deprovisioning issue)
/test-features-centos

03_launch_mgmt_cluster.sh
```diff
@@ -238,8 +222,8 @@ EOF
     "${BMOPATH}/tools/deploy.sh" false true "${IRONIC_TLS_SETUP}" "${IRONIC_BASIC_AUTH}" true
 
     # Restore original files
-    mv "${BMOPATH}/ironic-deployment/ironic/ironic.yaml.orig" "${BMOPATH}/ironic-deployment/ironic/ironic.yaml"
-    mv "${BMOPATH}/ironic-deployment/keepalived/keepalived_patch.yaml.orig" "${BMOPATH}/ironic-deployment/keepalived/keepalived_patch.yaml"
+    mv "${BMOPATH}/ironic-deployment/default/kustomization.yaml.orig" "${BMOPATH}/ironic-deployment/default/kustomization.yaml"
+    mv "${BMOPATH}/ironic-deployment/keepalived/kustomization.yaml.orig" "${BMOPATH}/ironic-deployment/keepalived/kustomization.yaml"
```
furkatgofurov7, line 226: Trying to understand why we are moving root kustomization.yaml instead of a specific file?

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -135,7 +135,7 @@
       chdir: "{{ BMOPATH }}"
 
   - name: Reinstate Baremetal Operator Manager
-    shell: "mv {{ BMOPATH }}/config/manager/manager.yaml.orig {{ BMOPATH }}/config/manager/manager.yaml"
+    shell: "mv {{ BMOPATH }}/config/default/kustomization.yaml.orig {{ BMOPATH }}/config/default/kustomization.yaml"
```
furkatgofurov7, line 138: Same question, why we are moving root kustomization.yaml instead of a specific file?

furkatgofurov7 (commented):
Let's trigger v1alpha4 tests as well since you have already run almost all available tests we have :)

/test-v1a4-centos-integration
/test-v1a4-ubuntu-integration

03_launch_mgmt_cluster.sh
lentzi90: It is because the [make targets](https://github.com/metal3-io/baremetal-operator/blob/main/Makefile#L156-L179) we use now (instead of `sed` on the specific files). They use `kustomize edit set image`, which changes the `kustomization.yaml` by adding something like this:
```yaml
images:
- name: name-of-image-in-manifest
  newName: new-image-name
  newTag: x.y
```

fmuyassarov:
/hold
@furkatgofurov7 I leave to you to remove the hold since you had a question

PR #912, opened 2022-01-24
Use the virt-rhel8.2.0 machine type in case of arm64

bn222:
The q35 machine isn't valid for arm64. Instead, use virt-rhel8.2.0. Also
rely on automatic UEFI firmware and secure boot settings instead of
manually configuring this which would need to be done in an architecure
specific manner.

Signed-off-by: Balazs Nemeth <bnemeth@redhat.com>

### DIFF
--- a/vm-setup/roles/libvirt/templates/baremetalvm.xml.j2
+++ b/vm-setup/roles/libvirt/templates/baremetalvm.xml.j2
@@ -5,6 +5,7 @@
 
   {{baremetal_vm_xml|default('')}}
 
+{% if libvirt_arch != 'aarch64' %}
   <os>
     <type arch='{{ libvirt_arch }}' machine='q35'>hvm</type>
 {% if libvirt_firmware  == 'uefi' %}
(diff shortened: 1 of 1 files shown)


furkatgofurov7 (commented):
LGTM
/cc @hardys
This is not a big change but still can you please take a look? Thanks

vm-setup/roles/libvirt/templates/baremetalvm.xml.j2
```diff
@@ -18,6 +19,12 @@
     <boot dev='network'/>
     <bootmenu enable='no'/>
   </os>
+{% else %}
+<os firmware='efi'>
+  <type arch='aarch64' machine='virt-rhel8.2.0'>hvm</type>
```
dtantsur, line 24: Are sure you want to disable the firmware configuration from line 11?

vm-setup/roles/libvirt/templates/baremetalvm.xml.j2
```diff
@@ -18,6 +19,12 @@
     <boot dev='network'/>
     <bootmenu enable='no'/>
   </os>
+{% else %}
+<os firmware='efi'>
```
dtantsur, line 23: We can probably use it for all architectures now that we no longer support RHEL 7.

dtantsur (approved):

hardys:
/lgtm 

Sorry for the slow response @furkatgofurov7 - this looks reasonable to me, and it seems like the question of always using uefi firmware raised by @dtantsur can be considered in a follow-up

PR #927, opened 2022-02-02
Temporary fix for cert-manager URL change

lentzi90:
Cert-manager has changed the organization on github which breaks the URL.

This temporary fix is based on the suggestion on CAPI slack here: https://kubernetes.slack.com/archives/C8TSNPY4T/p1643801670063279

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/templates/clusterctl-upgrade-test.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/clusterctl-upgrade-test.yaml
@@ -1,5 +1,10 @@
 # Cluster API config used for upgrade testing.
 
+## Cert-manager has changed the organization on github.
+## This is a temporary fix until CAPI has updated the URL.
+cert-manager:
+  url: "https://github.com/cert-manager/cert-manager/releases/latest/cert-manager.yaml"
+
 # Image is overriden in provider/container_name format
 # list of providers
 # cluster.x-k8s.io/provider: bootstrap-kubeadm
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -358,6 +358,13 @@ function patch_clusterctl(){
   mkdir -p "${HOME}"/.cluster-api
   touch "${HOME}"/.cluster-api/clusterctl.yaml
 
+  ## Cert-manager has changed the organization on github.
+  ## This is a temporary fix until CAPI has updated the URL.
+  cat <<EOF > "${HOME}"/.cluster-api/clusterctl.yaml
+cert-manager:
+  url: "https://github.com/cert-manager/cert-manager/releases/latest/cert-manager.yaml"
+EOF
+
   # At this point the images variables have been updated with update_images
   # Reflect the change in components files
   if [ -n "${CAPM3_LOCAL_IMAGE}" ]; then
--- a/vm-setup/roles/v1aX_integration_test/templates/clusterctl-vars.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/clusterctl-vars.yaml
@@ -2,6 +2,11 @@
 # For deployment to ~/.cluster-api/
 ## Note these settings are overridden by environment variables.
 
+## Cert-manager has changed the organization on github.
+## This is a temporary fix until CAPI has updated the URL.
+cert-manager:
+  url: "https://github.com/cert-manager/cert-manager/releases/latest/cert-manager.yaml"
+
 # Cluster settings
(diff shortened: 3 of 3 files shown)


kashifest:
/approved
Thanks @lentzi90 for the quick fix.

kashifest (approved):

vm-setup/roles/v1aX_integration_test/templates/clusterctl-upgrade-test.yaml
```diff
@@ -1,5 +1,10 @@
 # Cluster API config used for upgrade testing.
 
+## Cert-manager has changed the organization on github.
+## This is a temporary fix until CAPI has updated the URL.
+cert-manager:
+  url: "https://github.com/cert-manager/cert-manager/releases/latest/cert-manager.yaml"
+
```
furkatgofurov7, line 7: Since this is touching upgrade tests, triggering it to be on the safe side.

furkatgofurov7 (commented):
Looks good to me overall

vm-setup/roles/v1aX_integration_test/templates/clusterctl-upgrade-test.yaml
lentzi90: Thanks! I think there could be an issue with the upgrades now that I think about it. Not in this file, but in `03_launch_mgmt_cluster.sh`.

In the upgrade tests we are starting from CAPM3 v1alpha5 and I'm guessing we should not be using `latest/cert-manager.yaml` for that? If that is the case we will need to add some logic in `03_launch_mgmt_cluster.sh` so that it only adds this workaround for v1beta1.

What do you think?

furkatgofurov7: We have never pinned cert-manager AFAIK regardless of API version so far, but now I am also thinking that, I think specifying latest is not the best way ( based on the [conversion](https://kubernetes.slack.com/archives/C8TSNPY4T/p1643804703827699?thread_ts=1643801670.063279&cid=C8TSNPY4T) from slack cert-manager maintainer) and we have to pin it to be specific i.e use v1.7.0 instead of latest

furkatgofurov7: Sorry, I was a bit wrong, so we use cert-manager v1.5.3 which is the same with CAPI. There is a discussion going in CAPI side to switch to v1.7.0 or potentially wait for v1.7.1, ref: https://github.com/kubernetes-sigs/cluster-api/issues/6055#issuecomment-1027962331

kashifest: Cert-manager version comes with clusterctl. Lets not pin it and use whatever is fixed for a particular release, i.e. let clusterctl decide it.

lentzi90: Does this mean we should change the workaround to use v1.5.3 instead of latest? I mean we basically now override what clusterctl is providing with this workaround. Seems to be working fine though. :thinking:

kashifest: Ow true. Yes. We should use 1.5.3 or whatever that version of clusterctl is supposed to use.

kashifest: Hummm. This is not as simple as it seemed before. If we pin it without checking which CAPM3 version we are running the test for, then a newer version of cert manager might fail for older version of CAPI/CAPM3. i.e. 1.5.3 might not pass v1a4 tests. We need to run all the tests here.

furkatgofurov7: We can use v1.5.3 as a workaround for now (that is the version we were getting from clusterctl before it is broken) and remove these workaround once CAPI fixes it upstream.

kashifest: @lentzi90 lets hold the idea of pinning, lets see how v1a4 test react now that I have trig erred. Since this is temporary solution, I believe it should be ok to go for latest.

kashifest: I have changed my mind, I will go with what @furkatgofurov7 is suggesting after reading this https://kubernetes.slack.com/archives/C4NV3DWUC/p1643793752687769, I think pinning would be a better option.

PR #929, opened 2022-02-02
Use latest CAPI release v1.1.0

furkatgofurov7:

PR #930, opened 2022-02-03
Revert "Temporary fix for cert-manager URL change"

lentzi90:
Reverts metal3-io/metal3-dev-env#927

There are new releases of CAPI now with a fix so we should be able to revert this.


lentzi90:
Needs #929 I think.

kashifest (approved):
/lgtm
Taking this in to make the CI work with proper cert-manager versions.

PR #936, opened 2022-02-05
Add install ca-certificates before update

kashifest:
Ubuntu based CI is failing with  `Certificate verification failed: The certificate is NOT trusted. The received OCSP status response is invalid.  Could not handshake: Error in the certificate verification.`

This PR is adding the installation of ca-certificate


kashifest:
Doesn't seem to solve the problem. Closing it.
/close

PR #941, opened 2022-02-09
Always chown the working dir to the current user

macaptain:
If the working dir (/opt/metal3-dev-env) is created, but owned by another user, we hit a failure later when trying to create the certs directory.


fmuyassarov (approved):

furkatgofurov7 (commented):
/lgtm

PR #942, opened 2022-02-09
Fix chmod to run with sudo

macaptain:
The directory is created as root, so we need to be root to change the permissions on it at this point in the script.

Found this bug while working from a fresh environment.


smoshiur1237 (approved):
/lgtm

kashifest (approved):

PR #945, opened 2022-02-10
Make IMAGE_OS value lower case

lentzi90:
IMAGE_OS used to take values like "Ubuntu" or "Centos". This commit
changes it to "ubuntu" or "centos". This makes the variable useful in
places were it would be awkward to have a capitalized value. It will
allow us to remove some logic and extra variables from CI since we can
use IMAGE_OS directly.

Note that both capitalized and lower case values will be supported while
we work on changing the CI to use lower case values.

### DIFF
--- a/lib/images.sh
+++ b/lib/images.sh
@@ -1,8 +1,11 @@
 #!/bin/bash
 
 # Image name and location
-IMAGE_OS=${IMAGE_OS:-Centos}
-if [[ "${IMAGE_OS}" == "Ubuntu" ]]; then
+IMAGE_OS=${IMAGE_OS:-centos}
+# NOTE: ${VAR,,} is making the varable lower case.
+# This is temporary and will be removed once we have changed IMAGE_OS to be
+# lower case in the CI.
+if [[ "${IMAGE_OS,,}" == "ubuntu" ]]; then
   export IMAGE_NAME=${IMAGE_NAME:-UBUNTU_20.04_NODE_IMAGE_K8S_${KUBERNETES_VERSION}.qcow2}
   export IMAGE_LOCATION=${IMAGE_LOCATION:-https://artifactory.nordix.org/artifactory/metal3/images/k8s_${KUBERNETES_VERSION}}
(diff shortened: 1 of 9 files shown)


lib/images.sh
```diff
@@ -1,8 +1,11 @@
...
-IMAGE_OS=${IMAGE_OS:-Centos}
-if [[ "${IMAGE_OS}" == "Ubuntu" ]]; then
+IMAGE_OS=${IMAGE_OS:-centos}
+# NOTE: ${VAR,,} is making the varable lower case.
+# This is temporary and will be removed once we have changed IMAGE_OS to be
+# lower case in the CI.
+if [[ "${IMAGE_OS,,}" == "ubuntu" ]]; then
   export IMAGE_NAME=${IMAGE_NAME:-UBUNTU_20.04_NODE_IMAGE_K8S_${KUBERNETES_VERSION}.qcow2}
   export IMAGE_LOCATION=${IMAGE_LOCATION:-https://artifactory.nordix.org/artifactory/metal3/images/k8s_${KUBERNETES_VERSION}}
 elif [[ "${IMAGE_OS}" == "FCOS" ]]; then
```
lentzi90, line 11: I'm not sure about these FCOS and FCOS-ISO. We do not use them in CI so we can keep them all upper case as they are acronyms.

fmuyassarov: lets keep it for now

lib/images.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/bash
 
 # Image name and location
-IMAGE_OS=${IMAGE_OS:-Centos}
-if [[ "${IMAGE_OS}" == "Ubuntu" ]]; then
+IMAGE_OS=${IMAGE_OS:-centos}
+# NOTE: ${VAR,,} is making the varable lower case.
```
fmuyassarov, line 5: nice

fmuyassarov (approved):

furkatgofurov7 (commented):
LGTM

PR #947, opened 2022-02-11
Uplift CAPM3 reference to v1.1.0

kashifest:


lentzi90 (commented):
/lgtm

scripts/feature_tests/upgrade/upgrade_vars.sh
```diff
@@ -10,7 +10,7 @@ export CAPI_VERSION="${CAPI_VERSION:-v1alpha4}"
 export CAPI_REL_TO_VERSION="v1.0.1"
```
furkatgofurov7, line 10: Just a note, we could also update this to the latest version in upgrade tests, but it can be done in a separate patch

furkatgofurov7 (approved):

kashifest:
> /test-v1a5-centos-integration /test-v1b1-integration

Not sure if the tests would pass, we should take CAPM3 changes first.

furkatgofurov7:
> Not sure if the tests would pass, we should take CAPM3 changes first.

Correct, sorry, let's merge CAPM3 one first and retrigger them again here.

kashifest:
> > Not sure if the tests would pass, we should take CAPM3 changes first.
> 
> Correct, sorry, let's merge CAPM3 one first and retrigger them again here.

Actually I might be wrong, we might need to take this one in before CAPM3 .

PR #949, opened 2022-02-16
For BMC_Driver set to mixed, configure extra workers using redfish

bfournie:
When the BMC_DRIVER is set to *mixed* for CI, any extraworkers should be configured using Redfish instead of IPMI.


elfosardo (approved):

furkatgofurov7 (approved):
@bfournie as agreed during the community meeting can you please link downstream PR to this patch (as I can already see it in history, would be great to have it in PR description with some more context on `extraworker`), thanks!

/approve

PR #954, opened 2022-02-22
🐛 Fix network interface commands for centos hosts

Rozzii:
The fix was provided by @lentzi90 !


furkatgofurov7 (approved):
Great work, well done folks!

@Rozzii can you update the PR description and mention the PR this patch is reverting, and also the reason for changes regarding `ifup`, thanks!

/approve

lentzi90:
Just want to mention that I found these new commands by checking what the `ifup` script in the older version of NetworkManager used to do. So they should be more or less equivalent to the old command.

That said, I also discovered another command that seems to work well: `nmcli device connect eth0`. This was discovered by just checking the nmcli docs. I think it does not really matter which one we use, so why not go with the one we have tested now. :slightly_smiling_face: 

Background for future readers wondering what this is all about:

The centos jobs started failing. The reason was a new version of NetworkManager that removed the ifup/ifdown scripts, which made the `preKubeadmCommands` fail, which in turn made Ironic's keepalived fail.

/lgtm

Issue #957, opened 2022-03-04
kubectl issues warnings after setup of demo cluster


jklippel:
After successful setup of the demo cluster using "make all", "./scripts/provision/cluster.sh", "./scripts/provision/controlplane.sh" and "./scripts/provision/worker.sh" it is possible to access the controlplane node using ssh and the well-known ip-address using `ssh -lmetal3 192.168.111.100`.

On issueing any kubectl command, warnings regarding client-side throttling are issued:
```
[metal3@test1-l6hrb ~]$ kubectl get all
I0304 09:36:43.289341    7044 request.go:665] Waited for 1.157933116s due to client-side throttling, not priority and fairness, request: GET:https://192.168.111.249:6443/apis/authorization.k8s.io/v1?timeout=32s
I0304 09:36:53.489301    7044 request.go:665] Waited for 5.398778618s due to client-side throttling, not priority and fairness, request: GET:https://192.168.111.249:6443/apis/node.k8s.io/v1beta1?timeout=32s
I0304 09:37:03.489571    7044 request.go:665] Waited for 2.594917684s due to client-side throttling, not priority and fairness, request: GET:https://192.168.111.249:6443/apis/authentication.k8s.io/v1?timeout=32s
I0304 09:37:14.889117    7044 request.go:665] Waited for 1.199191731s due to client-side throttling, not priority and fairness, request: GET:https://192.168.111.249:6443/apis/storage.k8s.io/v1?timeout=32s
I0304 09:37:25.089056    7044 request.go:665] Waited for 4.998235097s due to client-side throttling, not priority and fairness, request: GET:https://192.168.111.249:6443/apis/coordination.k8s.io/v1?timeout=32s
I0304 09:37:35.289609    7044 request.go:665] Waited for 2.398534008s due to client-side throttling, not priority and fairness, request: GET:https://192.168.111.249:6443/apis/autoscaling/v2beta2?timeout=32s
NAME                 TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
service/kubernetes   ClusterIP   10.96.0.1    <none>        443/TCP   11m
```

jklippel:
Expected behaviour: No warnings are issued by kubectl.

furkatgofurov7:
@jklippel hi, thanks for opening an issue! Since you already proposed a fix, I am assigning this issue to you and we can discuss a fix in the PR itself.

/kind bug
/assign @jklippel

PR #958, opened 2022-03-04
allow kubectl caching after cluster creation

jklippel:
The .kube directory must belong to the (non-root) user issuing kubectl.
Otherwise kubectl will not be able to cache requests and will issue warnings
due to client-side throttling taking place when requests normally hitting the
cache are directed at the server instead.

fixes: #957

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane-kubeadm-config-centos.yaml
+++ b/vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane-kubeadm-config-centos.yaml
@@ -15,6 +15,7 @@ preKubeadmCommands:
   - systemctl start monitor.keepalived.service
 postKubeadmCommands:
   - mkdir -p /home/{{ IMAGE_USERNAME }}/.kube
+  - chown {{ IMAGE_USERNAME }}:{{ IMAGE_USERNAME }} /home/{{ IMAGE_USERNAME }}/.kube
   - cp /etc/kubernetes/admin.conf /home/{{ IMAGE_USERNAME }}/.kube/config
   - chown {{ IMAGE_USERNAME }}:{{ IMAGE_USERNAME }} /home/{{ IMAGE_USERNAME }}/.kube/config
 files:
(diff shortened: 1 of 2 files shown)


jklippel:
Also created pull-request https://github.com/metal3-io/cluster-api-provider-metal3/pull/552 to fix the examples in cluster-api-provider-metal3 accordingly.

vm-setup/roles/v1aX_integration_test/templates/cluster-template-controlplane-kubeadm-config-centos.yaml
```diff
@@ -15,6 +15,7 @@ preKubeadmCommands:
   - systemctl start monitor.keepalived.service
 postKubeadmCommands:
   - mkdir -p /home/{{ IMAGE_USERNAME }}/.kube
+  - chown {{ IMAGE_USERNAME }}:{{ IMAGE_USERNAME }} /home/{{ IMAGE_USERNAME }}/.kube
```
furkatgofurov7, line 18: should we be specific and give permission to cache folder only?
```suggestion
  - chown {{ IMAGE_USERNAME }}:{{ IMAGE_USERNAME }} /home/{{ IMAGE_USERNAME }}/.kube/cache
```

jklippel: As it is in the users home the whole directory should belong to the user. I do not see any reason why it should not be.
Generally from a users perspective I would not be amused if I cannot remove a directory in my home directory because it belongs to root.

furkatgofurov7: Fair enough, also I just noticed that we already chown `.kube/config` in the same file (Line 20)

jklippel: Yeah, that is still needed as the copy in line 19 runs as root as well and the config file gets created with root:root. I would leave it like this even though a
```
chown -R chown {{ IMAGE_USERNAME }}:{{ IMAGE_USERNAME }} /home/{{ IMAGE_USERNAME }}/.kube
```
replacing both current chown command would be possible, I think it is better to leave it as explicit as it is right now with the additional advantage of not having to use -R.

furkatgofurov7 (approved):
/approve

furkatgofurov7:
@fmuyassarov @namnx228 @lentzi90 @kashifest folks ptal

lentzi90 (commented):
Thanks for the fix!
/lgtm

PR #891, opened 2021-12-20
CI: prepare for the new ironic container

dtantsur:
We will switch to using a combined API+conductor Ironic process in https://github.com/metal3-io/baremetal-operator/pull/1065.

PR #959, opened 2022-03-07
🌱 Remove ironic container workarounds in upgrade and pivoting tests

furkatgofurov7:
After https://github.com/metal3-io/baremetal-operator/pull/1091/commits/623b65594be21c02e3a15117cdd6588d8352f0b9 upgrade tests started failing since we introduced some workarounds to tests in #891  but now we are using all in one container - ironic, instead of ironic-api+ironic-conductor). This removes those workarounds both in pivoting and upgrade tests

Also updates notes in feature tests README a bit.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/move.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/move.yml
@@ -9,13 +9,10 @@
           - vbmc
       ironic_containers:
           - ironic
-          - ironic-api
-          - ironic-conductor
           - ironic-inspector
           - ironic-endpoint-keepalived
           - ironic-log-watch
           - dnsmasq
-          - mariadb
 
   - name: Fetch container logs (kind cluster)
     block:
@@ -30,9 +27,6 @@
 
       - name: Fetch container logs before pivoting (kind cluster)
         shell: "sudo {{ CONTAINER_RUNTIME }} logs {{ item }} > /tmp/{{ CONTAINER_RUNTIME }}/{{ item }}/stdout.log 2> /tmp/{{ CONTAINER_RUNTIME }}/{{ item }}/stderr.log"
-        # FIXME(dtantsur): remove ignore_errors after the transition period
-        # from ironic-api+conductor to just ironic
-        ignore_errors: yes
         with_items:
           - "{{ ironic_containers }}"
           - "{{ general_containers }}"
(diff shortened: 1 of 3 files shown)


lentzi90 (commented):
Looks fine to me.
One thing I'm not understanding is what we are testing in the upgrade now. Are we going from the old multi-container Ironic to the new single container or are we starting directly with the single container? It would be very nice to test this transition!
I tried looking in the manifests and logs but I couldn't really tell based on that what is tested.

furkatgofurov7:
> we starting directly with the single container

We are doing this one.

lentzi90:
> We are doing this one.

Thanks for confirming! I'm fine with this fix. Hopefully we could improve the upgrade test later at some point so we can also test the transition.

dtantsur:
Note that we're still using mariadb for the time being (just didn't have time to get rid of it).

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -9,13 +9,10 @@
           - vbmc
       ironic_containers:
           - ironic
-          - ironic-api
-          - ironic-conductor
           - ironic-inspector
           - ironic-endpoint-keepalived
           - ironic-log-watch
           - dnsmasq
-          - mariadb
```
furkatgofurov7, line 18: @dtantsur are you suggesting leaving MariaDB in this list? I am not sure why the fetching logs task is complaining about this container if we are still using it.

furkatgofurov7: I double checked it, here is flow:

- when KIND used, we do run [RUN_LOCAL_SCRIPT](https://github.com/metal3-io/metal3-dev-env/blob/c539be459382279fb800d4fcf16e0d34c674d5cf/03_launch_mgmt_cluster.sh#L190) in 03_*
-  in run local ironic script we [set](https://github.com/metal3-io/baremetal-operator/blob/main/tools/run_local_ironic.sh#L93) using mariadb container by default to false and will [run](https://github.com/metal3-io/baremetal-operator/blob/main/tools/run_local_ironic.sh#L259-L266) it only when true. 

So, I do not think we need to keep MariaDB container in that list

furkatgofurov7: CI also confirms that  `IRONIC_USE_MARIADB=false`, i.e this run: https://jenkins.nordix.org/view/Metal3%20Main/job/metal3_main_v1a5_integration_test_ubuntu/40/consoleFull

dtantsur (approved):
Got it!

kashifest:
/hold
Until default branch is main

kashifest:
/hold cancel
Default branch is now main. Remember to rebase your PR on top of main branch.

kashifest (approved):

PR #962, opened 2022-03-11
WIP: Set BMO and Ironic tags based on CAPM3 release

lentzi90:
We should set the tags for Ironic and BMO based on what CAPM3 release we are using.


lib/common.sh
```diff
@@ -207,8 +210,17 @@ fi
...
-export BAREMETAL_OPERATOR_IMAGE=${BAREMETAL_OPERATOR_IMAGE:-"${CONTAINER_REGISTRY}/metal3-io/baremetal-operator"}
+# Baremetal operator and Ironic image
+# Set the tags based on the CAPM3RELEASE. Both BMO and Ironic has tags with the
+# following format: capm3-vX.Y.Z, where vX.Y.Z is the CAPM3 release.
+bmo_tag="latest"
+ironic_tag="latest"
+if [ -n "${CAPM3RELEASE}" ]; then
+  bmo_tag="capm3-${CAPM3RELEASE}"
+  ironic_tag="capm3-${CAPM3RELEASE}"
+fi
```
lentzi90, line 221: As far as I can tell `CAPM3RELEASE` will always be defined at this point. Should we just skip the if statement?
Another question would be if we want to use the latest tag for some tests? Now it will always use a tag like `capm3-v1.1.0` except for BMO PRs where it will build the image locally.

furkatgofurov7: CAPM3RELEASE should be always there since you are sourcing releases.sh meaning it will always override the previous  bmo_tag/ironic_tag vars, so I think we can skip the condition. 



> nother question would be if we want to use the latest tag for some tests?

if that is the case, depending on what kind of test it is, we can introduce condition so that:

```
if X-test; then
    bmo_tag="latest"
    ironic_tag="latest"
else
    bmo_tag="capm3-${CAPM3RELEASE}"
    ironic_tag="capm3-${CAPM3RELEASE}"
fi

```

lentzi90: Thanks for confirming. I think I'll just remove the if statement then.

Btw, I put the tag variables in lowercase here (`bmo_tag` and `ironic_tag`) as I wanted to make it clear they are not to be set by the user. If there is already some other convention for this please let me know.

fmuyassarov: > Now it will always use a tag like `capm3-v1.1.0` except for BMO PRs where it will build the image locally.

+ for Ironic

fmuyassarov (commented):
Looks good. I believe current functionality of building local images for BMO & Ironic is not broken.
/approve

lentzi90:
I have now removed the unnecessary if statement and squashed the commits. Rerunning tests also.
/test-v1a5-centos-integration
/test-v1b1-ubuntu-integration
/test-upgrade-features

furkatgofurov7 (approved):

smoshiur1237 (approved):
/lgtm

01_prepare_host.sh
```diff
@@ -29,8 +33,14 @@ elif [[ $OS == "centos" || $OS == "rhel" ]]; then
...
 fi
 
+# Ansible version
+export ANSIBLE_VERSION=${ANSIBLE_VERSION:-"4.10.0"}
 sudo pip3 install ansible=="${ANSIBLE_VERSION}"
 
+# Now that the basics are installed, we can source common.sh.
+# common.sh cannot be sourced before things like curl and jq are installed.
+# shellcheck disable=SC1091
+source lib/common.sh
```
mboukhalfa, line 43: Is not jq will be installed later in line 54 
```
# Install required packages
  -e "metal3_dir=$SCRIPTDIR" \
  -e "virthost=$HOSTNAME" \
  -i vm-setup/inventory.ini \
  -b vm-setup/install-package-playbook.yml
```

PR #969, opened 2022-03-17
Add CAPI labels on hardwareData CRD to pivot

fmuyassarov:
Add CAPI labels on hardwareData CRD (added in https://github.com/metal3-io/baremetal-operator/pull/1099) so that hadwareData object is also pivoted as part of `clusterctl move` operation.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/move.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/move.yml
@@ -66,10 +66,17 @@
       namespace: "{{ IRONIC_NAMESPACE }}"
     when: EPHEMERAL_CLUSTER == "minikube"
 
-  - name: Label BMO CRDs.
+  - name: Label baremetalhost CRD to pivot.
     shell: "kubectl label --overwrite crds baremetalhosts.metal3.io {{ item }}"
     with_items:
        - clusterctl.cluster.x-k8s.io=""
+    when: CAPM3_VERSION != "v1alpha4"
+
+  - name: Label hardwareData CRD to pivot.
+    shell: "kubectl label --overwrite crds hardwaredata.metal3.io {{ item }}"
+    with_items:
+       - clusterctl.cluster.x-k8s.io=""
+       - clusterctl.cluster.x-k8s.io/move=""
     when: CAPM3_VERSION != "v1alpha4"
 
   - name: Obtain target cluster kubeconfig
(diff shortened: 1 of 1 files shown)


vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -66,12 +66,19 @@
...
     shell: "kubectl label --overwrite crds baremetalhosts.metal3.io {{ item }}"
     with_items:
        - clusterctl.cluster.x-k8s.io=""
     when: CAPM3_VERSION != "v1alpha4"
 
+  - name: Label hostdata CRD to pivot.
+    shell: "kubectl label --overwrite crds hostdata.metal3.io {{ item }}"
+    with_items:
+       - clusterctl.cluster.x-k8s.io=""
+       - clusterctl.cluster.x-k8s.io/move=""
```
lentzi90, line 79: Should this label also be added to the BMH CRD above? I'm not sure why it is here but not there.

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -130,11 +137,26 @@
       chdir: "{{ BMOPATH }}"
 
   - name: Label BMO CRDs in target cluster.
+  - name: Restore original BMO Configmap
+    shell: "mv {{ BMOPATH }}/config/default/ironic.env.orig {{ BMOPATH }}/config/default/ironic.env"
+    when: CAPM3_VERSION != "v1alpha4"
+
+  - name: Restore original Ironic Configmap
+    shell: "mv {{ BMOPATH }}/ironic-deployment/keepalived/ironic_bmo_configmap.env.orig {{ BMOPATH }}/ironic-deployment/keepalived/ironic_bmo_configmap.env"
+
```
kashifest, line 146: Do we still have these files @lentzi90 ?

lentzi90: This looks like a bad rebase. These commands were removed in https://github.com/metal3-io/metal3-dev-env/pull/975

fmuyassarov: Fxied

vm-setup/roles/v1aX_integration_test/tasks/move.yml
fmuyassarov: @lentzi90  I'm not sure why we don't need move for BMH. But only clusterctl label on HD CRD wasn't enough to pivot it during my tests. And from the description, move label is actually what we need to ensure that CRD that's not controlled/installed via clusterctl, can be pivoted. Ref: https://github.com/kubernetes-sigs/cluster-api/blob/a6edb8e4355897830a1ea6006195db3a3b4f3565/cmd/clusterctl/api/v1alpha3/labels.go#L35

fmuyassarov:
/hold
Need to test labeling questions

vm-setup/roles/v1aX_integration_test/tasks/move.yml
lentzi90: I'm guessing that the BMH is linked to a cluster through owner references already so that's probably why we don't need it. Thanks for checking!

fmuyassarov: You are right, BMH is owned by Metal3Machine. And if the object is linked to cluster indirectly (via ownerRef chain) it can still be pivoted. But as said I will confirm soon after my tests

lentzi90 (commented):
Thanks for the thorough check @fmuyassarov !

I think the only concern with automatically setting those labels is that it can confuse people when the labels are set but you still need to manually make sure the BMO/Ironic work with the cluster you are moving to. But I think it would make sense to still set the labels by the controller and document clearly that the user must ensure that the actual controllers also work with the target cluster.

/lgtm

kashifest:
The move document should already have this documented AFAIR

kashifest:
I am thinking is setting cluster name on BMH a left over code from past when BMH was Part of CAPM3 and Shall we stop doing that now. IMO this should Always be the Job of user not the controller

PR #970, opened 2022-03-18
let Tilt handle image tag changing process.

Xenwar:
When using tilt, the bmo image tag is changing twice. 
1. When deploying bmo to` 192.168.111.1:5000/localimages/baremetal-operator`
2. When code changes and tilt rebuilds the image. 

(1) is not necessary and is causing for the tilt environment setup to fail due to missing image, even tough the image is in the local registry. Loading the image to the kind cluster did not help either. 

This PR lets tilt handle the tagging process.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -73,6 +73,7 @@ function launch_baremetal_operator() {
 
   # Deploy BMO using deploy.sh script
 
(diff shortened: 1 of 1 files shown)


fmuyassarov (approved):

furkatgofurov7 (commented):
/lgtm

PR #971, opened 2022-03-18
Ensure python3-libselinux is installed for centos/rhel

andfasano:
This is a requirement for the ansible.posix.selinux module


andfasano:
Looks like the underlying issue needs to be addressed differently

PR #972, opened 2022-03-18
Remove the "Management cluster forced to be minikube" message

dtantsur:
I'm not sure how useful it is, but it appears in the output of
the baremetal client, making scripting against it harder.


furkatgofurov7 (commented):
/lgtm

PR #974, opened 2022-03-23
Make sure jq is installed before common.sh is sourced

lentzi90:
I somehow missed this in the last PR, even though I added a comment on line 41 about it... It wasn't caught by CI since there we already have jq and curl installed. However, in other places this can be a real problem.


lentzi90:
Thanks for reporting this @mboukhalfa !

kashifest (approved):

PR #975, opened 2022-03-24
Use the same config for Ironic after pivot

lentzi90:
Before pivot, Ironic can run either as a local container on the host or
in the bootstrap cluster. After pivot, it always runs in the workload
cluster. There was an issue when pivoting from a local container into
the workload cluster that resulted in wrong configuration for Ironic.

The issue comes from the fact that we configure the local Ironic through
environment variables directly, but the in-cluster Ironic is configured
through deployment manifests that should be edited based on the
environment variables. What happened was that we never edited these
manifests when going for a local Ironic, which meant that the manifests
didn't have the correct configuration when pivoting.

To solve the issue, I have rearanged the 03 script so that we always
edit the manifests (even when running Ironic locally). I also removed
the restoration of these manifests to the original since they are used
in multiple places also after the 03 script. They are anyway cleaned on
every rerun and can be easily restored manually with a `git reset
--hard`.


ansible.cfg
```diff
@@ -1,5 +1,5 @@
 [defaults]
-callback_whitelist=ansible.posix.profile_tasks
```
lentzi90, line 2: There is a deprecation warning in the logs for this, telling us to switch to the new naming `callbacks_enabled`.

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -91,12 +91,6 @@
       state: present
       kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
 
-  # Configure Ironic configmap
-  - name: Configure Ironic Configmap
-    shell: |
-      cp {{ BMOPATH }}/ironic-deployment/keepalived/ironic_bmo_configmap.env {{ BMOPATH }}/ironic-deployment/keepalived/ironic_bmo_configmap.env.orig
-      cp {{ IRONIC_DATA_DIR }}/ironic_bmo_configmap.env  {{ BMOPATH }}/ironic-deployment/keepalived/ironic_bmo_configmap.env
-
```
kashifest, line 99: I thought this is the piece of code I added to exactly tackle the issue you are trying to solve. Check this https://github.com/metal3-io/metal3-dev-env/pull/554

lentzi90: It is not quite the same thing I think. The configmap was working correctly as far as I know, but the other manifests were not. I removed these lines to simplify the process. Like this

Old process:
1. In 03_launch_mgmt_cluster.sh, generate configmaps, modify manifests and deploy BMO/Ironic (here we have the issue that the manifests are not modified for Ironic when using Kind).
2. Clean up (but keep backup of configmap)
3. In move.yml, put the backup of the configmap back. Pivot.
4. Clean up again
5. In move_back.yml, pivot again (but here I think we are missing the correct configmap).

New process:
1. In 03_launch_mgmt_cluster.sh: generate configmaps and modify manifests (also when using kind).
2. Leave everything the way it is. This is simpler and safer than changing and changing back everything.
3. In move.yml, just pivot. No need to change the configmaps or manifests. Everything is already correct.
4. In move_back.yml. Same.

If we really want to clean the repo I think we should add that in the last step in `cleanup.yml`. But my 2 cents is that it is easier to understand what is happening if we let the repo remain dirty. The next time we run the 03 script it will anyway reset or we can manually `git reset --hard`.

namnx228 (commented):
I am a bit unsure about not restoring modified files. It would add manual work for the use case where FORCE_REPO_UPDATE=false. Does restoring cause any issues?

vm-setup/roles/v1aX_integration_test/tasks/move.yml
kashifest: We have always had a confusion if we should keep it dirty or we should keep it intact and we somehow kept it so that we would keep the files intact. If you check the conversation on the PR I mentioned you will get some idea.

mboukhalfa: +1 for simplifying the bmo ironic configuration steps by removing the repeated `creation and restoration of backups` and move cleaning the repo to the cleanup script or using a make phony where we can do the back up only once at the beginning and restore after the end of the test

lentzi90:
> I am a bit unsure about not restoring modified files. It would add manual work for the use case where FORCE_REPO_UPDATE=false. Does restoring cause any issues?

That is a good point. I can add a `git reset --hard` in the `cleanup.yml` to fix this. But I really want to get rid of the back and forth of backup/edit/restore that we do now in 03 script, move.yml and the same in e2e pivot and re pivot. We should also do in move_back.yml actually but I think we have forgotten it. It adds unnecessary code and makes it far too easy to make mistakes (like in move_back.yml).

kashifest:
I was talking about particularly mael's comment https://github.com/metal3-io/metal3-dev-env/pull/554#issuecomment-724740848 which suggests why we keep the files intact

namnx228:
> > I am a bit unsure about not restoring modified files. It would add manual work for the use case where FORCE_REPO_UPDATE=false. Does restoring cause any issues?
> 
> That is a good point. I can add a `git reset --hard` in the `cleanup.yml` to fix this. But I really want to get rid of the back and forth of backup/edit/restore that we do now in 03 script, move.yml and the same in e2e pivot and re pivot. We should also do in move_back.yml actually but I think we have forgotten it. It adds unnecessary code and makes it far too easy to make mistakes (like in move_back.yml).

Sound reasonable. Just one concern is that do we run the cleanup.yml when running the code locally (not in CI) and getting failed?

lentzi90:
> I was talking about particularly mael's comment [#554 (comment)](https://github.com/metal3-io/metal3-dev-env/pull/554#issuecomment-724740848) which suggests why we keep the files intact and this is what I wanted [#554 (comment)](https://github.com/metal3-io/metal3-dev-env/pull/554#issuecomment-724762227) :) essentially same as you.

I think I understand now. Let me ponder on this. I gave an idea for how to avoid making changes in the repos and instead use a kustomize overlay that is generated by metal3-dev-env. Checking how much effort it would be to implement this.

lentzi90:
I have pushed a second commit now that cleans all repos that we make changes in during `cluster_cleanup.sh`. My first idea was to add it in `cleanup.yml` playbook, but I think it makes more sense to keep the cleanup together with the cluster lifecycle (minikube/kind). This is because the changes we make are used when deploying BMO, Ironic and IPAM, so they should be kept until we tear down the bootstrap cluster where they are running.

I have tested this locally. It works so that when you run `make` the repos will be left dirty. When you run `make clean` they are also cleaned.
Please take a look and let me know if this is an acceptable solution!

(The other idea I had about using a separate kustomize overlay will unfortunately be a larger change. I would still like to implement that at some point but it should be a separate PR I think.)

cluster_cleanup.sh
```diff
@@ -22,3 +22,16 @@ fi
 if [ "${EPHEMERAL_CLUSTER}" == "minikube" ]; then
   sudo su -l -c "minikube delete" "${USER}"
 fi
+
+pushd "${CAPM3PATH}"
+  git reset --hard
```
mboukhalfa, line 27: I think for dev they might need to keep some changes in the local repo that will be removed with reset I would recommend resetting only the config files

lentzi90: I see. But this is only for the repos under the go path, and it is only run when you do `make clean`. Is it really necessary to keep changes in there?

mboukhalfa: It is hard to say `necessary` but if the repos under `go/src` are always only copies of the repos that might have some uncommitted changes (tested_repos) that will not be a problem. I have not seen the case of changing multiple repos in the same time in practice but theoretically it is the case where developer might want to keep his uncommitted changes in some of those repos

lentzi90:
I decided to drop the attempt at simplifying the deployment. It was getting too large. Now there is only the necessary changes to make Ironic use the correct tag after pivot.
I'll try to do another PR for the other changes we discussed here when I have time.

lentzi90:
/hold
I'm missing something still. It is not using the correct image for some reason. Will need to investigate a bit more.

lentzi90:
Part of the problem is fixed here: https://github.com/metal3-io/baremetal-operator/pull/1102
After that PR is merged I'm hoping this one will be enough to fix it completely.

lentzi90:
Things seems to be working as expected now! 
Before upgrade (after pivot) we get this Ironic image: `192.168.111.1:5000/localimages/ironic:capm3-v0.5.5`
After upgrade we get this: `registry.nordix.org/quay-io-proxy/metal3-io/ironic:main`

03_launch_mgmt_cluster.sh
```diff
@@ -185,49 +185,46 @@ EOF
     echo "IRONIC_KERNEL_PARAMS=console=ttyS0" | sudo tee -a "$IRONIC_DATA_DIR/ironic_bmo_configmap.env"
   fi
 
+  # Copy the generated configmap for ironic deployment
+  cp "$IRONIC_DATA_DIR/ironic_bmo_configmap.env"  "${BMOPATH}/ironic-deployment/keepalived/ironic_bmo_configmap.env"
+  # Backup Ironic manifests before changing the images in them.
+  # These are cleaned up in 05_test.sh
+  cp "${BMOPATH}/ironic-deployment/default/kustomization.yaml" "${BMOPATH}/ironic-deployment/default/kustomization.yaml.orig"
+  cp "${BMOPATH}/ironic-deployment/keepalived/kustomization.yaml" "${BMOPATH}/ironic-deployment/keepalived/kustomization.yaml.orig"
```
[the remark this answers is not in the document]
lentzi90: Sorry, I did not make this clear. I'm restoring them in [05_test.sh](https://github.com/Nordix/metal3-dev-env/blob/lentzi90/fix-pivot-config/05_test.sh#L36-L47).

The logic is that we will need them for pivot, repivot and upgrade (possibly some other things as well), so to avoid a lot of confusion and extra steps I'm moving the clean up to the last part. It is quite hard to know where these changes are needed, for example they are also affecting e2e tests run from CAPM3 so I think it is better this way.

This means that if for some reason a developer does not run `05_test.sh`, then the repos will be dirty. It is not ideal but I don't see a better alternative.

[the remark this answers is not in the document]
kashifest: I see less point of keeping this file restoration now at this point, if we need the files to be as it is throughout the test, lets just completely remove this copy and restoration as you were doing before, sorry for pinging you back and forth but the current version doesnt seem logical to me.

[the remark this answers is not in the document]
lentzi90: Ok, no problem! I'll remove it completely and also create an issue to investigate a better solution for this in the future. :slightly_smiling_face:

lentzi90:
/hold
Removing the cleanup and checking that things work properly before squashing commits.
/test-v1b1-ubuntu-integration

lentzi90:
It is working as intended :tada: 
Waiting on another review before I squash in case we still want to do something about the dirty files.

03_launch_mgmt_cluster.sh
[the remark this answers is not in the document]
lentzi90: @kashifest done, please take a look! Note: I put it as a separate commit for now so it is easy to change. If it looks good to you I'll squash

kashifest:
> It is working as intended tada Waiting on another review before I squash in case we still want to do something about the dirty files.

lgtm, please squash

03_launch_mgmt_cluster.sh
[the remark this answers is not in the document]
mboukhalfa: @lentzi90 can you edit the description accordingly I think "For this reason I have moved the
cleanup to the 05_test.sh script." is not true anymore /lgtm

PR #978, opened 2022-04-01
Fix unversioned python on CentOS

lentzi90:
CentOS does not have preconfigured alternatives for python anymore it
seems. It only has the "no-python" alternative even when python3 is
installed. To solve this we can simply create the link manually (same
solution as for CentOS Stream 9).

I tried first to use `alternatives --install /usr/bin/python python /usr/bin/python3 1` to add python3 as an alternative. However, it does not allow me to do so, saying that `/usr/bin/python` is not allowed, it must be `/usr/bin/unversioned-python`. But then we don't get `python`, we get `unversioned-python` which is not helpful... So then I reverted to creating the link manually which is the same as for CentOS Stream 9 a few lines below.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -31,18 +31,17 @@ elif [[ $OS == "centos" || $OS == "rhel" ]]; then
     8)
       sudo dnf config-manager --set-enabled powertools
       sudo dnf install -y epel-release
-      sudo alternatives --set python /usr/bin/python3
       ;;
     9)
       sudo dnf config-manager --set-enabled crb
       sudo dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm
-      sudo ln -s /usr/bin/python3 /usr/bin/python || true
       ;;
     *)
       echo -n "CentOS or RHEL version not supported"
       exit 1
       ;;
   esac
+  sudo ln -s /usr/bin/python3 /usr/bin/python || true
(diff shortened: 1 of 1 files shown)


lentzi90:
I have tested this in a CentOS VM by running `provision_metal3_image_centos.sh` against this branch. Without this patch I get the same error as we see in Jenkins:
```
/usr/bin/python3 has not been configured as an alternative for python
```
With the patch there is no error and the scripts runs successfully past this point.

smoshiur1237 (approved):
Thanks @lentzi90 for the fix.
/lgtm

01_prepare_host.sh
```diff
@@ -31,7 +31,7 @@ elif [[ $OS == "centos" || $OS == "rhel" ]]; then
     8)
       sudo dnf config-manager --set-enabled powertools
       sudo dnf install -y epel-release
-      sudo alternatives --set python /usr/bin/python3
+      sudo ln -s /usr/bin/python3 /usr/bin/python || true
```
elfosardo, line 34: maybe move this and L39 after esac on L45 ?
since the commands are the same for both versions

furkatgofurov7 (approved):
/hold
To address the [comment](https://github.com/metal3-io/metal3-dev-env/pull/978/files#r841072362) from @elfosardo 
/approve

01_prepare_host.sh
lentzi90: Good point, fixed

PR #979, opened 2022-04-04
Revert "Set BMO and Ironic tags based on CAPM3 release"

lentzi90:
This reverts commit de72620a2befaa6441d411ad5b747a38fe65e5e2.

It had the uninstended consequence of not testing the latest code of Ironic and BMO in the main jobs.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -52,7 +52,7 @@
        "datasourceTemplate": "github-releases"
      },
      {
-       "fileMatch": ["^01_prepare_hosts.sh$"],
+       "fileMatch": ["^lib/common.sh$"],
        "matchStrings": ["ANSIBLE_VERSION:-\"(?<currentValue>.*?)\"}"],
        "depNameTemplate": "ansible",
        "datasourceTemplate": "pypi"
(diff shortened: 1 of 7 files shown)


lentzi90:
/test-v1a5-centos-integration
If I'm not mistaken this should now break with the same error we are seeing in the upgrade job.

namnx228 (commented):
/lgtm

lentzi90:
After some investigation the conclusion is that we are not checking properly if Ironic is actually running after pivoting. That is why these tests are passing. I'll create an issue for this.

renovate.json
```diff
@@ -52,7 +52,7 @@
        "datasourceTemplate": "github-releases"
      },
      {
-       "fileMatch": ["^01_prepare_hosts.sh$"],
+       "fileMatch": ["^lib/common.sh$"],
        "matchStrings": ["ANSIBLE_VERSION:-\"(?<currentValue>.*?)\"}"],
```
furkatgofurov7, line 56: Looks like you forgot to move `ANSIBLE_VERSION` back to lib/common.sh, or leave this as is

furkatgofurov7 (changes requested):

renovate.json
lentzi90: Thanks, good find! There was a conflict when reverting and I must have missed this when resolving it.

furkatgofurov7 (approved):
/approve

PR #983, opened 2022-04-05
Check ironic deployment instead of pods after pivot

### LINKED ISSUE #981: Ironic health is not checked properly after pivot
After pivoting Ironic to the workload cluster, we check that "all pods are running" in this way:
https://github.com/metal3-io/metal3-dev-env/blob/481eae31ed28518668ebf21f7d6e7ebd8ca0e63d/vm-setup/roles/v1aX_integration_test/tasks/move.yml#L155-L166

Unfortunately this does not properly identify issues since it will pass as soon as all pods are (for a split second) in running state. If some pod start crash-looping it will still every now and then get to a running state before it crashes, which means that the test passes even when it should not.

mboukhalfa:
fix #981

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/move.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/move.yml
@@ -165,6 +165,16 @@
     until: (target_running_pods is succeeded) and
            (target_running_pods.resources | length == 0)
 
+  - name: Check if all deployments replicas are available
+    kubernetes.core.k8s_info:
+      kind: Deployment
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: deployments
+    retries: 150
+    delay: 20
+    until: (deployments is succeeded) and
+           (deployments.resources | filter_unavailable_replicas | length == 0 )
+
   - name: Pivot objects to target cluster
     shell: "clusterctl move --to-kubeconfig /tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml -n {{ NAMESPACE }} -v 10"
 
--- a/vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
+++ b/vm-setup/roles/v1aX_integration_test/filter_plugins/status.py
@@ -27,7 +27,7 @@ def filter_phase(resources, phase):
             if r["status"]["phase"].lower() == phase:
                 filtered.append(r)
         except KeyError:
-            display.warning(msg("['status']['phase']", resources))
+            display.warning(msg("['status']['phase']", r))
 
     return filtered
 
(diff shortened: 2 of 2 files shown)


mboukhalfa:
/assign @kashifest 
cc @lentzi90

lentzi90 (commented):
This looks good to me!
Can you add also a check for BMO? I think we should check both Ironic and BMO deployments. The old check was for all pods in the cluster but I think the main point here is to see that Ironic and BMO are healthy.

kashifest:
I would rather keep the pods check, it was done for multiple reason, one particular reason being, the provider components needs to be up and running before the pivot operation, you can add the deployment check in addition but do it in a loop for ironic and bmo rather than adding two different tasks.

lentzi90:
> I would rather keep the pods check, it was done for multiple reason, one particular reason being, the provider components needs to be up and running before the pivot operation, you can add the deployment check in addition but do it in a loop for ironic and bmo rather than adding two different tasks.

I guess the question then is if we should check the provider components also with the deployment check? As we have seen with Ironic, the "pods are running" check does not capture crash-looping pods in a good way.

kashifest:
> > I would rather keep the pods check, it was done for multiple reason, one particular reason being, the provider components needs to be up and running before the pivot operation, you can add the deployment check in addition but do it in a loop for ironic and bmo rather than adding two different tasks.
> 
> I guess the question then is if we should check the provider components also with the deployment check? As we have seen with Ironic, the "pods are running" check does not capture crash-looping pods in a good way.

Yeah thats what I meant

vm-setup/roles/v1aX_integration_test/tasks/move.yml
```diff
@@ -165,6 +165,16 @@
     until: (target_running_pods is succeeded) and
            (target_running_pods.resources | length == 0)
 
+  - name: Check if all deployments replicas are available
+    kubernetes.core.k8s_info:
+      kind: Deployment
+      kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+    register: ironic_bmo_deployments
```
lentzi90, line 172: This is all deployment, right?
```suggestion
    register: deployments
```

lentzi90 (commented):
The check looks good to me but it is quite spammy in the logs. Would it be possible to reduce the output somehow?

lentzi90 (changes requested):
The functionality is good, but it is still way too spammy I think.
Check the logs at "Check if pods in running state" and "Check if all deployments replicas are available". It is huge!

mboukhalfa:
> Check if all deployments replicas are available

it is printing some warnings when could not find the unavailable replicas he warning print the full deployment yaml. I can replace the full yaml with the deployment name or omit completely the warning but the final out put it also print all deployments yaml and that is how our ansible verbosity is configured do you want to reduce that also ?

lentzi90:
> > Check if all deployments replicas are available
> 
> it is printing some warnings when could not find the unavailable replicas the warning print the full deployment yaml. I can replace the full yaml with the deployment name or omit completely the warning but the final out put it also print all deployments yaml and that is how our ansible verbosity is configured do you want to reduce that also ?

I would say getting rid of the warning is the most important part. The normal print is ok.

fmuyassarov (approved):

Issue #915, opened 2022-01-26
Container registry not respected for Ironic after pivot


lentzi90:
Ironic can be deployed "locally" or in the cluster. Currently we deploy in the cluster if using minikube as management cluster and deploy locally otherwise (KinD).

https://github.com/metal3-io/metal3-dev-env/blob/50fa7121a7680ea6965ed3d77219fb6715681ab3/03_launch_mgmt_cluster.sh#L222-L243

The images used in each scenario are the same, but they are taken from different places. For the local deployment only environment variables matter, but for deploying to the cluster we need to make changes to the manifests and then apply them.

Now a problem appears when we start with a local deployment and pivot to run Ironic in the target cluster. The images in the manifests are not updated since we started with a local deployment. This means that after pivot we use the default images in the manifests instead of the local registry or proxy.

This can be seen by checking the artifacts of the [airship_master_feature_tests_upgrade_ubuntu job in Jenkins](https://jenkins.nordix.org/view/Airship/job/airship_master_feature_tests_upgrade_ubuntu/) where we log the container images after pivot in `manifests/container_images_before_upgrade.txt`.

**Suggested fix:** Update the manifests in `03_launch_mgmt_cluster.sh` no matter if Ironic is running locally or not. Alternatively, add the necessary changes to the playbook `move.yml` so that the manifests are updated there before deploying Ironic in the cluster.

PR #985, opened 2022-04-08
Support setting tag for BMO and Ironic images

lentzi90:
Currently BMO and Ironic will always use the latest images and there is no way to configure it. This PR introduces variables for the tags that we can use for example in upgrade tests when we do not want to use the latest.


lentzi90:
Things seems to be working as intended. I can see the correct tags being used in the upgrade test. However, due to #915 Ironic will use the wrong image after pivot. This is fixed in #975.

fmuyassarov (commented):
/lgtm

PR #988, opened 2022-04-14
Replace curl with wget

kashifest:


kashifest:
Not replacing all the curls in 01 script since wget doesn't have the same alternative to those curl command line options. But this PR should be enough to get rid of the curl issue with centos stream 9 image builds.

lentzi90 (commented):
/lgtm

furkatgofurov7 (approved):
/approve

PR #991, opened 2022-04-25
Delete host as part of cluster clean up

fmuyassarov:
Delete host as part of clean up script. We already wait for BMH deprovisioning, as such adding a task to delete the host should not increase overall test run time much (usually ~2-3 mins). This will ensure that we don't miss a bug in BMO code that would break deletion workflow like it was missed in https://github.com/metal3-io/baremetal-operator/issues/1113.

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
@@ -91,3 +91,21 @@
     delay: 3
     until: (deprovisioned_cluster is succeeded) and
            (deprovisioned_cluster.resources | length ==  0)
+
+  - name: Delete baremetalhosts
+    kubernetes.core.k8s:
+      state: absent
+      src: "{{ WORKING_DIR }}/bmhosts_crs.yaml"
+      namespace: "{{ NAMESPACE }}"
+    ignore_errors: yes
+
+  - name: Wait until no baremetalhost is remaining
+    kubernetes.core.k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+    register: dedeleted_baremetalhost
+    retries: 100
+    delay: 3
+    until: (dedeleted_baremetalhost is succeeded) and
+           (dedeleted_baremetalhost.resources | length ==  0)
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -246,11 +246,12 @@ function make_bm_hosts() {
 # Apply the BMH CRs
 #
 function apply_bm_hosts() {
+  NAMESPACE=$1
   pushd "${BMOPATH}"
   list_nodes | make_bm_hosts > "${WORKING_DIR}/bmhosts_crs.yaml"
   if [[ -n "$(list_nodes)" ]]; then
     echo "bmhosts_crs.yaml is applying"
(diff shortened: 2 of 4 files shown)


vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
```diff
@@ -91,3 +91,21 @@
...
+      src: "{{ WORKING_DIR }}/bmhosts_crs.yaml"
+      namespace: "{{ NAMESPACE }}"
+    ignore_errors: yes
+
+  - name: Wait until no baremetalhost is remaining
+    kubernetes.core.k8s_info:
+      api_version: metal3.io/v1alpha1
+      kind: BareMetalHost
+      namespace: "{{ NAMESPACE }}"
+    register: dedeleted_baremetalhost
```
kashifest, line 107: typo

vm-setup/roles/v1aX_integration_test/tasks/verify.yml
```diff
@@ -48,6 +48,22 @@
...
+    with_items:
+      - etcd
+      - kube-apiserver
+      - kube-controller-manager
+      - kube-scheduler
+
+  - name: Wait for control-plane pods become ready
+    shell: "kubectl wait --namespace=kube-system --for=condition=Ready pods --selector tier=control-plane --timeout=600s"
+    register: control_plane_pods_ready
+
```
kashifest, line 66: Do we need to have shell commands here ? I was under the impression that these could be replace by k8s ansible modules ?

kashifest (changes requested):

vm-setup/roles/v1aX_integration_test/tasks/verify.yml
fmuyassarov: done

fmuyassarov:
@kashifest @dtantsur 
CI checks will fail if we run it now since BMO bug is not fixed yet. Should we merge this and try to fix the BMO bug, or vice versa ?

kashifest:
> @kashifest @dtantsur CI checks will fail if we run it now since BMO bug is not fixed yet. Should we merge this and try to fix the BMO bug, or vice versa ?

I dont think we should block the CI, lets wait for BMO fix.

vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
smoshiur1237:
```suggestion
    register: deleted_baremetalhost
```
typo

vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
```diff
@@ -91,3 +91,21 @@
...
+    retries: 100
+    delay: 3
+    until: (dedeleted_baremetalhost is succeeded) and
```
smoshiur1237, line 110: typo

vm-setup/roles/v1aX_integration_test/tasks/cleanup.yml
```diff
@@ -91,3 +91,21 @@
...
+           (dedeleted_baremetalhost.resources | length ==  0)
```
smoshiur1237, line 111: typo

fmuyassarov:
> > @kashifest @dtantsur CI checks will fail if we run it now since BMO bug is not fixed yet. Should we merge this and try to fix the BMO bug, or vice versa ?
> 
> I dont think we should block the CI, lets wait for BMO fix.

Ignore it, apparently I was using an old image of BMO. If CI passed we should be good to merge this patch.

smoshiur1237 (approved):
/lgtm

PR #992, opened 2022-04-25
Wait for controlplane pods to be Ready before proceeding with Calico

fmuyassarov:
/cc @kashifest @lentzi90

### DIFF
--- a/vm-setup/roles/v1aX_integration_test/tasks/verify.yml
+++ b/vm-setup/roles/v1aX_integration_test/tasks/verify.yml
@@ -48,6 +48,33 @@
                       - name: IP_AUTODETECTION_METHOD
                         value: "cidr={{ EXTERNAL_SUBNET_V4_HOST }}/{{ EXTERNAL_SUBNET_V4_PREFIX }}"
 
+  - name: Wait for all control-plane pods become created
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
I think it is kube-proxy that is causing the issue actually, not the control-plane components. But I guess the real question is why this started happening suddenly? :thinking: Does anyone know if kube-proxy used to be baked in the image?

fmuyassarov:
> I think it is kube-proxy that is causing the issue actually, not the control-plane components. But I guess the real question is why this started happening suddenly? thinking Does anyone know if kube-proxy used to be baked in the image?

my bad, I forgot to add the kube-proxy :)

fmuyassarov:
> I think it is kube-proxy that is causing the issue actually, not the control-plane components. But I guess the real question is why this started happening suddenly? thinking Does anyone know if kube-proxy used to be baked in the image?

I don't think we back kube-proxy image, it is stored in Google container registry, and we don't usually had pulling issues like we have with Docker registry.

lentzi90:
I think I found the issue. The changes I made to centos cloud-init affected both metal3 and node images (I thought it was only metal3). This PR should hopefully fix it: https://github.com/Nordix/metal3-dev-tools/pull/522

mboukhalfa:
@lentzi90 can we close this PR ?

lentzi90:
Yes it is fine by me to close it.

PR #1000, opened 2022-05-10
Introduce Ubuntu 22.04 uplift related changes

kashifest:


kashifest:
/hold
Until https://github.com/Nordix/metal3-dev-tools/pull/532 merges and appropriate images are created.

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -28,11 +28,13 @@
         owner: root
         group: root
         mode: '0644'
+      when: ansible_distribution_version != "22.04"
```
lentzi90, line 31: Is this opensuse repo not needed for 22.04?

kashifest: It was missing when I created this PR, the links for Ubuntu 22.04 got added on 18th of May, I will do a followup PR on this, good catch.

kashifest: Removing it doesnt hamper the tests, so I am not sure if this is needed anymore or not.

lentzi90: Ah now I think I get it! It looks like the same we already do when we build the metal3 image here: https://github.com/Nordix/metal3-dev-tools/blob/master/ci/scripts/image_scripts/install_crio_on_ubuntu.sh#L33-L42
So in the CI it probably doesn't matter, but when starting from a fresh cloud image it could be needed

furkatgofurov7 (approved):

vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
kashifest: Yes perhaps

kashifest: Is it ok @lentzi90 if I do it in a follow up PR to avoid running the tests here or do you want me to push it with this commit?

lentzi90: Yes that's fine with me!

PR #1001, opened 2022-05-10
Move BMH cleanup tasks to main.yaml

fmuyassarov:
We don't want the BMH CRs to be deleted for every types of the tests like integration, features, upgrade. Deletion check is only needed in integration tests which we run on every single PR in most of the repos. This patch is just moving around the deletion Ansible task so that deletion takes places only for the integration tests.
/cc @kashifest


smoshiur1237 (commented):
@fmuyassarov  Thanks for the fix. 
/lgtm

PR #1006, opened 2022-05-17
Change Centos target nodes network configuration to nmcli commands

Sunnatillo:
These changes make network configuration of target nodes plugin(keyfile, ifcfg-rh) independent


Sunnatillo:
New PR opened #1009

PR #1008, opened 2022-05-18
🐛 Pass ns to apply_bm_hosts func in tilt to avoid creating BMHs in default ns

furkatgofurov7:
Without this, BMHs will be created in the default namespace when tilt is used as an ephemeral cluster. In normal m3-dev-env deployment (using kind/minikube) namespace is passed to apply_bm_hosts func properly: https://github.com/metal3-io/metal3-dev-env/blob/main/03_launch_mgmt_cluster.sh#L519 so this went in unnoticed


lentzi90 (approved):
/lgtm

PR #1013, opened 2022-05-26
Do not check for passwordless sudo if running as root

javipolo:
When running as root it should not be needed passwordless sudo, so skip the check if running as root

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -2,18 +2,20 @@
 
 [[ ":$PATH:" != *":/usr/local/go/bin:"* ]] && PATH="$PATH:/usr/local/go/bin"
 
+USER="$(whoami)"
+export USER=${USER}
 # Verify that passwordless sudo is configured correctly
-if ! sudo -nl | grep -q '(ALL) NOPASSWD: ALL';then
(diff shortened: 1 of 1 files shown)


lib/common.sh
```diff
@@ -2,18 +2,19 @@
...
+export USER="$(whoami)"
 # Verify that passwordless sudo is configured correctly
-if ! sudo -nl | grep -q '(ALL) NOPASSWD: ALL';then
-    echo "ERROR: metal3-dev-env requires passwordless sudo configuration!"
-    exit 1
+if [[ "$USER" != "root" ]]; then
+    if ! sudo -nl | grep -q '(ALL) NOPASSWD: ALL';then
+        echo "ERROR: metal3-dev-env requires passwordless sudo configuration!"
+        exit 1
+    fi
```
mboukhalfa, line 11: Please consider line https://github.com/metal3-io/metal3-dev-env/blob/048058d20b8546816ac9d3f965c77077fcbc4d54/lib/common.sh#L289

mboukhalfa: Not sure I just thought might include this restriction when checking if the user is root for the clarity of the script

furkatgofurov7 (commented):
/lgtm

PR #1014, opened 2022-05-27
Add ubuntu-22-04 docker firewall bug fix

kashifest:


furkatgofurov7:
@kashifest do we still need this? if not let's close it.

PR #1019, opened 2022-06-07
WIP Test NICs to avoid crashing the kernel during tests

Rozzii:


Rozzii:
The issue was solved .
/close

PR #1022, opened 2022-06-10
Add cdrom for booting from virtual media

Xenwar:
I would like to cdrom to support a use case related to booting from an attached virtual media.


smoshiur1237 (commented):
/lgtm

PR #1031, opened 2022-06-29
Update renovate configuration

kashifest:
This PR updates renovate configuration to monitor patch versions only since we need to uplift to minor versions manually after tests. Also updating the reviewers list here.


lentzi90 (commented):
/lgtm
I also checked that the config is ok with renovate-config-validator. :heavy_check_mark:

furkatgofurov7 (approved):

PR #1033, opened 2022-07-01
Make wget less verbose

lentzi90:
This turns off the progress log for wget but will still show errors and a message when the download is completed. For example like this:

```console
❯ wget --no-verbose -O kubectl https://storage.googleapis.com/kubernetes-release/release/v1.24.2/bin/linux/amd64/kubectl
2022-07-01 08:44:25 URL:https://storage.googleapis.com/kubernetes-release/release/v1.24.2/bin/linux/amd64/kubectl [45711360/45711360] -> "kubectl" [1]
```

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -104,20 +104,20 @@ fi
 
 if [ "${EPHEMERAL_CLUSTER}" == "minikube" ]; then
   if ! command -v minikube 2>/dev/null || [[ "$(minikube version --short)" != "${MINIKUBE_VERSION}" ]]; then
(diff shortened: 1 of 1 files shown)


01_prepare_host.sh
```diff
@@ -127,18 +127,18 @@ else
 fi
 
 KUBECTL_LATEST=$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)
-KUBECTL_LOCAL=$(kubectl version --client --short | cut -d ":" -f2 | sed 's/[[:space:]]//g' 2> /dev/null)
```
lentzi90, line 130: This was not working. At some point the command started to give both the kubectl and the kustomize version.

PR #1054, opened 2022-08-17
Inject ssh key also in ironic ramdisk

lentzi90:
We already  inject the public ssh key in most places, but IPA was missing. This makes it possible to ssh into IPA for debugging.


Sunnatillo (commented):
/lgtm

kashifest:
Does our integration test say anything if its working or not ? No I guess ?

kashifest:
/approve
/hold 
unhold it if you are satisfied it will work

lentzi90:
@mboukhalfa can you double check if this will work? I think you used this functionality recently.

lib/common.sh
```diff
@@ -84,6 +84,8 @@ if [ ! -f "${SSH_KEY}" ]; then
   mkdir -p "$(dirname "$SSH_KEY")"
   ssh-keygen -f "${SSH_KEY}" -P ""
 fi
+SSH_PUB_KEY_CONTENT="$(cat "${SSH_PUB_KEY}")"
+export SSH_PUB_KEY_CONTENT
```
mboukhalfa, line 89: Why do not use the `SSH_PUB_KEY` var that already exported then creating a new var?

lentzi90: `SSH_PUB_KEY` is the path to the file, but `SSH_PUB_KEY_CONTENT` is what is inside the file. (Ironic wants the content, not the path, so it can inject it directly)

mboukhalfa: what about `IRONIC_RAMDISK_SSH_KEY="$(cat "${SSH_PUB_KEY}")"` in the `ironic_bmo_configmap.env`

lentzi90: Sure, that is fine by me. I guess I was thinking that we may need `SSH_PUB_KEY_CONTENT` somewhere else also at some point.
Do you want me to change it?

mboukhalfa: well if we have some reasons for an extra var that is fine! I just thought less vars simplify the code

mboukhalfa (approved):

lib/common.sh
lentzi90: Ok, I'll let it be like it is then. The variable is anyway not exposed to the user directly (set directly internally)

PR #1055, opened 2022-08-18
For live-iso setting one node to always use redfish-virtualmedia

adilGhaffarDev:
The live-ISO e2e test is failing because nodes were only using ipmi or redfish, which do not support live-iso. This PR make sure that at least one node uses redfish-virtualmedia. It also adds new flavor for nodes "redfishvmnode", this flavor of nodes will use redfish-virtualmedia as vm_driver which will be used in live-ISO testing.


02_configure_host.sh
```diff
@@ -18,6 +18,7 @@ fi
 ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -e "working_dir=$WORKING_DIR" \
     -e "num_nodes=$NUM_NODES" \
+    -e "num_redfishvmnodes=$NUM_REDFISHVMNODES" \
```
mboukhalfa, line 21: adding two extra nodes does not seem efficient cannot we have one node and another refishvmnode for integration tests for example ?

adilGhaffarDev: changed it kindly check.

lib/common.sh
```diff
@@ -271,9 +272,9 @@ RESULT_STR=""
 export ANSIBLE_DISPLAY_SKIPPED_HOSTS=no
 
 # Sanity check for number of nodes
-if [ "${NUM_NODES}" -lt "$((NUM_OF_CONTROLPLANE_REPLICAS + NUM_OF_WORKER_REPLICAS))" ]; then
+if [ "$((NUM_NODES + NUM_REDFISHVMNODES))" -lt "$((NUM_OF_CONTROLPLANE_REPLICAS + NUM_OF_WORKER_REPLICAS))" ]; then
```
lentzi90, line 275: This feels like it can become very confusing later. I get why it is this way, but I don't like it. One way to make it easier to understand I think, would be to rename `NUM_NODES` to something like `NUM_NORMAL_NODES` or `NUM_MIXED_NODES` or something like this.

smoshiur1237 (approved):
/lgtm

lentzi90 (commented):
/lgtm
This is a good way to fix the issue, but I think it will become confusing (see comment below). But let's get this merged first and CI back to green, then we can think about how to make this easier to understand.

lib/common.sh
adilGhaffarDev: I agree, I will make separate PRs for this fix since this also need changes in capm3 and capm3 PR for this issue is already merged. Lets merge this one and open separate PRs for name changes.

tests/roles/run_tests/vars/main.yml
```diff
@@ -14,7 +14,8 @@ CLUSTER_APIENDPOINT_IP: "{{ lookup('env', 'CLUSTER_APIENDPOINT_IP') }}"
 CLUSTER_NAME: "{{ lookup('env', 'CLUSTER_NAME') | default('test1', true) }}"
 NUM_OF_CONTROLPLANE_REPLICAS: "{{ lookup('env', 'NUM_OF_CONTROLPLANE_REPLICAS') | default(1, true) }}"
 NUM_OF_WORKER_REPLICAS: "{{ lookup('env', 'NUM_OF_WORKER_REPLICAS') | default(1, true) }}"
-NUM_NODES: "{{ lookup('env', 'NUM_NODES') | default(2, true) }}"
+NUM_NODES: "{{ lookup('env', 'NUM_NODES') | default(1, true) }}"
+NUM_REDFISHVMNODES: "{{ lookup('env', 'NUM_REDFISHVMNODES') | default(1, true) }}"
```
kashifest, line 18: the defaults here aren't matching with lib/common?

vm-setup/roles/common/defaults/main.yml
```diff
@@ -11,7 +11,8 @@ libvirt_nic_model: virtio
 default_disk: 50
 default_memory: 8192
 default_vcpu: 2
-num_nodes: 2
+num_nodes: 1
+num_redfishvmnodes: 1
```
kashifest, line 15: confused here, didnt we agree that we will leave the default setup as it is and only create redfishvmnode for live-iso test case? Wouldn't this change mean you are always using 1 normal node and 1 refishvm node regardless of the test ?

adilGhaffarDev: I changed in common.sh and missed this one, now its fixed.

tests/roles/run_tests/vars/main.yml
adilGhaffarDev: fixed

kashifest:
@adilGhaffarDev do you want to keep the commits or squash them?

adilGhaffarDev:
> @adilGhaffarDev do you want to keep the commits or squash them?

let me squash them

kashifest:
/test-ubuntu-integration-main
/test-centos-integration-release-1-1
/approve
/lgtm
lgtm was removed because of commit squashing.

PR #1059, opened 2022-08-24
Removed redfish virtual media flavor

adilGhaffarDev:
Removed redfish-virtualmedia flavor. We still set nodes to use redfish-virtualmedia but we no longer use separate flavor for it. Check following for new implementation:
vm-setup/roles/libvirt/templates/ironic_nodes.json.j2


dtantsur (changes requested):
Testing with redfish-virtualmedia is a useful feature. This PR lacks some justification.

adilGhaffarDev:
> Testing with redfish-virtualmedia is a useful feature. This PR lacks some justification.

We are still using redfish-virtualmedia but flavor is removed, implementation is bit different now. You can check vm-setup/roles/libvirt/templates/ironic_nodes.json.j2 to see how we are setting redfish-vitualmedia node.

furkatgofurov7 (commented):
@dtantsur we will be still testing the live iso and keep the tests the same way, however the current way of testing led to unneccesary and cumbersome changes (related to NUM_NODES var changes especially) we had to take care of in project infra and in this repo as well. But with the approach in this PR, we are introducing a bit of modification to how BMC_DRIVER will be set when mixed option (the default) is chosen. It makes sure the default behaviour is kept when NUM_NODES=2 (when mixed is set, one of the nodes will get ipmi and other redfish), but when NUM_NODES=4, one of them would be set to redfish-virtualmedia (check modular logic in the jinja template) and remaining either ipmi/redfish. Thus, there would not be any need for previously introduced flavor.

vm-setup/roles/libvirt/templates/ironic_nodes.json.j2
```diff
@@ -20,16 +20,15 @@
...
     {% if vm_driver == 'mixed' -%}
-      {# For live-iso setting one node to always use redfish-virtualmedia #}
-      {% if 'redfishvmnode' in node.name -%}
-        {% set vm_driver_tmp = 'redfish-virtualmedia' -%}
       {# For extraworker always use redfish #}
-      {% elif 'extraworker' in node.name -%}
+      {% if 'extraworker' in node.name -%}
         {% set vm_driver_tmp = 'redfish' -%}
-      {% elif loop.index is divisibleby 2 -%}
+      {% elif loop.index % 3 == 0 -%}
```
furkatgofurov7, line 26: I checked the bmh manifests, and it is always ipmi/redfish-virtualmedia combination when NUM_NODES=2, leaving out redfish. That is because jinja template when `loop.index` is used starts the index with 1 and not 0. If you want it to start the index with 0, then you have to change it to `loop.index0`, xref: https://stackoverflow.com/a/21459009

furkatgofurov7 (changes requested):

vm-setup/roles/libvirt/templates/ironic_nodes.json.j2
adilGhaffarDev: fixed

kashifest (approved):

lentzi90 (commented):
/lgtm

lentzi90:
/hold
Holding while waiting for manual tests to finish. Remove when ready @adilGhaffarDev

PR #1060, opened 2022-08-26
Clean defaut connection before adding ironicendpoint

lentzi90:
The "System eth0" connection is automatically created and "blocks" the ironicendpoint. This deletes the default connection before we create the ironicendpoint.


lentzi90:
/close
Will implement a declarative solution instead.

PR #1061, opened 2022-08-26
Generate templates in tests folder

mboukhalfa:
and rename test file since it does not need the prefix order number


Makefile
```diff
@@ -21,7 +21,7 @@ host_cleanup:
 	./host_cleanup.sh
 
 test:
-	./tests/05_test.sh
+	./tests/test.sh
```
furkatgofurov7, line 24: but `test.sh` does not exist under the `tests` folder: https://github.com/metal3-io/metal3-dev-env/tree/main/tests

tests/roles/run_tests/vars/main.yml
```diff
@@ -5,7 +5,7 @@ NAMESPACE: "{{ lookup('env', 'NAMESPACE') | default('metal3', true) }}"
 IRONIC_NAMESPACE: "{{ lookup('env', 'IRONIC_NAMESPACE') | default('baremetal-operator-system', true) }}"
 NAMEPREFIX: "{{ lookup('env', 'NAMEPREFIX') | default('baremetal-operator', true) }}"
 CRS_PATH: "{{ metal3_dir }}/tests/roles/run_tests/templates"
-TEMP_GEN_DIR: "{{ metal3_dir }}/vm-setup/roles/run_tests/files/manifests"
+TEMP_GEN_DIR: "{{ metal3_dir }}/tests/roles/run_tests/files/manifests"
 MANIFESTS_DIR: "{{ metal3_dir }}/vm-setup/roles/run_tests/files"
```
furkatgofurov7, line 9: what are these folders for and why one `MANIFESTS_DIR` is different from the other `TEMP_GEN_DIR` with this change?

furkatgofurov7 (changes requested):

tests/roles/run_tests/vars/main.yml
mboukhalfa: MANIFESTS_DIR was updated already and it is the path to all manifests I think and TEMP_GEN_DIR is one contains the generated templates

Makefile
mboukhalfa: yes I changing `05_test.sh` that exists to `test.sh` I mean just renaming a file

furkatgofurov7: I am not sure if I understand this

tests/roles/run_tests/vars/main.yml
furkatgofurov7: but the paths are same, we do not have `run_tests` folder under `vm-setup/roles` anymore AFAIU, why the `MANIFESTS_DIR` should not be changed?

Makefile
furkatgofurov7: ah I overlooked file renaming change completely, sorry

tests/roles/run_tests/vars/main.yml
mboukhalfa: nice catch, I will check this. I was thinking that  already under `tests`

mboukhalfa: That var `MANIFESTS_DIR` is not used anywhere so I think we do not needed ! removed is that ok ?

furkatgofurov7 (approved):
/cc @lentzi90 @Sunnatillo @smoshiur1237

lentzi90 (commented):
/lgtm

Sunnatillo (commented):
/lgtm

PR #1062, opened 2022-08-28
Add autoconnect priority to centos target node network config

Sunnatillo:
This PR solves issue of two connection profiles competing for eth0 interface. We set higher priority to eth0.nmconnection profile.
The dev-tools part  of the solution is [here](https://github.com/Nordix/metal3-dev-tools/pull/587)

### DIFF
--- a/tests/roles/run_tests/templates/cluster-template-controlplane-kubeadm-config-centos.yaml
+++ b/tests/roles/run_tests/templates/cluster-template-controlplane-kubeadm-config-centos.yaml
@@ -104,6 +104,8 @@ files:
       interface-name=eth0
       master={{ IRONIC_ENDPOINT_BRIDGE }}
       slave-type=bridge
+      autoconnect=yes
+      autoconnect-priority=999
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
/lgtm

tests/roles/run_tests/templates/cluster-template-controlplane-kubeadm-config-centos.yaml
```diff
@@ -104,6 +104,8 @@ files:
       interface-name=eth0
       master={{ IRONIC_ENDPOINT_BRIDGE }}
       slave-type=bridge
+      autoconnect=yes
+      autoconnect-priority=999
```
kashifest, line 108: 999 means highest priority? How did you come up with that number ?

Sunnatillo: not the highest number. It is int32. I used an example from [here](https://www.redhat.com/sysadmin/becoming-friends-networkmanager). Basically any number above 0 works in our case.

kashifest (approved):

PR #1064, opened 2022-08-30
Fix cert-manager version to v1.9.1 and also fix clusterctl.yaml templating

kashifest:
This PR fixes cert-manager version to v1.9.1. This will be removed once clusterctl used in dev-env also uses the same version of cert manager. This PR also fixes clusterctl.yaml templating to make sure the content is always appended instead of overriden.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -341,6 +341,13 @@ function patch_clusterctl(){
   mkdir -p "${HOME}"/.cluster-api
   touch "${HOME}"/.cluster-api/clusterctl.yaml
 
+  ## Remove these lines
+  ## This is hard-coded until we fix the issue with v1.9.1
+  cat << EOF | sudo tee "${HOME}/.cluster-api/clusterctl.yaml"
+cert-manager:
+  version: "v1.9.1" 
+EOF
+
   # At this point the images variables have been updated with update_images
   # Reflect the change in components files
   if [ -n "${CAPM3_LOCAL_IMAGE}" ]; then
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -342,7 +342,7 @@ function patch_clusterctl(){
   touch "${HOME}"/.cluster-api/clusterctl.yaml
 
   ## Remove these lines
-  ## This is hard-coded until we fix the issue with v1.9.1
+  ## This is hard-coded until we use clusterctl with cert-manager v1.9.1
   cat << EOF | sudo tee "${HOME}/.cluster-api/clusterctl.yaml"
 cert-manager:
   version: "v1.9.1" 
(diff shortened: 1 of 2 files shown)


smoshiur1237 (approved):
/lgtm

03_launch_mgmt_cluster.sh
```diff
@@ -341,6 +341,13 @@ function patch_clusterctl(){
   mkdir -p "${HOME}"/.cluster-api
   touch "${HOME}"/.cluster-api/clusterctl.yaml
 
+  ## Remove these lines
+  ## This is hard-coded until we fix the issue with v1.9.1
```
lentzi90, line 345: This comment looks strange. If there is some issue with v1.9.1 why are we using it?

lentzi90 (commented):
Can you add something in the description about why we fix the version to v1.9.1? Is the goal to always have the same version instead of getting whatever CAPI is using?

03_launch_mgmt_cluster.sh
kashifest: Thats a wrong comment. Good catch, fixing it.

furkatgofurov7: +1, it could be rephrased a bit

kashifest: fixed

tests/roles/run_tests/tasks/generate_templates.yml
```diff
@@ -18,9 +18,11 @@
       mode: '0755'
 
   - name: Deploy clusterctl variables to clusterctl config
-    template:
-      src: clusterctl-vars.yaml
-      dest: "{{ HOME }}/.cluster-api/clusterctl.yaml"
+    ansible.builtin.blockinfile:
+      block: "{{lookup('ansible.builtin.template', '{{ CRS_PATH }}/clusterctl-vars.yaml')}}"
```
furkatgofurov7, line 22: can you explain what we are trying to do here exactly?

lentzi90 (commented):
/lgtm

tests/roles/run_tests/tasks/generate_templates.yml
kashifest: [Append content in clusterctl.yaml instead of overwriting.](https://github.com/metal3-io/metal3-dev-env/pull/1064/commits/1b30849afd07171c6870700917344b061edaf297)

lentzi90: I was also a bit wondering about this first. Previously we used ansible `template`, but this overwrites the file completely. This will instead render the template and insert it as a "block" in the file.

furkatgofurov7: ah yes, what we discussed offline, got it.

furkatgofurov7: CI runs are using the specific version and happy!
/approve

PR #1067, opened 2022-09-08
WIP Organize kubeadm netwrok config on centos

Rozzii:


Rozzii:
The issue I was testing here was solved.

PR #1070, opened 2022-09-12
Upgrade k8s from v1.23.8 instead of v1.24.0

mboukhalfa:
Instead of testing k8s upgrade between patch version we want to test upgrade between minor versions i.e:
instead of upgrade from v1.24.0 to v1.24.1 if should be from v1.23.8 to v1.24.1


furkatgofurov7 (commented):
Can we run feature tests first, please?

furkatgofurov7 (approved):

PR #1072, opened 2022-09-13
Add BMORELEASE as the default BMOBRANCH

kashifest:
This PR ensures that BMO is always checked out at the latest release. Only for main periodic test, BMOBRANCH will be set to main branch from Project Infra.

It also removes assuming that users will be providing different releasepaths because there arent any different releasepaths other than one only.

### DIFF
--- a/lib/releases.sh
+++ b/lib/releases.sh
@@ -27,9 +27,9 @@ function get_latest_release() {
 }
 
 # CAPM3, CAPI and BMO release path
-CAPM3RELEASEPATH="${CAPM3RELEASEPATH:-https://api.github.com/repos/${CAPM3_BASE_URL:-metal3-io/cluster-api-provider-metal3}/releases}"
-CAPIRELEASEPATH="${CAPIRELEASEPATH:-https://api.github.com/repos/${CAPI_BASE_URL:-kubernetes-sigs/cluster-api}/releases}"
-BMORELEASEPATH="${https://api.github.com/repos/${BMO_BASE_URL:-metal3-io/baremetal-operator}/releases}"
+CAPM3RELEASEPATH="{https://api.github.com/repos/${CAPM3_BASE_URL:-metal3-io/cluster-api-provider-metal3}/releases}"
+CAPIRELEASEPATH="{https://api.github.com/repos/${CAPI_BASE_URL:-kubernetes-sigs/cluster-api}/releases}"
+BMORELEASEPATH="{https://api.github.com/repos/${BMO_BASE_URL:-metal3-io/baremetal-operator}/releases}"
 
 # CAPM3, CAPI and BMO releases
 if [ "${CAPM3RELEASEBRANCH}" == "release-0.5" ] || [ "${CAPM3_VERSION}" == "v1alpha5" ]; then
--- a/lib/releases.sh
+++ b/lib/releases.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-set -x
-
 function get_latest_release() {
   set +x
   if [ -z "${GITHUB_TOKEN:-}" ]; then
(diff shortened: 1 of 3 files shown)


lib/releases.sh
```diff
@@ -42,6 +45,8 @@ else
   export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.2.")}"
 fi
 
+export BMORELEASE="${BMORELEASE:-$(get_latest_release "${BMORELEASEPATH}" "v0.1.")}"
+
 CAPIBRANCH="${CAPIBRANCH:-${CAPIRELEASE}}"
```
furkatgofurov7, line 50: most probably you have to set BMOBRANCH here same way since you are not setting any default in common.sh and this file is not sourced later when needed.

kashifest: Indeed, added that, lets see.

kashifest: Its working it seems. Thanks

lib/releases.sh
```diff
@@ -26,11 +26,12 @@ function get_latest_release() {
...
-# CAPM3 and CAPI release path
-CAPM3RELEASEPATH="${CAPM3RELEASEPATH:-https://api.github.com/repos/${CAPM3_BASE_URL:-metal3-io/cluster-api-provider-metal3}/releases}"
-CAPIRELEASEPATH="${CAPIRELEASEPATH:-https://api.github.com/repos/${CAPI_BASE_URL:-kubernetes-sigs/cluster-api}/releases}"
+# CAPM3 , CAPI and BMO release path
+CAPM3RELEASEPATH="{https://api.github.com/repos/${CAPM3_BASE_URL:-metal3-io/cluster-api-provider-metal3}/releases}"
+CAPIRELEASEPATH="{https://api.github.com/repos/${CAPI_BASE_URL:-kubernetes-sigs/cluster-api}/releases}"
+BMORELEASEPATH="{https://api.github.com/repos/${BMO_BASE_URL:-metal3-io/baremetal-operator}/releases}"
 
-# CAPM3 and CAPI releases
+# CAPM3 CAPI and BMO releases
```
furkatgofurov7, line 34:
```suggestion
# CAPM3, CAPI and BMO releases
```

lib/releases.sh
```diff
@@ -26,11 +26,12 @@ function get_latest_release() {
   echo "$release_tag"
 }
 
-# CAPM3 and CAPI release path
-CAPM3RELEASEPATH="${CAPM3RELEASEPATH:-https://api.github.com/repos/${CAPM3_BASE_URL:-metal3-io/cluster-api-provider-metal3}/releases}"
-CAPIRELEASEPATH="${CAPIRELEASEPATH:-https://api.github.com/repos/${CAPI_BASE_URL:-kubernetes-sigs/cluster-api}/releases}"
+# CAPM3 , CAPI and BMO release path
```
furkatgofurov7, line 29:
```suggestion
# CAPM3, CAPI and BMO release paths
```

furkatgofurov7 (commented):
Few last nits, otherwise this looks good:

kashifest:
@furkatgofurov7 @Rozzii PTAL again

furkatgofurov7:
One last question to make sure we are on the same page: Are we planning to always run all the tests (main, release-1.1, or future ones, i.e release-1.2) with the `latest` BMO release version and `main branch` of BMO will be tested on the periodic job?

kashifest:
> One last question to make sure we are on the same page: Are we planning to always run all the tests (main, release-1.1, or future ones, i.e release-1.2) with the `latest` BMO release version and `main branch` of BMO will be tested on the periodic job?

Yes thats what we agreed. BMO main branch will be tested in the main branch periodic test only.

furkatgofurov7 (approved):
/approve

smoshiur1237 (commented):
/lgtm

PR #1073, opened 2022-09-13
Fix commands for running keepalived monitor

lentzi90:
Systemd has changed so that it is no longer possible to first link a unit and then enable it by its name. Instead we use the full path for enabling. Enabling also creates a link so linkin is not necessary.


lentzi90:
Same fix was already done for e2e tests in https://github.com/metal3-io/cluster-api-provider-metal3/pull/711/files#diff-96ef3588f6a503604835e4f0b3eaa8008ccea1a973d17244e502290f5571e168

kashifest (approved):

PR #1076, opened 2022-09-15
Improve the local image branch config

Rozzii:
The configuration logic that is changed in this
commit is used when container images are built
locally in metal3-dev-env.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -223,14 +223,11 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
   # Is it a git repo?
   if [[ "$IMAGE" =~ "://" ]] ; then
     REPOPATH=~/${IMAGE##*/}
-    # Clone to ~ if not there already
-    if [ -e "${REPOPATH}" ]; then
-       cd "${REPOPATH}" || exit
-    else
-      git clone "${IMAGE}" "${REPOPATH}"
-      cd "${REPOPATH}" || exit
-      [ "${BRANCH}" = "main" ] || git checkout "${BRANCH}"
-    fi
+    # Clone to ~
+    # Clean the directory if it already exists
+    rm --recursive --force "${REPOPATH}"
+    git clone --branch "${BRANCH}" "${IMAGE}" "${REPOPATH}"
+    cd "${REPOPATH}" || exit
   # Assume it is a path
   else
     cd "${IMAGE}" || exit
--- a/vars.md
+++ b/vars.md
@@ -89,6 +89,23 @@ assured that they are persisted.
 | MARIADB_CAKEY_FILE | Path to the CA key of MariaDB | | /opt/metal3-dev-env/certs/ironic-ca.key |
 | MARIADB_CACERT_FILE | Path to the CA certificate of MariaDB | | /opt/metal3-dev-env/certs/ironic-ca.pem |
 
+## Local images
+
+Environment variables with `_LOCAL_IMAGE` in their name are used to specify either git repositories or directories that contain the code that is used to build the components locally e.g. `CAPM3_LOCAL_IMAGE`.
+
+Branches can be specified for the aforementioned repositories by following the same naming convention
+e.g. `CAPM3_LOCAL_IMAGE_BRANCH`.
+
+In case a git repository is specified instead of a local path the repository will be pulled and
+controlled by dev-env thus all manual changes done in the related directory will be disregarded by
+the build process.
+
+The intended use of git repository adresses for local image building is to stay synced with
+a branch of the component's upstream repository.
+
+In case there is a need to specify a directory that will keep local changes during the image building
+process just use local path instead of a git repository address.
+
 ## Additional networks
 
 By default two libvirt networks are created `baremetal` and `provisioning`


02_configure_host.sh
```diff
@@ -226,10 +226,12 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
     # Clone to ~ if not there already
     if [ -e "${REPOPATH}" ]; then
        cd "${REPOPATH}" || exit
+       git fetch upstream
+       git checkout "${BRANCH}"
```
mboukhalfa, line 230: will not this fail if the where any uncommitted changes not sure if this case can happen

lentzi90: Yes I'm not convinced about this either. Do we have `upstream` actually or is it just `origin`?

lentzi90 (commented):
IMO we need to add these folders to the cleanup in the BML playbook in project-infra also. That avoids any kinds of conflicts that could come from checkout or pull --rebase.
As an example of when issues could arise, say I force push to the branch I'm testing. There is most likely then a conflict while pulling.

02_configure_host.sh
Rozzii: these commands are executed on working directories that are created and by metal3-dev-env by default in the home directory just to use for the local image builds. This feature is only enabled if you do something like this ==> CAPM3_LOCAL_IMAGE=https://github.com/metal3-io/cluster-api-provider-metal3.git because the code checks the "://" section of the CAPM3_LOCAL_IMAGE and any other *_LOCAL_IMAGE variable. If it finds that substring then it enables this feature. So if you would need to specify a local directory where you can fiddle with the code then you just give the path to the code without the like CAPM3_LOCAL_IMAGE=/home/user/capm3

furkatgofurov7: +1, it is highly error-prone IMO

Rozzii:
> IMO we need to add these folders to the cleanup in the BML playbook in project-infra also. That avoids any kinds of conflicts that could come from checkout or pull --rebase. As an example of when issues could arise, say I force push to the branch I'm testing. There is most likely then a conflict while pulling.

Yeah there is no issue from my part with deleting these repos in project infra for bml but I think this feature  was not implemented with the aim to handle manual changes. I think these repos that are created if one enables this specific feature by using git adresses instead of a path that means that there is no plan to modify it and the goal is to work with a specifc upstream branch.

02_configure_host.sh
Rozzii: This should be not touched by anyone manually, if a git address is used as image path then that means that you want to stick to given upstream branch. If you want to modify it locally then you just specify a repository path and there will be no git issues.

Rozzii:
I see where the confusion comes from in relation to handling changes so i have added cleanup commands so if someone touches the repos that were created by the local image build process the changes will be discarded. This feature IMO does not supposed to handle manual on the spot changes. @lentzi90  @mboukhalfa  @furkatgofurov7

02_configure_host.sh
Rozzii: btw @lentzi90  thans for spotting the "upstream" I have fixed it.

mboukhalfa:
> I see where the confusion comes from in relation to handling changes so i have added cleanup commands so if someone touches the repos that were created by the local image build process the changes will be discarded. This feature IMO does not supposed to handle manual on the spot changes. I think this way it is clear that there is no point touching the directories @lentzi90 @mboukhalfa @furkatgofurov7

if it is the case then is not it simpler to just force clone the repo even if it does exist ? rather than check if it exist update ..

Rozzii:
> > I see where the confusion comes from in relation to handling changes so i have added cleanup commands so if someone touches the repos that were created by the local image build process the changes will be discarded. This feature IMO does not supposed to handle manual on the spot changes. I think this way it is clear that there is no point touching the directories @lentzi90 @mboukhalfa @furkatgofurov7
> 
> if it is the case then is not it simpler to just force clone the repo even if it does exist ? rather than check if it exist update ..

Well I guess you are right I will modify it!

vars.md
```diff
@@ -89,6 +89,23 @@ assured that they are persisted.
...
+e.g. `CAPM3_LOCAL_IMAGE_BRANCH`.
+
+In case a git repository is specified instead of a local path the repository will be pulled and
+controlled by dev-env thus all manual changes done in the related directory will be disregarded by
+the build process.
+
+The intended use of git repository adresses for local image building is to stay synced with
+a branch of the component's upstream repository.
+
+In case there is a need to specify a directory that will keep local cahges during the image building
```
mboukhalfa, line 106:
```suggestion
In case there is a need to specify a directory that will keep local changes during the image building
```

Rozzii: Done

lentzi90 (commented):
/lgtm
Thanks for fixing this @Rozzii !

I just want to clarify one thing about why the cleaning is necessary, for future reference. What I said about force push was not meant as "in this local folder". The reason we need the cleanup is because there can be problems in persistent environments if the folders are not cleaned. Here is an example:

1. I work on a branch `bug-fix` on my laptop. Push it, create a PR.
2. Then I decide to test it in a VM, I create the VM, set the proper `_LOCAL_IMAGE` vars and run the tests.
3. There is some issue with my patch, so I change (on my laptop) and force push on the `bug-fix` branch.
4. Now, if the VM is still running and didn't clean its local repo, there will be conflicts. The branch it has locally has diverged (because of my force push) from origin.

This is why I think we need the cleaning.

mboukhalfa:
LGTM

kashifest (approved):

PR #1080, opened 2022-09-23
bug host_cleanup.sh: Make baremetal net deletion in Centos optional

### LINKED ISSUE #1079: Removal of  'baremetal' connection during host cleanup fails in CentOS runs
After we changed network settings from network-scripts to key file format for CentOS builds in https://github.com/metal3-io/metal3-dev-env/commit/5bbb97f3c82d4c95a2fd3f479d5987dbd9d12d41 I have accidentally noticed in CI that removing of `baremetal` connection is failing:

```
+ sudo nmcli con delete provisioning
Connection 'provisioning' (545aceb3-0b16-34f7-b044-1d4950d9627b) successfully deleted.
+ '[' y == y ']'
+ sudo nmcli con delete baremetal
Error: unknown connection 'baremetal'.
Error: cannot delete unknown connection(s): 'baremetal'.
```

however, removal of `provisioning` does not! Most probably there is no need to remove the baremetal connection at all or not sure :)

**Additional info:**

- Xref to failing code: https://github.com/metal3-io/metal3-dev-env/blob/2670dc323698d661c4660cdccf325115202196ba/host_cleanup.sh#L60-L63
- CI run where you can see the error: https://jenkins.nordix.org/job/metal3_metal3_dev_env_release-1-2_integration_test_centos/3/consoleFull (ctrl + F for  `sudo nmcli con delete baremetal`)

mquhuy:
This network should be cleaned up by the "Delete libvirt networks" ansible job, so this step is likely redundant. For now keeping it as an optional step.

Fixes #1079

### DIFF
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -57,9 +57,10 @@ if [[ $OS == "centos" || $OS == "rhel" ]]; then
   if [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
       sudo nmcli con delete provisioning
   fi
-  # Leaving this around causes issues when the host is rebooted
+  # Baremetal net should have been cleaned already at this stage, but we double
+  # check as leaving it around causes issues when the host is rebooted
   if [ "$MANAGE_BR_BRIDGE" == "y" ]; then
-      sudo nmcli con delete baremetal
+      sudo nmcli con delete baremetal || true
   fi
 fi
 


mboukhalfa:
/retitle :bug: host_cleanup.sh: Skip baremetal net deletion in Centos
In fact we are not skipping the deletion but skipping the reporting error

mquhuy:
/retitle bug host_cleanup.sh: Make baremetal net deletion in Centos optional

mquhuy:
> /retitle bug host_cleanup.sh: Skip baremetal net deletion in Centos In fact we are not skipping the deletion but skipping the reporting the failure

PR #1082, opened 2022-09-27
Delete only the Cluster when cleaning up

lentzi90:
We used to delete the complete manifests, including MachineDeployment, KubeadmControlPlane, Metal3Cluster, etc. Deleting the Cluster is enough as all other resources will follow from the owner chain. This should avoid any issues with things deleted in the wrong order also.
This is also the recommended way to clean up in the CAPI book: https://cluster-api.sigs.k8s.io/user/quick-start.html#clean-up

### DIFF
--- a/tests/roles/run_tests/tasks/main.yml
+++ b/tests/roles/run_tests/tasks/main.yml
@@ -45,27 +45,13 @@
   include_tasks: move_back.yml
   when: v1aX_integration_test_action in repivot_actions
 
-- name: Deprovision worker nodes
-  kubernetes.core.k8s:
-    state: absent
-    src: "{{ TEMP_GEN_DIR }}/{{ CAPM3_VERSION }}_workers_{{ IMAGE_OS }}.yaml"
-    namespace: "{{ NAMESPACE }}"
-  ignore_errors: yes
-  when: v1aX_integration_test_action in deprovision_workers_actions
-
-- name: Deprovision control plane
-  kubernetes.core.k8s:
-    state: absent
-    src: "{{ TEMP_GEN_DIR }}/{{ CAPM3_VERSION }}_controlplane_{{ IMAGE_OS }}.yaml"
-    namespace: "{{ NAMESPACE }}"
-  ignore_errors: yes
-  when: v1aX_integration_test_action in deprovision_controlplane_actions
-
 - name: Deprovision cluster
   kubernetes.core.k8s:
     state: absent
-    src: "{{ TEMP_GEN_DIR }}/{{ CAPM3_VERSION }}_cluster_{{ IMAGE_OS }}.yaml"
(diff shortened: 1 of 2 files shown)


lentzi90:
Looks good! I will run the tests a couple of times more to see if we hit any flakes.
/test-ubuntu-integration-main

lentzi90:
Two more successful runs. :heavy_check_mark: 

/test-ubuntu-integration-main

kashifest:
Please run the feature tests also

tests/roles/run_tests/tasks/main.yml
```diff
@@ -45,27 +45,13 @@
...
-    state: absent
-    src: "{{ TEMP_GEN_DIR }}/{{ CAPM3_VERSION }}_controlplane_{{ IMAGE_OS }}.yaml"
-    namespace: "{{ NAMESPACE }}"
-  ignore_errors: yes
-  when: v1aX_integration_test_action in deprovision_controlplane_actions
-
 - name: Deprovision cluster
   kubernetes.core.k8s:
     state: absent
-    src: "{{ TEMP_GEN_DIR }}/{{ CAPM3_VERSION }}_cluster_{{ IMAGE_OS }}.yaml"
```
mboukhalfa, line 67: Not sure this was used to delete also the IPPools ? are they deleted automatically when deleting the cluster

lentzi90: They should be deleted when the cluster is deleted because they have an ownerRef to it.

kashifest:
Do we need to explicitly check if controlplane/s and worker/s are getting deprovisioned properly or not ? Or is it checked already

lentzi90:
It is checked already here: https://github.com/metal3-io/metal3-dev-env/blob/main/tests/roles/run_tests/tasks/cleanup.yml#L25-L93
We can of course add more. Now we check and wait for the following:
1. Cluster is deprovisioning
2. All BMHs become available
3. No Metal3Machines left
4. No Machines left
5. No Metal3Cluster left
6. No Cluster left

So we could for example check also the KubeadmControlPlane and MachineDeployment. Also maybe IPPool?

kashifest:
> It is checked already here: https://github.com/metal3-io/metal3-dev-env/blob/main/tests/roles/run_tests/tasks/cleanup.yml#L25-L93 We can of course add more. Now we check and wait for the following:
> 
> 1. Cluster is deprovisioning
> 2. All BMHs become available
> 3. No Metal3Machines left
> 4. No Machines left
> 5. No Metal3Cluster left
> 6. No Cluster left
> 
> So we could for example check also the KubeadmControlPlane and MachineDeployment. Also maybe IPPool?

IMO those checks are enough. Thanks.
/approve

PR #1083, opened 2022-09-28
Improve metal3-dev-env repository cloning

Rozzii:
Two issues are being fixed by this commit.

1. Currently the dev-env have 2 independent repo
cloning functionality. One for building local images
and one for customizing CRDs.

The two cloning process is incompatible because the cloning
used for CRD customization happens after the images
specified with _LOCAL_IMAGE postfix have been
built so a clean environment where the repositories
are not yet cloned to $HOME/go/src/github.com/metal3-io/
would not support the local image building based on the same repos
that were used for CRD customization.

2. The clone_repo bash function now have support cloning
specific commits from branches of CAPI, CAPM3 and IPAM.

NOTE: With the changes in this commit the git repo support
implemented exclusively for _LOCAL_IMAGE variables will become redundant
because the _LOCAL_IMAGEs can use the same repositories that are
now cloned by the clone_repo functions thus the redundant functionality
should be removed in later commits in order to keep the scope of this
commit manageable.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -7,6 +7,8 @@ source lib/logging.sh
 source lib/common.sh
 # shellcheck disable=SC1091
 source lib/network.sh
+# shellcheck disable=SC1091
+source lib/releases.sh
 
 # Root needs a private key to talk to libvirt
 # See tripleo-quickstart-config/roles/virtbmc/tasks/configure-vbmc.yml
@@ -197,6 +199,15 @@ elif [[ "$reg_state" != "running" ]]; then
   sudo "${CONTAINER_RUNTIME}" run -d -p "${REGISTRY_PORT}":5000 --name registry "$DOCKER_REGISTRY_IMAGE"
 fi
 sleep 5
+
+
+# Clone all needed repositories (CAPI, CAPM3, BMO, IPAM)
+mkdir -p "${M3PATH}"
+clone_repo "${BMOREPO}" "${BMOBRANCH}" "${BMOPATH}" "${BMOCOMMIT}"
+clone_repo "${CAPM3REPO}" "${CAPM3BRANCH}" "${CAPM3PATH}" "${CAPM3COMMIT}"
+clone_repo "${IPAMREPO}" "${IPAMBRANCH}" "${IPAMPATH}" "${IPAMCOMMIT}"
+clone_repo "${CAPIREPO}" "${CAPIBRANCH}" "${CAPIPATH}" "${CAPICOMMIT}"
+
 
 # Pushing images to local registry
 for IMAGE_VAR in $(env | grep -v "_LOCAL_IMAGE=" | grep "_IMAGE=" | grep -o "^[^=]*") ; do
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -89,21 +89,29 @@ export SSH_PUB_KEY_CONTENT
 
 FILESYSTEM=${FILESYSTEM:="/"}
 
-# Environment variables
-# M3PATH : Path to clone the Metal3 Development Environment repository
-# BMOPATH : Path to clone the Bare Metal Operator repository
-# CAPM3PATH : Path to clone the Cluster API Provider Metal3 repository
-# CAPIPATH : Path to clone the Cluster API repository
-#
-# BMOREPO : Baremetal Operator repository URL
-# BMOBRANCH : Baremetal Operator repository branch to checkout (its set it lib/releases.sh)
-# CAPM3REPO : Cluster API Provider Metal3 repository URL
-# CAPM3BRANCH : Cluster API Provider Metal3 repository branch to checkout
-# FORCE_REPO_UPDATE : discard existing directories
-#
-# BMO_RUN_LOCAL : run the Baremetal Operator locally (not in Kubernetes cluster)
-# CAPM3_RUN_LOCAL : run the Cluster API Provider Metal3 locally
+# Reusable repository cloning function
+function clone_repo() {
+  local REPO_URL="$1"
+  local REPO_BRANCH="$2"
+  local REPO_PATH="$3"
+  local REPO_COMMIT="${4:-HEAD}"
 
+  if [[ -d "${REPO_PATH}" && "${FORCE_REPO_UPDATE}" == "true" ]]; then
+    rm -rf "${REPO_PATH}"
+  fi
+  if [ ! -d "${REPO_PATH}" ] ; then
+    pushd "${M3PATH}" || exit
+    git clone "${REPO_URL}" "${REPO_PATH}"
+    popd || exit
+    pushd "${REPO_PATH}" || exit
+    git checkout "${REPO_BRANCH}"
+    git checkout "${REPO_COMMIT}"
+    git pull -r || true
+    popd || exit
+  fi
+}
+
+# Configure common environment variables
 CAPM3_VERSION_LIST="v1alpha5 v1beta1"
 export CAPM3_VERSION="${CAPM3_VERSION:-"v1beta1"}"
 
@@ -155,6 +163,9 @@ BMOREPO="${BMOREPO:-https://github.com/metal3-io/baremetal-operator.git}"
 export BMO_BASE_URL="${BMO_BASE_URL:-metal3-io/baremetal-operator}"
 FORCE_REPO_UPDATE="${FORCE_REPO_UPDATE:-true}"
 BMOCOMMIT="${BMOCOMMIT:-HEAD}"
+CAPM3COMMIT="${CAPM3COMMIT:-HEAD}"
+IPAMCOMMIT="${IPAMCOMMIT:-HEAD}"
+CAPICOMMIT="${CAPICOMMIT:-HEAD}"
 
 BMO_RUN_LOCAL="${BMO_RUN_LOCAL:-false}"
 CAPM3_RUN_LOCAL="${CAPM3_RUN_LOCAL:-false}"
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -22,45 +22,6 @@ source lib/ironic_tls_setup.sh
 # shellcheck disable=SC1091
 source lib/ironic_basic_auth.sh
 
-# -----------------------
-# Repositories management
-# -----------------------
-
-#
-# Clone and checkout a repo
-#
-function clone_repo() {
-  local REPO_URL="$1"
-  local REPO_BRANCH="$2"
-  local REPO_PATH="$3"
-  local REPO_COMMIT="${4:-HEAD}"
-
-  if [[ -d "${REPO_PATH}" && "${FORCE_REPO_UPDATE}" == "true" ]]; then
-    rm -rf "${REPO_PATH}"
-  fi
-  if [ ! -d "${REPO_PATH}" ] ; then
-    pushd "${M3PATH}"
-    git clone "${REPO_URL}" "${REPO_PATH}"
-    popd
-    pushd "${REPO_PATH}"
-    git checkout "${REPO_BRANCH}"
-    git checkout "${REPO_COMMIT}"
-    git pull -r || true
-    popd
-  fi
-}
-
-#
-# Clone all needed repositories
-#
-function clone_repos() {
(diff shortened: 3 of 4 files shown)


lentzi90 (commented):
This looks good to me. But I want to point out one thing:

> The cloning process happens after the images specified with _LOCAL_IMAGE postfix have been built so in a clean environment where the repositories are not yet cloned to $HOME/go/src/github.com/metal3-io/ the local building won't work.

I don't think this is true. The thing is we are cloning multiple times. One time to build the "local" images and again later when launching the management cluster. Here is the cloning for building the image: https://github.com/metal3-io/metal3-dev-env/blob/main/02_configure_host.sh#L224-L230

But anyway, I think this is an improvement and I look forward to getting rid of the extra clone at some point in the future :smile: 

/lgtm

Rozzii:
> This looks good to me. But I want to point out one thing:
> 
> > The cloning process happens after the images specified with _LOCAL_IMAGE postfix have been built so in a clean environment where the repositories are not yet cloned to $HOME/go/src/github.com/metal3-io/ the local building won't work.
> 
> I don't think this is true. The thing is we are cloning multiple times. One time to build the "local" images and again later when launching the management cluster. Here is the cloning for building the image: https://github.com/metal3-io/metal3-dev-env/blob/main/02_configure_host.sh#L224-L230
> 
> But anyway, I think this is an improvement and I look forward to getting rid of the extra clone at some point in the future smile
> 
> /lgtm

Yeah I agree I think we/I will remove the section that you have highlighted as it is a duplication IMO also.

smoshiur1237 (commented):
Nice improvement.
LGTM

lib/common.sh
```diff
@@ -90,20 +90,52 @@ export SSH_PUB_KEY_CONTENT
...
 
 # Environment variables
-# M3PATH : Path to clone the Metal3 Development Environment repository
-# BMOPATH : Path to clone the Bare Metal Operator repository
+# M3PATH    : Path to clone the Metal3 Development Environment repository
+# BMOPATH   : Path to clone the Bare Metal Operator repository
 # CAPM3PATH : Path to clone the Cluster API Provider Metal3 repository
-# CAPIPATH : Path to clone the Cluster API repository
+# CAPIPATH  : Path to clone the Cluster API repository
+# IPAMPATH  : Path to clone Ip Address Manager
```
furkatgofurov7, line 97:
```suggestion
# IPAMPATH  : Path to clone IP Address Manager repository
```

lib/common.sh
```diff
@@ -90,20 +90,52 @@ export SSH_PUB_KEY_CONTENT
...
+#
+# CAPIREPO    : Cluster API git repository URL
+# CAPIBRANCH  : Cluster API git repository branch to checkout
+# CAPICOMMIT  : Cluster API git commit to check out on CAPIBRANCH
+# BMOREPO     : Baremetal Operator git repository URL
+# BMOBRANCH   : Baremetal Operator git repository branch to checkout (its set it lib/releases.sh)
```
furkatgofurov7, line 103:
```suggestion
# BMOBRANCH   : Baremetal Operator git repository branch to checkout (set in lib/releases.sh)
```

lib/common.sh
```diff
@@ -90,20 +90,52 @@ export SSH_PUB_KEY_CONTENT
...
+# BMOCOMMIT   : BMO git commit to check out on BMOBRANCH
+# CAPM3REPO   : Cluster API Provider Metal3 git repository URL
+# CAPM3BRANCH : Cluster API Provider Metal3 git repository branch to checkout
+# CAPM3COMMIT : CAPM3 git commit to check out on CAPM3BRANCH
```
furkatgofurov7, line 107:
```suggestion
# CAPM3COMMIT : Cluster API Provider Metal3 git commit to check out on CAPM3BRANCH
```

lib/common.sh
```diff
@@ -90,20 +90,52 @@ export SSH_PUB_KEY_CONTENT
 FILESYSTEM=${FILESYSTEM:="/"}
 
 # Environment variables
-# M3PATH : Path to clone the Metal3 Development Environment repository
```
furkatgofurov7, line 93: If we need to keep these all in https://github.com/metal3-io/metal3-dev-env/blob/main/vars.md, then we need to update it (CAPI, IPAM repo/branch notes + with commit vars for each)

lib/common.sh
```diff
@@ -90,20 +90,52 @@ export SSH_PUB_KEY_CONTENT
...
+# CAPICOMMIT  : Cluster API git commit to check out on CAPIBRANCH
+# BMOREPO     : Baremetal Operator git repository URL
+# BMOBRANCH   : Baremetal Operator git repository branch to checkout (its set it lib/releases.sh)
+# BMOCOMMIT   : BMO git commit to check out on BMOBRANCH
+# CAPM3REPO   : Cluster API Provider Metal3 git repository URL
+# CAPM3BRANCH : Cluster API Provider Metal3 git repository branch to checkout
+# CAPM3COMMIT : CAPM3 git commit to check out on CAPM3BRANCH
+# IPAMREPO    : IP Address Manager git repository URL
+# IPAMBRANCH  : IP Address Manager git repository branch to checkout
+# IPAMCOMMIT  : IP Address Manager git commit to check out on IPAMBRANCH
```
furkatgofurov7, line 110: maybe we should stick to one: `checkout` or `check out` everywhere (first looks better IMO)

furkatgofurov7 (approved):
/hold 

Thanks, this I always did locally :)
to address the comments, feel free to unhold when fixed

furkatgofurov7:
> 2\. The clone_repos bash function now have support cloning specific commits from branches of CAPI, CAPM3 and IPAM.

it is `clone_repo` which does that, `clone_repos` have been removed

lib/common.sh
Rozzii: I don't understand what you are saying  could you please clarify?

02_configure_host.sh
```diff
@@ -198,6 +200,15 @@ elif [[ "$reg_state" != "running" ]]; then
...
 sleep 5
 
+
+# Clone all needed repositories (CAPI, CAPM3, BMO, IPAM)
+mkdir -p "${M3PATH}"
+clone_repo "${BMOREPO}" "${BMOBRANCH}" "${BMOPATH}" "${BMOCOMMIT}"
+clone_repo "${CAPM3REPO}" "${CAPM3BRANCH}" "${CAPM3PATH}" "${CAPM3COMMIT}"
+clone_repo "${IPAMREPO}" "${IPAMBRANCH}" "${IPAMPATH}" "${IPAMCOMMIT}"
+clone_repo "${CAPIREPO}" "${CAPIBRANCH}" "${CAPIPATH}" "${CAPICOMMIT}"
+
```
kashifest, line 210: Keep in mind that 02 script was meant for configuring host and 03 for mgmt cluster. With this change , you would be taking a part of the  management cluster work here in 02. So the naming of the scripts would need to change perhaps? Also, how would it affect [make  delete_mgmt_cluster](https://github.com/metal3-io/metal3-dev-env/blob/main/Makefile#L17) and [make launch_mgmt_cluster](https://github.com/metal3-io/metal3-dev-env/blob/main/Makefile#L9) commands, cause currently it is only related to 03 script.

lib/common.sh
furkatgofurov7: Sorry, maybe I highlighted not the correct line. Request was: if you check the file, there we provide - CAPM3REPO/CAPM3BRANCH or BMOREPO/BMOBRANCH, but not IPAM or CAPI. Also, var `REPO`COMMIT is completely missing. So my suggestion is: either remove existing ones as well OR add missing ones to the list.

02_configure_host.sh
Rozzii: It simply clones repos to the host, the mgmt cluster customization still happens later, the repos are only used in potential local buildings in 02 nothing more. IMO this action should have been in  02 not in 03. 

No effect on cleanup this cloning is self contained as the clone_repo function handles the cleanup based on the FORCE_REPO_UPDATE var and this process is independent from the cluster cleanup.

Rozzii: To rephrase myself, repo cloning  IMO is not part of mgmt kustomization and lunching just as much as libvirt vm creation is not part of it. These actions should be completed before  we launch mgmt and not during.

mboukhalfa:
>     1. The cloning process happens after the images specified with _LOCAL_IMAGE postfix have been
>        built so in a clean environment where the repositories are not yet cloned to $HOME/go/src/github.com/metal3-io/ the local building won't work.

why you think on a clean environment the repos will not be cloned I see lines:

https://github.com/metal3-io/metal3-dev-env/blob/8ed7a28e079d95e883008d3d0e37ae0ba736db42/02_configure_host.sh#L227-L229
cloning the repo

02_configure_host.sh
kashifest: ok so I would be able to delete and launch mgmt cluster from the same state/settings irrespective of this as it is now. I agree with cloning should be a part of 02 since it is where image is being built.

Rozzii:
> > ```
> > 1. The cloning process happens after the images specified with _LOCAL_IMAGE postfix have been
> >    built so in a clean environment where the repositories are not yet cloned to $HOME/go/src/github.com/metal3-io/ the local building won't work.
> > ```
> 
> why you think on a clean environment the repos will not be cloned I see lines:
> 
> https://github.com/metal3-io/metal3-dev-env/blob/8ed7a28e079d95e883008d3d0e37ae0ba736db42/02_configure_host.sh#L227-L229
> 
> cloning the repo

The cloning will happen, the issue is just that the local image building can't rely on the cloned repos because the image building happens before these repos are cloned and that is why the image building had the support for the git repo cloning independent from these clonings  and that is why if someone wanted to use local build images for example in CI then dev-env cloned things twice.

mboukhalfa:
> > This looks good to me. But I want to point out one thing:
> > > The cloning process happens after the images specified with _LOCAL_IMAGE postfix have been built so in a clean environment where the repositories are not yet cloned to $HOME/go/src/github.com/metal3-io/ the local building won't work.
> > 
> > 
> > I don't think this is true. The thing is we are cloning multiple times. One time to build the "local" images and again later when launching the management cluster. Here is the cloning for building the image: https://github.com/metal3-io/metal3-dev-env/blob/main/02_configure_host.sh#L224-L230
> > But anyway, I think this is an improvement and I look forward to getting rid of the extra clone at some point in the future smile
> > /lgtm
> 
> Yeah I agree I think we/I will remove the section that you have highlighted as it is a duplication IMO also.

I do not think it is a duplication because on lines 228 we clean any existing repo the thing that the new cloning function does not do

Rozzii:
> > > This looks good to me. But I want to point out one thing:
> > > > The cloning process happens after the images specified with _LOCAL_IMAGE postfix have been built so in a clean environment where the repositories are not yet cloned to $HOME/go/src/github.com/metal3-io/ the local building won't work.
> > > 
> > > 
> > > I don't think this is true. The thing is we are cloning multiple times. One time to build the "local" images and again later when launching the management cluster. Here is the cloning for building the image: https://github.com/metal3-io/metal3-dev-env/blob/main/02_configure_host.sh#L224-L230
> > > But anyway, I think this is an improvement and I look forward to getting rid of the extra clone at some point in the future smile
> > > /lgtm
> > 
> > 
> > Yeah I agree I think we/I will remove the section that you have highlighted as it is a duplication IMO also.
> 
> I do not think it is a duplication because on lines 228 we clean any existing repo the thing that the new cloning function does not do

What you are talking about is a different cloning process , we clone into 2 different places and that doesn't make any sense I am already working na a followup that will remove the cloning prcess that you have pointed out all together.

(The cloning process that you are showing also has no commit support)

The endgoal is to have a single repo cloning process that gives the same repo/branch/commit to the local image building 02 script and to the CRD customization process in 03 script

mboukhalfa:
/lgtm
As moving the cloning function will be needed in the follow up PR 
Because as it is in the current changes moving the cloning to script 02 does not have a significant impact on the local_image cloning

Rozzii:
Fixed some typos and minor things in the commit message, please lgtm again.

lentzi90 (commented):
/lgtm

PR #1084, opened 2022-10-03
improve local build of CAPI,CAPM3,IPAM,BMO

Rozzii:
This commit improves 3 different things.

1. Add default build option for CAPI,CAPM3,IPAM,BMO.
If the default build option is set to true for any of the
aforementioned repositories then the dev-env will build
the container images based on the same directory that is used
for CRD customization.

2. In order to unify the local container image build process
with the CRD customization process this commit removes the git
repository supoort feature from 02 script because that was exclusively
wirrten to support the local image building thus duplicating the
repository cloning process and by default separating the directories
used for CRD customization and image building.

3. Fix CAPI build options for local image building.

This PR is the continuation of #1083

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -227,22 +227,8 @@ done
 
 # Support for building local images
 for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
-  BRANCH_IMAGE_VAR="${IMAGE_VAR}_BRANCH"
   IMAGE="${!IMAGE_VAR}"
-  BRANCH="${!BRANCH_IMAGE_VAR:-main}"
-
-  # Is it a git repo?
-  if [[ "$IMAGE" =~ "://" ]] ; then
-    REPOPATH=~/${IMAGE##*/}
-    # Clone to ~
-    # Clean the directory if it already exists
-    rm --recursive --force "${REPOPATH}"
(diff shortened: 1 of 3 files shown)


02_configure_host.sh
```diff
@@ -227,22 +227,8 @@ done
 
 # Support for building local images
 for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
-  BRANCH_IMAGE_VAR="${IMAGE_VAR}_BRANCH"
   IMAGE="${!IMAGE_VAR}"
-  BRANCH="${!BRANCH_IMAGE_VAR:-main}"
-
-  # Is it a git repo?
-  if [[ "$IMAGE" =~ "://" ]] ; then
```
elfosardo, line 235: why removing support for cloning from a git repo ?

elfosardo (changes requested):

02_configure_host.sh
Rozzii: It was a duplication, dev-env still supports git repo cloning but now both CRD customization and local image building will rely on the repo cloning process done in 02 script in the clone_repo functions. Previously in case of local builds we had situations where dev-env cloned everything 2 times into different directories. 

This change was prepared in  https://github.com/metal3-io/metal3-dev-env/pull/1083 PR.

elfosardo (approved):

lentzi90 (commented):
/lgtm

kashifest (approved):

PR #1086, opened 2022-10-04
Ironic kustomize components

lentzi90:
Changes needed for https://github.com/metal3-io/baremetal-operator/pull/1172

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -126,7 +126,7 @@ function launch_ironic() {
     # Currently we leave mariadb there even when using sqlite.
 
     # Update Configmap parameters with correct urls
-    cat << EOF | sudo tee "$IRONIC_DATA_DIR/ironic_bmo_configmap.env"
+    cat << EOF | sudo tee "${IRONIC_DATA_DIR}/ironic_bmo_configmap.env"
 HTTP_PORT=${HTTP_PORT}
 PROVISIONING_IP=${CLUSTER_PROVISIONING_IP}
 PROVISIONING_CIDR=${PROVISIONING_CIDR}
@@ -147,7 +147,7 @@ EOF
     echo "DEPLOY_ISO_URL=${DEPLOY_ISO_URL}" | sudo tee -a "${IRONIC_DATA_DIR}/ironic_bmo_configmap.env"
   fi
 
-  if [ "$NODES_PLATFORM" == "libvirt" ] ; then
+  if [[ "${NODES_PLATFORM}" == "libvirt" ]] ; then
     echo "IRONIC_KERNEL_PARAMS=console=ttyS0" | sudo tee -a "${IRONIC_DATA_DIR}/ironic_bmo_configmap.env"
   fi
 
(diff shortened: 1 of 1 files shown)


lentzi90:
Putting hold so we don't accidentally merge this before https://github.com/metal3-io/baremetal-operator/pull/1172 is also ready to merge.
/hold

TODO:
- [ ] Remove test commit once reviewers are happy with the tests.

03_launch_mgmt_cluster.sh
```diff
@@ -147,7 +147,8 @@ EOF
   fi
 
   # Copy the generated configmap for ironic deployment
-  cp "$IRONIC_DATA_DIR/ironic_bmo_configmap.env"  "${BMOPATH}/ironic-deployment/keepalived/ironic_bmo_configmap.env"
+  # We always use the keepalived component so this is where we put the configmap
```
mboukhalfa, line 150: Cannot we use default folder that will be the base of all the other instead of an independent component like keepalived or tls ..

lentzi90: No we cannot use the default as base for the other, because it has a different ConfigMap than the keepalived component. ConfigMaps cannot be easily merged, instead we replace it completely. So we use the [base](https://github.com/Nordix/baremetal-operator/tree/lentzi90/ironic-kustomize-components/ironic-deployment/base) as base instead since it has no ConfigMap at all.

In the future I would like to change this so that we create a temporary overlay in metal3-dev-env instead of messing with BMO, similar to what the [deploy.sh](https://github.com/Nordix/baremetal-operator/blob/lentzi90/ironic-kustomize-components/tools/deploy.sh#L101-L136) script is doing in my PR). But I think it makes sense to take this in first as it is a smaller change.

mboukhalfa: ok 
/lgtm

lib/common.sh
```diff
@@ -149,6 +149,9 @@ else
   IPAMBRANCH="${IPAMBRANCH:-main}"
 fi
 
+export BMOREPO="https://github.com/Nordix/baremetal-operator.git"
```
Rozzii, line 152: Will you remove these after the test is done? (I assume yes, but just in case)

lentzi90: Yes indeed! There is a separate commit for testing that I will drop once reviewers are happy :slightly_smiling_face:

03_launch_mgmt_cluster.sh
```diff
@@ -147,7 +147,8 @@ EOF
...
+  cp "$IRONIC_DATA_DIR/ironic_bmo_configmap.env"  "${BMOPATH}/ironic-deployment/components/keepalived/ironic_bmo_configmap.env"
```
tuminoid, line 151: Inconsistent use of curlys.

kashifest (approved):
/hold
Please address @tuminoid's comment and remove the test commit to unhold.

furkatgofurov7 (approved):
Overall looks good to me, except the commit message could be improved a bit (it contains WIP)

/approve

smoshiur1237 (commented):
This PR looks good, once you remove the test code,  will add the lgtm.

lentzi90:
Test commit removed. 
Once https://github.com/metal3-io/baremetal-operator/pull/1172 is merged we can run the tests here again and merge.

kashifest:
You have a merge conflict here @lentzi90

lentzi90:
Fixed!

kashifest:
/lgtm
We need to force merge this.

PR #1087, opened 2022-10-05
Add extra tag to locally built images

Rozzii:
The goal of the additional tag is to help debugging images that were built locally.


kashifest:
how would you test it if its working properly or not ?

lentzi90 (commented):
Nice! I would even want to replace that "latest" tag with this but maybe safer to start like this first :slightly_smiling_face: 
/lgtm

PR #1088, opened 2022-10-06
Improve local ironic image building

Rozzii:


Rozzii:
/hold
I have to add a few things.

lentzi90 (commented):
/lgtm
/hold
Did you manage to test this locally?

Rozzii:
> /lgtm /hold Did you manage to test this locally?

Yeah I tested it with a positive case, I haven't checked negative cases. So I have checked if I set IRONIC_FROM_SOURCE it behaves as expected. We can keep the hold for a bit longer until I run a few more variations of the test if you would like me to do so.

lentzi90:
It would be good to check that we don't get any errors from unset variables when we run without them set. But now that I think about it, I guess this is already tested in the integration test, right?

Rozzii:
> It would be good to check that we don't get any errors from unset variables when we run without them set. But now that I think about it, I guess this is already tested in the integration test, right?

Yes, I think simple unset variable situation has been checked, I will try to think about some more complex case(s) that could happen and I will check them.

lentzi90:
Thanks! Feel free to unhold when ready :slightly_smiling_face:

smoshiur1237 (commented):
/lgtm

kashifest (commented):
I am confused why do we need 2 different vars BUILD_IRONIC_LOCALLY and BUILD_IRONIC_IMAGE_LOCALLY and also when to check IRONIC_FROM_SOURCE and when not? Because while cloning you are checking both but while building it seems you are not.

Rozzii:
> I am confused why do we need 2 different vars BUILD_IRONIC_LOCALLY and BUILD_IRONIC_IMAGE_LOCALLY and also when to check IRONIC_FROM_SOURCE and when not? Because while cloning you are checking both but while building it seems you are not.

Yes it was a typo, thanks for mentioning.

lib/common.sh
```diff
@@ -153,24 +158,32 @@ export IPAMPATH="${IPAMPATH:-${M3PATH}/ip-address-manager}"
...
+export IRONIC_IMAGE_PATH="${IRONIC_IMAGE_PATH:-/tmp/ironic-image}"
+export IRONIC_IMAGE_REPO="${IRONIC_IMAGE_REPO:-https://github.com/metal3-io/ironic-image.git}"
+export IRONIC_IMAGE_BRANCH="${IRONIC_IMAGE_BRANCH:-main}"
+export IRONIC_IMAGE_COMMIT="${IRONIC_IMAGE_COMMIT:-HEAD}"
+
+export BUILD_CAPM3_LOCALLY="${BUILD_CAPM3_LOCALLY:-false}"
+export BUILD_IPAM_LOCALLY="${BUILD_IPAM_LOCALLY:-false}"
+export BUILD_BMO_LOCALLY="${BUILD_BMO_LOCALLY:-false}"
+export BUILD_CAPI_LOCALLY="${BUILD_CAPI_LOCALLY:-false}"
+export BUILD_IRONIC_IMAGE_LOCALLY="${BUILD_IRONIC_IMAGE_LOCALLY:-false}"
```
kashifest, line 179: Is there any reason to add` _IMAGE` for ironic only and not for other components?

Rozzii: Well nothing super special, only that the repo name is ironic-image. I was also debating whether it should be just IRONIC but Ironic is written in python so the application itself is never built strictly speaking. But I am fine with just IRONIC if you would like to change to that.

lentzi90 (commented):
/lgtm
I'm fine with either BUILD_IRONIC_IMAGE_LOCALLY or BUILD_IRONIC_LOCALLY.

lib/common.sh
kashifest: Not a big deal.
/approve

PR #1094, opened 2022-10-24
Pin cert manager to v1.10.0 in the CI

furkatgofurov7:
/hold


kashifest (approved):

lentzi90 (commented):
/lgtm
I think we may need a PR also in CAPM3 if we want to use it for all e2e tests, or maybe it is using the same config file?

furkatgofurov7:
> I think we may need a PR also in CAPM3 if we want to use it for all e2e tests, or maybe it is using the same config file?

e2e upgrade tests use an older version of it on purpose AFAIR https://github.com/metal3-io/cluster-api-provider-metal3/pull/755/files#r997285628 otherwise bootstrapping dev env when starting normal e2e should use the latest. 
However, other files, i.e Tiltfile, and scripts under hack can be updated.

PR #1098, opened 2022-11-07
IPA downloader exit on error

lentzi90:
We used to wait for the container to finish, which just prints the exit code. If we instead run it without -d, it will block on that step and propagate the exit code properly.

Hides the output since it is very verbose.
Adds retry logic (including removing the container after every run in order to be able to run it again).

I think the reason we did it with -d and wait originally was to run things in parallel [as seen here](https://github.com/metal3-io/metal3-dev-env/blob/efd81b30111bfe905e7245ee0419beb4e49680d5/02_configure_host.sh#L14-L21). Two containers were started in parallel before we waited for the IPA downloader to finish. But this is no longer the case so we can now drop the -d and wait.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -239,14 +239,17 @@ for IMAGE_VAR in $(env | grep -v "_LOCAL_IMAGE=" | grep "_IMAGE=" | grep -o "^[^
   pull_container_image_if_missing "$IMAGE"
  done
 
-if ${IPA_DOWNLOAD_ENABLED}; then
-    # Start image downloader container
-    #shellcheck disable=SC2086
-    sudo "${CONTAINER_RUNTIME}" run -d --net host --name ipa-downloader ${POD_NAME} \
-       -e IPA_BASEURI="$IPA_BASEURI" \
-       -v "$IRONIC_DATA_DIR":/shared "${IPA_DOWNLOADER_IMAGE}" /usr/local/bin/get-resource.sh
-
-    sudo "${CONTAINER_RUNTIME}" wait ipa-downloader
+if ${IPA_DOWNLOAD_ENABLED} || [ ! -f "${IRONIC_DATA_DIR}/html/images/ironic-python-agent.kernel" ]; then
+    # Run image downloader container. The output is very verbose and not that interesting so we hide it.
+    for i in {1..5}; do
+        echo "Attempting to download IPA. $i/5"
+        #shellcheck disable=SC2086
+        sudo "${CONTAINER_RUNTIME}" run --rm --net host --name ipa-downloader ${POD_NAME} \
+          -e IPA_BASEURI="$IPA_BASEURI" \
+          -v "$IRONIC_DATA_DIR":/shared "${IPA_DOWNLOADER_IMAGE}" \
+          /bin/bash -c "/usr/local/bin/get-resource.sh &> /dev/null" && s=0 && break || s=$?
+    done
+    (exit $s)
 fi
 
 function configure_minikube() {


kashifest:
/approve
Thanks @lentzi90 for fixing the CI.

furkatgofurov7 (commented):
Lgtm pending a squash

lentzi90:
> Lgtm pending a squash

Oh will do! Just dropped the test commit and forgot to squash the rest...

kashifest:
> Lgtm pending a squash

/lgtm

01_prepare_host.sh
```diff
@@ -239,14 +239,17 @@ for IMAGE_VAR in $(env | grep -v "_LOCAL_IMAGE=" | grep "_IMAGE=" | grep -o "^[^
...
 
-if ${IPA_DOWNLOAD_ENABLED}; then
-    # Start image downloader container
-    #shellcheck disable=SC2086
-    sudo "${CONTAINER_RUNTIME}" run -d --net host --name ipa-downloader ${POD_NAME} \
-       -e IPA_BASEURI="$IPA_BASEURI" \
-       -v "$IRONIC_DATA_DIR":/shared "${IPA_DOWNLOADER_IMAGE}" /usr/local/bin/get-resource.sh
-
-    sudo "${CONTAINER_RUNTIME}" wait ipa-downloader
+if ${IPA_DOWNLOAD_ENABLED} || [ ! -f "${IRONIC_DATA_DIR}/html/images/ironic-python-agent.kernel" ]; then
```
tuminoid, line 242: Not a fan of this if statement, as it executes whatever command is entered in IPA_DOWNLOAD_ENABLED env. It happens to work because true|false are commands too. 

if [ "${IPA_DOWNLOAD_ENABLED}" = true ] makes it a string comparison.

furkatgofurov7 (commented):
/lgtm

tuminoid (commented):
Just noting already existing issue on the line that is being changed in this review.

01_prepare_host.sh
lentzi90: Good point, I agree! This would be nice to clean up throughout the repository. Maybe we can do it as a follow up after we have the CI working again

Rozzii: It is shell, everything is a string.

Rozzii: But I can understand if you like better the the simplified comparison .

lentzi90: @Rozzii isn't there a difference between `if "${VARIABLE}"` and `if [[ "${VARIABLE}" == "true" ]]`? I think the first will actually try to execute VARIABLE and check the exit code, whereas the second will do a string comparison.

Rozzii: yes, I can see why it can be a security issue

tuminoid: @Rozzii If "${VARIABLE}" definitely executes the content of variable. If [ ... does not, as [ is an actual binary in /usr/bin, so the "${VARIABLE}" = true are all just parameters to that command. (disclaimer: [ is built-in for bash, but sh actually calls the [ binary)

Rozzii: I have actually missread your comments I thought the original was [[ "${VARIABLE}" == "true" ]] and we are changing it to ${IPA_DOWNLOAD_ENABLED}  but you are doing it the other way around.

So no complaint from me .

Rozzii: > yes, I can see why it can be a security issue

This is why I have answered with this comment I thought we are changing with the [[ -less

PR #1102, opened 2022-11-10
Changes needed to run metal3-dev-env in the Ironic CI

dtantsur:
See commit messages for details.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -13,7 +13,7 @@ fi
 
 if [[ $OS == ubuntu ]]; then
   sudo apt-get update
-  sudo apt -y install python3-pip jq curl wget
+  sudo apt -y install python3-pip jq curl wget pkg-config
 
   # Set update-alternatives to python3
   if [[ ${DISTRO} == "ubuntu18" ]]; then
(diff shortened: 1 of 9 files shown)


01_prepare_host.sh
```diff
@@ -13,7 +13,7 @@ fi
 
 if [[ $OS == ubuntu ]]; then
   sudo apt-get update
-  sudo apt -y install python3-pip jq curl wget
+  sudo apt -y install python3-pip jq curl wget pkg-config
 
   # Set update-alternatives to python3
   if [[ ${DISTRO} == "ubuntu18" ]]; then
```
elfosardo, line 19: for ubuntu jammy we need to reload the correct kvm module based on brand (amd or intel), not just kvm_intel
I guess we could check cpuinfo for the virtualization feature, vmx on intel, svm for amd, and decide based on that

dtantsur: I agree, but we're also in a nested VM situation, and I think at least some VMs don't have hardware-assisted nested virtualization...

elfosardo: right, then we need to remove the reload in case we don't have kvm at all

elfosardo:
probably need to rebase this as we changed k8s version

01_prepare_host.sh
elfosardo: this should avoid issues with nested virtualization or in general when we don't have kvm  https://github.com/dtantsur/metal3-dev-env/pull/1

elfosardo:
I think we can close this, the changes merged already in https://github.com/metal3-io/metal3-dev-env/pull/1160

PR #1105, opened 2022-11-16
🌱 Pin CAPM3 to v1.3.0-beta.0 pre-release for main jobs

furkatgofurov7:


lib/releases.sh
```diff
@@ -43,9 +43,10 @@ else
   export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.2.")}"
 fi
 
-# Remove the following line when CAPI release v1.3.0 is available
+# Remove the following lines when CAPI/CAPM3 v1.3.0 releases are available
 if [ "${CAPM3RELEASEBRANCH}" == "main" ]; then
  export CAPIRELEASE="v1.3.0-beta.1"
+ export CAPM3RELEASE="v1.3.0-beta.0"
```
mboukhalfa, line 49: cannot we force this workaround from capm3 so it can be easier to test full changes when introducing and removing this changes 
here https://github.com/metal3-io/cluster-api-provider-metal3/blob/7fcfe8a3beef3ddeb096366e15a8779936c2be50/scripts/ci-e2e.sh#L23

furkatgofurov7: we can, but we want this release to be used not only in capm3 & e2e tests, but with whole CI tests in general.

lentzi90 (commented):
/lgtm

lib/releases.sh
```diff
@@ -43,9 +43,10 @@ else
   export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.2.")}"
 fi
 
-# Remove the following line when CAPI release v1.3.0 is available
+# Remove the following lines when CAPI/CAPM3 v1.3.0 releases are available
 if [ "${CAPM3RELEASEBRANCH}" == "main" ]; then
  export CAPIRELEASE="v1.3.0-beta.1"
```
kashifest, line 48: we should probably use https://github.com/kubernetes-sigs/cluster-api/releases/tag/v1.3.0-rc.0 here now

kashifest (approved):

PR #1106, opened 2022-11-17
Use CAPI/CAPM3 pre-releases in main tests

furkatgofurov7:
Since we are using pre-releases of CAPM3/CAPI for some time now, it is better to fetch automatically regardless of how many pre-release will come out in the coming weeks. We can switch back get_latest_release logic to fetch the latest once normal vx.y.0 is out

### DIFF
--- a/lib/releases.sh
+++ b/lib/releases.sh
@@ -7,8 +7,8 @@ function get_latest_release() {
   else
     release="$(curl -H "Authorization: token ${GITHUB_TOKEN}" -sL "${1}")" || ( set -x && exit 1 )
   fi
-  # This gets the latest release as vx.y.z , ignoring any version with a suffix starting with - , for example -rc0
-  release_tags_unsorted="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\")) | select(contains(\"-\")==false)]" \
+  # This gets the latest release as vx.y.z or vx.y.z-rc.0, including any version with a suffix starting with - , for example -rc.0
+  release_tags_unsorted="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\"))]" \
     | cut -sf2 -d\"  | tr '.' ' ' )"
   if [[ $OS == ubuntu ]]; then
     release_tags_sorted="$(echo "$release_tags_unsorted" | sort -n +1 +2 )"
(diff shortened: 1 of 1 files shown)


lib/releases.sh
```diff
@@ -38,15 +38,12 @@ if [ "${CAPM3RELEASEBRANCH}" == "release-0.5" ] || [ "${CAPM3_VERSION}" == "v1al
...
-else
+elif [ "${CAPM3RELEASEBRANCH}" == "release-1.2" ]; then
   export CAPM3RELEASE="${CAPM3RELEASE:-$(get_latest_release "${CAPM3RELEASEPATH}" "v1.2.")}"
   export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.2.")}"
-fi
-
-# Remove the following lines when CAPI/CAPM3 v1.3.0 releases are available
-if [ "${CAPM3RELEASEBRANCH}" == "main" ]; then
- export CAPIRELEASE="v1.3.0-beta.1"
- export CAPM3RELEASE="v1.3.0-beta.0"
```
kashifest, line 49: I would rather just keep this and not change anything else since this is more visible in the code what we are using.

furkatgofurov7: I did not want to change these each time a new pre-release is out from CAPM3/CAPI (rc.0, rc.1 from CAPM3 & rc.1 from CAPI) and fetch it automatically.

kashifest: I prefer this way rather. But ok for me if you dont want it this way.
/approve

furkatgofurov7: okay, I can fix it as before too and have to change some more things here.

/hold

kashifest: No leave it as you have done it in this PR now. I see your point since for the next pre release we have to come again and change it.

furkatgofurov7: Sure, left it as is

furkatgofurov7: /hold cancel

lib/releases.sh
```diff
@@ -8,7 +8,7 @@ function get_latest_release() {
     release="$(curl -H "Authorization: token ${GITHUB_TOKEN}" -sL "${1}")" || ( set -x && exit 1 )
   fi
   # This gets the latest release as vx.y.z , ignoring any version with a suffix starting with - , for example -rc0
-  release_tags_unsorted="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\")) | select(contains(\"-\")==false)]" \
+  release_tags_unsorted="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-""}\"))]" \
```
lentzi90, line 11: Could you change the comment above so it matches the code or add some TODO for putting back the old behavior?

furkatgofurov7: Good point, fixed now

furkatgofurov7:
**NOTE:** Feature upgrade tests will be tackled separately and there is a tracking [issue](https://github.com/metal3-io/metal3-dev-env/issues/1107), this PR just deals with fetching pre-releases properly

lentzi90 (commented):
/lgtm

PR #1109, opened 2022-11-22
Use SQlite for Ironic

lentzi90:
This explicitly configures Ironic to use SQlite instead of Mariadb. We are already using sqlite since it is [the default](https://github.com/metal3-io/ironic-image/blob/main/scripts/runironic#L5) so this is not a change in behavior, it just makes it explicit. With this change it is also possible to set IRONIC_USE_MARIADB=true in the dev-env and actually run with mariadb.  However, there is no kustomization yet for running Ironic without Mariadb, so we will always have a Mariadb container no matter what this variable is set to.

The long term plan for this is to
1. Make the ironic kustomization more flexible: https://github.com/metal3-io/baremetal-operator/pull/1172
2. Add separate sqlite and mariadb flavors
3. Adapt metal3-dev-env to make use of these flavors so that we can truly run without the mariadb container.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -121,6 +121,10 @@ function update_images(){
 function launch_ironic() {
   pushd "${BMOPATH}"
 
+    # TODO(lentzi90): When the Ironic kustomizations support running without mariadb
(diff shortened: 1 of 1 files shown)


smoshiur1237 (approved):
/lgtm

furkatgofurov7 (approved):

PR #1113, opened 2022-11-28
Check v1a5 upgrade tests with higher k8s version

kashifest:


kashifest:
This is not needed anymore.
/close

PR #1115, opened 2022-11-29
Use ansible 7.0.0 only for Ubuntu 22.x

kashifest:
Recent Ansible uplift to 7.0.0 has broken ubuntu 18.x and ubuntu 20.x builds because the default python version in those distros are older than 3.10.x python version which is a requirement for ansible 7.0.0

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -308,7 +308,13 @@ fi
 export KIND_NODE_IMAGE=${KIND_NODE_IMAGE:-"${DOCKER_HUB_PROXY}/kindest/node:${KIND_NODE_IMAGE_VERSION}"}
 
 # Ansible version
-export ANSIBLE_VERSION=${ANSIBLE_VERSION:-"7.0.0"}
+# Older ubuntu version do no support 7.0.0 because of older python versions
+# Ansible 7.0.0 requires python 3.10+
+if [ "${DISTRO}" == "ubuntu22" ]; then
+    export ANSIBLE_VERSION=${ANSIBLE_VERSION:-"7.0.0"}
+else 
+    export ANSIBLE_VERSION=${ANSIBLE_VERSION:-"6.6.0"}
+fi
 
 # Test and verification related variables
 SKIP_RETRIES="${SKIP_RETRIES:-false}"


smoshiur1237 (commented):
/lgtm

lib/common.sh
```diff
@@ -308,7 +308,13 @@ fi
 export KIND_NODE_IMAGE=${KIND_NODE_IMAGE:-"${DOCKER_HUB_PROXY}/kindest/node:${KIND_NODE_IMAGE_VERSION}"}
 
 # Ansible version
-export ANSIBLE_VERSION=${ANSIBLE_VERSION:-"7.0.0"}
+# Older ubuntu version do no support 7.0.0 because of older python versions
+# Ansible 7.0.0 requires python 3.10+
+if [[ ${DISTRO} == "ubuntu22" ]]; then
```
tuminoid, line 313: [[ is unnecessary here as string comparison is basic sh functionality.

[ "${DISTRO}" = "ubuntu22" ] is more portable.

lib/common.sh
```diff
@@ -308,7 +308,13 @@ fi
...
+    export ANSIBLE_VERSION=${ANSIBLE_VERSION:-"7.0.0"}
+else 
+    export ANSIBLE_VERSION=${ANSIBLE_VERSION:-"6.5.0"}
```
tuminoid, line 316: 6.6.0 seems to be latest 6.x, could we use that?

furkatgofurov7 (approved):
Looks good, and python versioning based on distro is decided here: https://github.com/metal3-io/metal3-dev-env/blob/main/01_prepare_host.sh#L21-L22. I wonder why we should stick to it though, because you should be able to [run/install](https://computingforgeeks.com/how-to-install-python-on-ubuntu-linux-system/) python 3.10.X in both ubuntu18/20 as well

lib/common.sh
smoshiur1237: There are many instances of double [[ and single [ in this file for string comparison. does it make any difference? or we can keep them as it is.

furkatgofurov7:
/hold

to avoid merging, just saw @tuminoid comment, feel free to unhold when ready

lib/common.sh
Sunnatillo: previous one we had 6.5.0, and we updated to 7.0.0.  We normally test each version of ansible, so sticking to  6.5.0 version is preferred, since 6.6.0 is not tested.

lib/common.sh
kashifest: Shellcheck fails if  I do that @tuminoid , keeping it as it is.

lib/common.sh
kashifest: Agree with @Sunnatillo

lib/common.sh
kashifest: AH no sorry, shellcheck was saying to double quote

kashifest: Changed.

kashifest:
@smoshiur1237 please re-lgtm

lib/common.sh
kashifest: As @smoshiur1237 mentioned, the convention was so far to use double brackets most of the places in this file and thoughout the repo, we need to start changing it as some point otherwise mixed conventions would create confusion.

lib/common.sh
```diff
@@ -308,7 +308,13 @@ fi
 export KIND_NODE_IMAGE=${KIND_NODE_IMAGE:-"${DOCKER_HUB_PROXY}/kindest/node:${KIND_NODE_IMAGE_VERSION}"}
 
 # Ansible version
-export ANSIBLE_VERSION=${ANSIBLE_VERSION:-"7.0.0"}
+# Older ubuntu version do no support 7.0.0 because of older python versions
+# Ansible 7.0.0 requires python 3.10+
+if [ "${DISTRO}" == "ubuntu22" ]; then
+    export ANSIBLE_VERSION=${ANSIBLE_VERSION:-"7.0.0"}
+else 
+    export ANSIBLE_VERSION=${ANSIBLE_VERSION:-"6.5.0"}
```
elfosardo, line 316: let's use 6.6.0 ?

kashifest: Ok, seems this is more desirable option as @tuminoid also noted , I will push a patch here.

kashifest:
> Looks good, and python versioning based on distro is decided here: https://github.com/metal3-io/metal3-dev-env/blob/main/01_prepare_host.sh#L21-L22. I wonder why we should stick to it though, because you should be able to [run/install](https://computingforgeeks.com/how-to-install-python-on-ubuntu-linux-system/) python 3.10.X in both ubuntu18/20 as well

Good point. I will test it separately

smoshiur1237 (commented):
/lgtm

kashifest:
/test-ubuntu-integration-main
Not waiting to BML jobs to pass since it is not working properly with dev-env, however @adilGhaffarDev has tested it in project-infra and the BML job seem to be happy with ansible version https://jenkins.nordix.org/job/metal3_project_infra_bml_integration_tests_centos/212/console (The failure in deprovisioning is unrelated)

tuminoid:
lgtm

PR #1116, opened 2022-11-30
Fix BMOBRANCH var while kustomization

kashifest:
We always checkout using BMOBRANCH no matter if its a release tag or main branch, so this comparison has to be compared to main branch as well.


lentzi90 (approved):

furkatgofurov7 (approved):
/approve

PR #1118, opened 2022-12-02
WIP: lib: change release fetching logic

tuminoid:
Remove custom sorting, and rely on the order of the releases in Github.

Good news is that it will always pick the latest matching release, and its way simpler. This is OK as long as nobody retags v1.2.0-pre0 again after tagging v1.2.0 release.

Bad news is that selecting only official releases is not possible, while selecting prereleases or betas is when using specific enough prefix, like v1.3.0-pre


kashifest:
I really like this change, thanks a lot @tuminoid , lets do some tests to see if its working for older branches as well
/test-ubuntu-integration-main 
/test-ubuntu-integration-release-1-1
/test-ubuntu-integration-release-0-5

kashifest (approved):

Sunnatillo (commented):
/lgtm

PR #1126, opened 2022-12-14
bind registry to EXTERNAL_SUBNET_V4_HOST

tuminoid:
Registry should be binding to an IP by default. In case dev-env is ran on a machine that has internet connectivity, binding to all interfaces exposes unauthenticated registry to anyone. This would allow anyone to replace images with arbitrary code, or perform DOS by filling the machine's disk, etc.

Here we replace binding "$REGISTRY_PORT":5000 with $REGISTRY:5000 as $REGISTRY is defined as $EXTERNAL_SUBNET_V4_HOST:$REGISTRY_PORT and by default that is 192.168.11.1:5000

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -196,7 +196,7 @@ if [[ "$reg_state" == "exited" ]]; then
   sudo "${CONTAINER_RUNTIME}" start registry
 elif [[ "$reg_state" != "running" ]]; then
   sudo "${CONTAINER_RUNTIME}" rm registry -f || true
-  sudo "${CONTAINER_RUNTIME}" run -d -p "${REGISTRY_PORT}":5000 --name registry "$DOCKER_REGISTRY_IMAGE"
+  sudo "${CONTAINER_RUNTIME}" run -d -p "${REGISTRY}":5000 --name registry "$DOCKER_REGISTRY_IMAGE"
 fi
 sleep 5
 


02_configure_host.sh
```diff
@@ -196,7 +196,7 @@ if [[ "$reg_state" == "exited" ]]; then
   sudo "${CONTAINER_RUNTIME}" start registry
 elif [[ "$reg_state" != "running" ]]; then
   sudo "${CONTAINER_RUNTIME}" rm registry -f || true
-  sudo "${CONTAINER_RUNTIME}" run -d -p "${REGISTRY_PORT}":5000 --name registry "$DOCKER_REGISTRY_IMAGE"
+  sudo "${CONTAINER_RUNTIME}" run -d -p "${REGISTRY}":5000 --name registry "$DOCKER_REGISTRY_IMAGE"
```
lentzi90, line 199: I think we should drop the port, right?
```suggestion
  sudo "${CONTAINER_RUNTIME}" run -d -p "${REGISTRY}" --name registry "$DOCKER_REGISTRY_IMAGE"
```

tuminoid: No, the syntax is `IP:LOCAL PORT:CONTAINER PORT`.

lentzi90: Thanks for explaining!

lentzi90 (commented):
/lgtm

tuminoid:
/hold
until enough tests in various environments are done

Sunnatillo (approved):
/lgtm

tuminoid:
Thanks @Sunnatillo.

/hold cancel

PR #1127, opened 2022-12-19
Support deploying Ironic with/without MariaDB

lentzi90:
Note: We still set the mariadb container image to use even though we don't use mariadb. Not sure if it is worth it to add a check there also.
I removed mariadb from the containers we bump in the upgrade test though. This is because it would make the test more complex and upgrading a database seems a bit outside of scope. We will not anyway use mariadb in the normal tests.

Related:
- https://github.com/metal3-io/baremetal-operator/pull/1196

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -14,6 +14,19 @@ export IRONIC_HOST="${CLUSTER_URL_HOST}"
 export IRONIC_HOST_IP="${CLUSTER_PROVISIONING_IP}"
 export REPO_IMAGE_PREFIX="quay.io"
 
+declare -a BMO_IRONIC_ARGS
+# -k is for keepalived
+BMO_IRONIC_ARGS=(-k)
+if [ "${IRONIC_TLS_SETUP}" == "true" ]; then
+  BMO_IRONIC_ARGS+=("-t")
+fi
+if [ "${IRONIC_BASIC_AUTH}" == "false" ]; then
+  BMO_IRONIC_ARGS+=("-n")
+fi
+if [ "${IRONIC_USE_MARIADB:-false}" == "true" ]; then
+  BMO_IRONIC_ARGS+=("-m")
+fi
+
 sudo mkdir -p "${IRONIC_DATA_DIR}"
 sudo chown -R "${USER}:${USER}" "${IRONIC_DATA_DIR}"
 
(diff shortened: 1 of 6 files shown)


lentzi90:
Adding hold so we don't merge it before removing the test commits.
/hold

lentzi90:
Successful tests with mariadb enabled:
- https://jenkins.nordix.org/job/metal3_metal3_dev_env_main_integration_test_centos/45/
- https://jenkins.nordix.org/job/metal3_metal3_dev_env_feature_tests_ubuntu/126/
- https://jenkins.nordix.org/job/metal3_metal3_dev_env_main_integration_test_ubuntu/202/
- https://jenkins.nordix.org/job/metal3_metal3_dev_env_feature_tests_upgrade_ubuntu/95/

lentzi90:
Running tests without mariadb enabled now.
/test-features-ubuntu

lentzi90:
Test commit removed! I think this should work fine also without the BMO changes, so this PR is safe to merge. Checking with tests:
/test-ubuntu-integration-main

kashifest (approved):

03_launch_mgmt_cluster.sh
```diff
@@ -14,6 +14,18 @@ export IRONIC_HOST="${CLUSTER_URL_HOST}"
 export IRONIC_HOST_IP="${CLUSTER_PROVISIONING_IP}"
 export REPO_IMAGE_PREFIX="quay.io"
 
+# -k is for keepalived
+BMO_IRONIC_ARGS=(-k)
+if [[ "${IRONIC_TLS_SETUP}" == "true" ]]; then
```
tuminoid, line 19: Nit: [[ is not necessary here, and convention in this file is single [.

(== is also not necessary, but is a convention in the file, so ignoring it)

03_launch_mgmt_cluster.sh
```diff
@@ -14,6 +14,18 @@ export IRONIC_HOST="${CLUSTER_URL_HOST}"
 export IRONIC_HOST_IP="${CLUSTER_PROVISIONING_IP}"
 export REPO_IMAGE_PREFIX="quay.io"
 
+# -k is for keepalived
+BMO_IRONIC_ARGS=(-k)
```
tuminoid, line 18: It would be nice if you'd explicitly declare this as an array (declare -a) before use

tuminoid (commented):
nits

03_launch_mgmt_cluster.sh
lentzi90: Thanks, fixed

03_launch_mgmt_cluster.sh
lentzi90: Thanks, fixed

PR #1129, opened 2022-12-20
Fix BMO Local image building var naming

kashifest:


lentzi90 (commented):
/lgtm
Explanation for why this is needed: We used a different variable name [here](https://github.com/metal3-io/metal3-dev-env/blob/main/03_launch_mgmt_cluster.sh#L40).

lib/common.sh
```diff
@@ -195,7 +195,7 @@ if [ "${BUILD_IPAM_LOCALLY}" == "true" ]; then
   export IPAM_LOCAL_IMAGE="${IPAMPATH}"
 fi
 if [ "${BUILD_BMO_LOCALLY}" == "true" ]; then
-  export BMO_LOCAL_IMAGE="${BMOPATH}"
+  export BAREMETAL_OPERATOR_LOCAL_IMAGE="${BMOPATH}"
```
furkatgofurov7, line 198: Where is this `BMO_LOCAL_IMAGE` var is coming from

PR #1130, opened 2022-12-21
Bump k8s to v1.26.0

kashifest:


lentzi90 (commented):
/lgtm

tests/feature_tests/feature_test_vars.sh
```diff
@@ -2,5 +2,5 @@
 
 export CONTROL_PLANE_MACHINE_COUNT="3"
 export WORKER_MACHINE_COUNT="1"
-export FROM_K8S_VERSION="v1.24.1"
+export FROM_K8S_VERSION="v1.25.2"
```
furkatgofurov7, line 5: this was left out, rerunning the tests again

tests/feature_tests/upgrade/upgrade_vars.sh
```diff
@@ -3,13 +3,13 @@ set -eux
 
 # Folder created for specific capi release when running
 # ${CLUSTER_API_REPO}/cmd/clusterctl/hack/create-local-repository.py
-export CAPIRELEASE_HARDCODED="v1.2.99"
+export CAPIRELEASE_HARDCODED="v1.3.99"
 
 export CAPI_VERSION="${CAPI_VERSION:-v1alpha4}"
-export CAPI_REL_TO_VERSION="v1.2.2"
+export CAPI_REL_TO_VERSION="v1.3.1"
```
furkatgofurov7, line 9: upgrade tests were still using old CAPI/CAPM3 release versions, so had to fix this too

tests/roles/run_tests/tasks/upgrade.yml
```diff
@@ -328,7 +328,8 @@
   - name: Upgrade k8s version and boot-image
     include_tasks: "upgrade_k8s.yml"
     loop:
-      - "v1.24.1"
+      - "v1.24.9"
+      - "v1.25.2"
```
furkatgofurov7, line 332: with this, we are now testing 3 folded upgrades (v1.23>v1.24>v1.25>v1.26) with the same test, maybe we need to think soon about how we would handle v1a5>v1b1 upgrade tests

tests/roles/run_tests/tasks/upgrade.yml
```diff
@@ -328,7 +328,8 @@
...
       - "{{ UPGRADED_K8S_VERSION }}"
```
furkatgofurov7, line 333: with this, we are now testing 3 folded upgrades (v1.23>v1.24>v1.25>v1.26) with the same test, maybe we need to think soon about how we would handle v1a5>v1b1 upgrade tests

lentzi90 (commented):
/lgtm
But we need to get rid of the k8s upgrade loop after this IMO. The tests will take too much time like this.

kashifest:
> /lgtm But we need to get rid of the k8s upgrade loop after this IMO. The tests will take too much time like this.

Yes, there is two way to handle this. Modify upgrade tests here and replicate what e2e is doing i.e. separate out provider component upgrade and k8s version upgrade or completely abandon the ansible test. I am more in favor of two. The reason to keep it going is to wait for e2e upgrade to include all the use cases we have in ansbile currently and then abandon it afterwards. Until then, I plan to keep this test green with least effort.

furkatgofurov7 (approved):

PR #1131, opened 2022-12-30
Configure vlan interface in centos when EXTERNAL_VLAN_ID is set 

Sunnatillo:
Configure vlan interface via kubeadm config when EXTERNAL_VLAN_ID is set.


Sunnatillo:
/hold
until we test it in bml

furkatgofurov7 (approved):

PR #1133, opened 2023-01-09
.github: secure github actions

tuminoid:
Set top-level permissions to read.

Pin actions by a sha.

### DIFF
--- a/.github/workflows/kubesec.yml
+++ b/.github/workflows/kubesec.yml
@@ -5,6 +5,9 @@ on:
   schedule:
     - cron: '30 7 * * 2'
 
+permissions:
+  contents: read
+
 jobs:
   setup:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


.github/workflows/kubesec.yml
```diff
@@ -13,7 +16,7 @@ jobs:
       contents: read
```
kashifest, line 16: Curious to know why do we need top level permission set if we have set it here in more granular steps.

kashifest: So why not `permissions: {} `at the top level and let the jobs decide what it needs?

tuminoid: Avoids repetition. But it could be done like that too.

kashifest: I am fine with the current change as well. Was curious to know the impact of different configurations.
/approve

tuminoid:
> You can actually change the default behavior for all workflows in settings only. This makes sure none of the workflows would have write access by default. This can be done in org level as well as individual repo level.

They can be set, but then they're decoupled from actual actions. The further away you set them (repo or org level), the more likely it is that there is a gap. Setting them in the workflow files keeps them up to date with the jobs.

Thanks for the discussion! It is good that changes are not rubber-stamped, but well scrutinized.

lentzi90 (commented):
/lgtm

PR #1138, opened 2023-01-17
Control-plane and worker deprovision scripts are removed.

adilGhaffarDev:
This is a follow-up of this PR: https://github.com/metal3-io/metal3-dev-env/pull/1082
Separate de-provisioning of worker and controlplane was removed in the above PR and the shell scripts were not removed. This PR is removing those orphaned shell scripts.

### DIFF
--- a/tests/scripts/deprovision/controlplane.sh
+++ b/tests/scripts/deprovision/controlplane.sh
@@ -1,7 +1,0 @@
-#!/bin/bash
-
-METAL3_DIR="$(dirname "$(readlink -f "${0}")")/../../.."
-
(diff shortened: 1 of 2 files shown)


Rozzii (approved):

Sunnatillo (approved):

adilGhaffarDev:
closing this PR, discussed with @furkatgofurov7 he suggested keeping deprovision scripts and adding the ansible blocks back, so if someone wants to do the deprovisioning separately they can just use these scripts. I will not include them in CI, CI will use the same logic of cleaning the whole cluster.

PR #1141, opened 2023-01-19
install kubectl completion file by default

tuminoid:
Install bash-completion package (ubuntu has it by default, centos doesn't), and have kubectl generate their bash completion file upon installation. This way developer can use kubectl autocompletion without any extra steps, making life a bit more bearable.


lentzi90 (commented):
Thank you!
/lgtm

furkatgofurov7 (approved):

PR #1164, opened 2023-03-03
fix dev-env installation on ubuntu 18.04/centos 8

tuminoid:
Ubuntu 18.04/Centos8 cannot install Ansible newer than 4.x as they have
only Python 3.6.

Ubuntu 18.04 needs python3-bcrypt as well. It must be installed as
system package as bcrypt bugs out pip's older than 19.3, but pip
upgrade breaks other module installations.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -151,7 +151,7 @@ elif [ "${CAPM3RELEASEBRANCH}" == "release-1.2" ]; then
   export IPAMBRANCH="${IPAMBRANCH:-release-1.2}"
 elif [ "${CAPM3RELEASEBRANCH}" == "release-1.3" ]; then
   export CAPM3BRANCH="${CAPM3BRANCH:-release-1.3}"
-  export IPAMBRANCH="${IPAMBRANCH:-release-1.3}"  
+  export IPAMBRANCH="${IPAMBRANCH:-release-1.3}"
 else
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
/approve

tuminoid:
Yeah agree with @Rozzii but let's align next week on dropping support for Ubuntu 18.04 (which is still in support), Centos 8 (EOL) and maybe also RHEL 8 (supported but shares base with Centos 8, which we'd drop) and maybe even RHEL 9 (a paid OS is not exactly environment where people will run dev env in OSS community anyway).

@elfosardo @dtantsur we could use your opinion on dropping RHEL (8, possibly 9) from dev-env, we'll bring it to community meeting if needed?

tuminoid:
PR to remove unsupported OS instead: https://github.com/metal3-io/metal3-dev-env/pull/1167 to be discussed in Community meeting next Wed.

dtantsur:
@tuminoid our dev-scripts for OCP is based on metal3-dev-env. If you remove RHEL 9, you'll break us. We'd appreciate if you keep it and are available to help with maintenance.

Actually, my personal environment is still based on Stream 8, but I understand if you don't want to maintain it.

tuminoid:
We keep support, and hence need this PR. https://github.com/metal3-io/cluster-api-provider-metal3/pull/882#issuecomment-1460251628

/unhold

Issue #1162, opened 2023-03-01
Add pagination when requesting github API


mboukhalfa:
It seems that by default github api return only limited list of resources and in our script e.g https://github.com/Nordix/cluster-api-provider-metal3/blob/0c80fcd789608e3299aaba286fc01b1596332516/hack/gen_tilt_settings.sh#L20 we are not exploring the rest of resources using pagination which might cause issue if the target resource is not in the first page.

We should improve these script to explore the next pages if the target resource is not in the first page

PR #1168, opened 2023-03-06
Use pagination when requesting github API

mboukhalfa:
Fix https://github.com/metal3-io/metal3-dev-env/issues/1162

### DIFF
--- a/lib/releases.sh
+++ b/lib/releases.sh
@@ -1,30 +1,51 @@
 #!/bin/bash
 
 function get_latest_release() {
-  set +x
+  # get last page
   if [ -z "${GITHUB_TOKEN:-}" ]; then
-    release="$(curl -sL "${1}")" || ( set -x && exit 1 )
+    last_page="$(curl -s -I "${1}" | grep '^link:' | sed -e 's/^link:.*page=//g' -e 's/>.*$//g')"
   else
-    release="$(curl -H "Authorization: token ${GITHUB_TOKEN}" -sL "${1}")" || ( set -x && exit 1 )
+    last_page="$(curl -s -I "${1}" -H "Authorization: token $GITHUB_TOKEN" | grep '^link:' | sed -e 's/^link:.*page=//g' -e 's/>.*$//g')"
   fi
-  # This gets the latest release as vx.y.z or vx.y.z-rc.0, including any version with a suffix starting with - , for example -rc.0
-  # The order is exactly as released in Github.
-  # Downside is that selecting official releases only isn't possible, while pre-release
-  # selection is possible given specific enough prefix, like v1.3.0-pre
-  release_tag="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-}\"))] | .[0]")"
+  # default last page to 1
+  last_page="${last_page:-1}"
 
-  if [[ "$release_tag" == "null" ]]; then
-    set -x
+  set +x
+
+  for current_page in $(seq 1 "$last_page"); do
+
+    url="${1}?page=${current_page}"
+
+    if [ -z "${GITHUB_TOKEN:-}" ]; then
+      release="$(curl -sL "${url}")" || { set -x && exit 1; }
+    else
+      release="$(curl -H "Authorization: token ${GITHUB_TOKEN}" -sL "${url}")" || { set -x && exit 1; }
+    fi
+
+    # This gets the latest release as vx.y.z or vx.y.z-rc.0, including any version with a suffix starting with - , for example -rc.0
+    # The order is exactly as released in Github.
+    # Downside is that selecting official releases only isn't possible, while pre-release
+    # selection is possible given specific enough prefix, like v1.3.0-pre
+    release_tag="$(echo "${release}" | jq -r "[.[].tag_name | select( startswith(\"${2:-}\"))] | .[0]")"
+    # if release tag found
+    if [[ "${release_tag}" != "null" ]]; then
+      break
+    fi
+
+  done
+
+  set -x
+  # if not found
+  if [[ "${release_tag}" == "null" ]]; then
     exit 1
   fi
-  set -x
   # shellcheck disable=SC2005
-  echo "$release_tag"
+  echo "${release_tag}"
 }
 
 # CAPM3, CAPI and BMO release path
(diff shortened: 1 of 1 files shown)


lib/releases.sh
```diff
@@ -2,29 +2,49 @@
 
 function get_latest_release() {
   set +x
+  # get last page
   if [ -z "${GITHUB_TOKEN:-}" ]; then
-    release="$(curl -sL "${1}")" || ( set -x && exit 1 )
+    last_page=$(curl -s -I "${1}" -H "Authorization: token $GITHUB_TOKEN" | grep '^link:' | sed -e 's/^link:.*page=//g' -e 's/>.*$//g') || (set -x && exit 1)
```
tuminoid, line 7: `(set -x && exit 1)` creates a subshell, the parenthesis does that, so the `exit 1` exits only that subshell, and not the script. If you want to perform two actions, use curly braces, `{set -x && exit 1}`.

Also, you want to assign the subshell output of `$( ... )` with quotes, so do `last_page="$( ... )"`

lib/releases.sh
```diff
@@ -2,29 +2,49 @@
...
   else
-    release="$(curl -H "Authorization: token ${GITHUB_TOKEN}" -sL "${1}")" || ( set -x && exit 1 )
+    last_page=$(curl -s -I "${1}" | grep '^link:' | sed -e 's/^link:.*page=//g' -e 's/>.*$//g') || (set -x && exit 1)
```
tuminoid, line 9: same here

lib/releases.sh
```diff
@@ -2,29 +2,49 @@
...
   fi
-  # This gets the latest release as vx.y.z or vx.y.z-rc.0, including any version with a suffix starting with - , for example -rc.0
-  # The order is exactly as released in Github.
-  # Downside is that selecting official releases only isn't possible, while pre-release
-  # selection is possible given specific enough prefix, like v1.3.0-pre
-  release_tag="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-}\"))] | .[0]")"
+  # default last page to 1
+  last_page=${last_page:-1}
```
tuminoid, line 12: Quotes do no harm here either.

lib/releases.sh
```diff
@@ -2,29 +2,49 @@
...
 
+  for current_page in $(seq 1 "$last_page"); do
```
tuminoid, line 14: Any time you want to use `seq` to create sequences in Bash for loops, use `{1..$last_page}` instead. This avoids various issues with subshells, output etc.

lib/releases.sh
```diff
@@ -2,29 +2,49 @@
...
+
+    url="${1}?page=${current_page}"
+
+    if [ -z "${GITHUB_TOKEN:-}" ]; then
+      release="$(curl -sL "${url}")" || (set -x && exit 1)
```
tuminoid, line 19: same as the first one

lib/releases.sh
```diff
@@ -2,29 +2,49 @@
...
+    else
+      release="$(curl -H "Authorization: token ${GITHUB_TOKEN}" -sL "${url}")" || (set -x && exit 1)
```
tuminoid, line 21: same as the first one

lib/releases.sh
```diff
@@ -2,29 +2,49 @@
...
+    fi
+
+    # This gets the latest release as vx.y.z or vx.y.z-rc.0, including any version with a suffix starting with - , for example -rc.0
+    # The order is exactly as released in Github.
+    # Downside is that selecting official releases only isn't possible, while pre-release
+    # selection is possible given specific enough prefix, like v1.3.0-pre
+    release_tag="$(echo "$release" | jq -r "[.[].tag_name | select( startswith(\"${2:-}\"))] | .[0]")"
```
tuminoid, line 28: Add curlys to `$release`

lib/releases.sh
```diff
@@ -2,29 +2,49 @@
...
+    # if release tag found
+    if [[ "$release_tag" != "null" ]]; then
```
tuminoid, line 30: same

lib/releases.sh
mboukhalfa: done

mboukhalfa: (warning) Bash doesn't support variables in brace range expansions !

tuminoid: Duh, my bad. See, Bash is hard :)

lib/releases.sh
tuminoid: `${set -x ...}` is definitely wrong, but what does it complain about `{ set -x ... }` ?

mboukhalfa: thanks! done

mboukhalfa:
/assign @furkatgofurov7
cc @lentzi90

lib/releases.sh
```diff
@@ -49,13 +70,13 @@ BMOBRANCH="${BMOBRANCH:-${BMORELEASE}}"
...
 
 if [[ "$CAPM3RELEASE" == "" ]]; then
-  command -v jq &> /dev/null && echo "Failed to fetch CAPM3 release from Github" && exit 1
+  command -v jq &>/dev/null && echo "Failed to fetch CAPM3 release from Github" && exit 1
 fi
 
 if [[ "$BMORELEASE" == "" ]]; then
-  command -v jq &> /dev/null && echo "Failed to fetch BMO release from Github" && exit 1
+  command -v jq &>/dev/null && echo "Failed to fetch BMO release from Github" && exit 1
 fi
```
lentzi90, line 82: Why remove the space here? I think it is easier to read with the space so we should add it back.

furkatgofurov7 (approved):
Checked locally, it works fine

/approve
/hold to address the comments from @lentzi90

lib/releases.sh
mboukhalfa: it was aut reformatted by editor linter what do we have coding style for that

lentzi90: @tuminoid any preference on this? I'm inclined to approve it as just a matter of preference if there are no objections.

tuminoid: That space is completely up to personal taste. I tend to have the space there for readability, but it doesn't really matter.

lentzi90: Alright I'm putting lgtm then :slightly_smiling_face:

PR #1173, opened 2023-03-10
Fetch target cluster logs before re-pivot

smoshiur1237:
This PR will help to fetch target cluster logs from pods and save them in tmp folder during integration and feature test.

### DIFF
--- a/tests/feature_tests/pivoting/Makefile
+++ b/tests/feature_tests/pivoting/Makefile
@@ -13,6 +13,9 @@ fetch_manifests:
 pivoting:
 	./pivot.sh
 
+fetch_target_logs:
+	./../../scripts/fetch_target_logs.sh
+
 repivoting:
 	./repivot.sh
 
--- a/tests/roles/run_tests/vars/main.yml
+++ b/tests/roles/run_tests/vars/main.yml
@@ -162,7 +162,6 @@ pivot_actions:
     - "pivoting"
     - "upgrading"
 repivot_actions:
-    - "ci_test_provision"
     - "repivoting"
     - "upgrading"
 cleanup_actions:
--- a/tests/scripts/fetch_target_logs.sh
+++ b/tests/scripts/fetch_target_logs.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -x
+
+KUBE_CONFIG="$(sudo find /tmp/ -type f -name "kubeconfig*")"
+DIR_NAME="/tmp/target_cluster_logs"
+NAMESPACES="$(kubectl --kubeconfig="${KUBE_CONFIG}" get namespace -o json | jq -r '.items[].metadata.name')"
+mkdir -p "${DIR_NAME}"
+for NAMESPACE in ${NAMESPACES}
+do
+  mkdir -p "${DIR_NAME}/${NAMESPACE}"
+# Fetch logs from target cluster according to the pods name
+  PODS="$(kubectl --kubeconfig="${KUBE_CONFIG}" get pods -n "${NAMESPACE}" -o json | jq -r '.items[].metadata.name')"
+  for POD in ${PODS}
+  do
+    mkdir -p "${DIR_NAME}/${NAMESPACE}/${POD}"
+    CONTAINERS="$(kubectl --kubeconfig="${KUBE_CONFIG}" get pods -n "${NAMESPACE}" "${POD}" -o json | jq -r '.spec.containers[].name')"
+    for CONTAINER in ${CONTAINERS}
+    do
+      LOG_DIR="${DIR_NAME}/${NAMESPACE}/${POD}/${CONTAINER}"
+      mkdir -p "${LOG_DIR}"
+      kubectl --kubeconfig="${KUBE_CONFIG}" logs -n "${NAMESPACE}" "${POD}" "${CONTAINER}" \
+        > "${LOG_DIR}/stdout.log" 2> "${LOG_DIR}/stderr.log"
+    done
+  done
+done
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ pivoting_test:
 	make -C ./tests/feature_tests/pivoting/
 
 repivoting_test:
+	make -C ./tests/feature_tests/pivoting/ fetch_target_logs
 	make -C ./tests/feature_tests/pivoting/ repivoting
 	make -C ./tests//feature_tests/pivoting/ deprovision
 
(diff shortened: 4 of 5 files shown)


tests/feature_tests/pivoting/Makefile
```diff
@@ -2,7 +2,7 @@ M3PATH := "$(dirname "$(readlink -f "${0}")")../../../"
 export CONTROL_PLANE_MACHINE_COUNT := 3
 export WORKER_MACHINE_COUNT := 1
 
-all: provision fetch_manifests pivoting
+all: provision fetch_manifests pivoting fetch_target_logs
```
smoshiur1237, line 5: This addition of fetch_target_logs will cover the log collection for feature tests.

tests/roles/run_tests/vars/main.yml
```diff
@@ -162,7 +162,7 @@ pivot_actions:
     - "pivoting"
     - "upgrading"
 repivot_actions:
-    - "ci_test_provision"
```
kashifest, line 165: trying to understand removal of this line, can you elaborate why this action is removed?

kashifest: and why not changing the previous ci_test_provision entries in other actions?

kashifest: this seems incorrect to me.

kashifest: Thanks for the explanation offline, it should work as I understand now, but my assumption was we are going to do it ansible way i.e. adding tasks at the beginning of move_back.yml. Anyways, if this works I am fine with this approach as well, as long as it doesnt break local as well as CI tests.

tests/feature_tests/pivoting/Makefile
```diff
@@ -13,6 +13,9 @@ fetch_manifests:
 pivoting:
 	./pivot.sh
 
+fetch_target_logs:
+	./../../scripts/fetch_target_logs.sh
```
tuminoid, line 17: It would be more readable and maintainable if path would not be relative, but from the repository root.

tests/scripts/fetch_target_logs.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
tuminoid, line 1: Use `#!/usr/bin/env bash` when not in container.

tests/scripts/fetch_target_logs.sh
```diff
@@ -0,0 +1,26 @@
...
+
+set -x
```
tuminoid, line 3: Why not `set -eux`? `-eu` is safeguarding against many hidden errors, failing early rather than late.

tests/scripts/fetch_target_logs.sh
```diff
@@ -0,0 +1,26 @@
...
+  PODS="$(kubectl --kubeconfig="${KUBE_CONFIG}" get pods -n "${NAMESPACE}" -o json | jq -r '.items[].metadata.name')"
+  for POD in ${PODS}
+  do
+    mkdir -p "${DIR_NAME}/${NAMESPACE}/${POD}"
+    CONTAINERS="$(kubectl --kubeconfig="${KUBE_CONFIG}" get pods -n "${NAMESPACE}" "${POD}" -o json | jq -r '.spec.containers[].name')"
+    for CONTAINER in ${CONTAINERS}
+    do
+      mkdir -p "${DIR_NAME}/${NAMESPACE}/${POD}/${CONTAINER}"
+      kubectl --kubeconfig="${KUBE_CONFIG}" logs -n "${NAMESPACE}" "${POD}" "${CONTAINER}" \
+      > "${DIR_NAME}/${NAMESPACE}/${POD}/${CONTAINER}/stdout.log"\
```
tuminoid, line 22: Add space before `\` for readability.

tests/scripts/fetch_target_logs.sh
```diff
@@ -0,0 +1,26 @@
...
+  mkdir -p "${DIR_NAME}/${NAMESPACE}"
+# Fetch logs from target cluster according to the pods name
+  PODS="$(kubectl --kubeconfig="${KUBE_CONFIG}" get pods -n "${NAMESPACE}" -o json | jq -r '.items[].metadata.name')"
+  for POD in ${PODS}
+  do
+    mkdir -p "${DIR_NAME}/${NAMESPACE}/${POD}"
+    CONTAINERS="$(kubectl --kubeconfig="${KUBE_CONFIG}" get pods -n "${NAMESPACE}" "${POD}" -o json | jq -r '.spec.containers[].name')"
+    for CONTAINER in ${CONTAINERS}
+    do
+      mkdir -p "${DIR_NAME}/${NAMESPACE}/${POD}/${CONTAINER}"
```
tuminoid, line 20: `${DIR_NAME}/${NAMESPACE}/${POD}/${CONTAINER}"` is repeated few times in this loop. It would make sense to assign to `LOG_DIR` or such and use that. Makes entire loop much shorter and more readable.

tuminoid (changes requested):

tests/scripts/fetch_target_logs.sh
smoshiur1237: I think we don't want to fail the CI run because of the log collection error. That is why we use only `-x`.

tests/feature_tests/pivoting/Makefile
tuminoid: Agreed to be fixed in another PR, following the convention in the script.

tests/scripts/fetch_target_logs.sh
tuminoid: Yes, true.

tuminoid (approved):

tests/roles/run_tests/vars/main.yml
```diff
@@ -162,7 +162,7 @@ pivot_actions:
...
+    - "ci_test_repivoting"
```
furkatgofurov7, line 165: this is confusing to me as well. You are introducing `ci_test_repivoting` here but just line below there is `repivoting` as well

smoshiur1237: Previously it was` ci_test_provision`, just to keep it similar I have added the `ci_test_repivoting`. It allows to run the fetch log from target cluster before re-pivot. It used only in integration test.

smoshiur1237: Thanks for the discussion. I am reusing the `repivoting` action.

Sunnatillo (commented):
/lgtm

smoshiur1237:
Log collection test with project-infra is working as expected. Here in the following test, the target logs are saved in `target_cluster_logs`.
https://jenkins.nordix.org/job/metal3_project_infra_main_integration_test_ubuntu/87/

furkatgofurov7 (approved):

PR #1169, opened 2023-03-07
Refactor dev-env variable names

Rozzii:
The goal of this commit is to make the variable names involved in
network configuration more explicit. The main goal of this commit is to
make the difference between the "provisioner" network and the "provisioning"
network more prononced.

In the future the dev-env will support at least 2 different
network topologies thus the provisioner (e.g. ironic) won't be necessarily available
in the provisioning network the same way as it is available now.

This commit is a forruner of future changes that will introduce additional network
organization options.

List of name changes:

 - CLUSTER_PROVISIONING_INTERFACE -> BARE_METAL_PROVISIONER_INTERFACE
 - PROVISIONING_IP -> BARE_METAL_PROVISIONER_IP
 - PROVISIONING_IPV6 -> BARE_METAL_PROVISIONER_SUBNET_IPV6_ONLY
 - PROVISIONING_NETWORK -> BARE_METAL_PROVISIONER_NETWORK
 - PROVISIONING_CIDR -> BARE_METAL_PROVISIONER_CIDR
 - PROVISIONING_NETMASK -> BARE_METAL_PROVISIONER_NETMASK
 - CLUSTER_PROVISIONING_IP -> CLUSTER_BM_PROVISIONER_IP
 - PROVISIONING_URL_HOST -> BARE_METAL_RPOVISIONER_URL_HOST
 - CLUSTER_URL_HOST -> CLUSTER_BARE_METAL_PROVISIONER_HOST
 - VIRSH_DHCP_V4_START -> EXTERNAL_DHCP_V4_START
 - VIRSH_DHCP_V4_END -> EXTERNAL_DHCP_V4_START
 - BAREMETALV4_POOL_RANGE_START -> IPAM_EXTERNALV4_POOL_RANGE_START
 - BAREMETALV4_POOL_RANGE_END -> IPAM_EXTERNALV4_POOL_RANGE_END
 - VIRSH_DHCP_V6_START ->  EXTERNAL_DHCP_V6_START
 - VIRSH_DHCP_V6_END -> EXTERNAL_DHCP_V6_END
 - BAREMETALV6_POOL_RANGE_START ->  IPAM_EXTERNALV6_POOL_RANGE_START
 - BAREMETALV6_POOL_RANGE_END -> IPAM_EXTERNALV6_POOL_RANGE_END
 - INITIAL_IRONICBRIDGE_IP -> INITIAL_BM_PROVISIONER_BRIDGE_IP
 - baremetal_network_cidr_v4 -> external_network_cidr_v4
[...]

PR #1174, opened 2023-03-13
Add optional fully NATed network configuration 

Rozzii:
In certain development scenarios the user might need to bootstrap
an environment where all the interfaces have NAT forwarding modes.

This commit also improves some kubeadm scripts to make the target node's bootstrap process more reliable.
The modifications related to interface setup and keepalived are added in this commit because
without the improvements the NATed topology seemed to be unstable.

This commit does the following:

NATed topology:
- Adds optional fully NATed network topology where both the external network.
  and the provisioning network are in separate subnets connected to the host network via NAT.
- IPs for both the external and the provisioning network are assigned by libvirt's dhcp during inspection.
- Introduces a new variable to make the provisioning subnet customizable.

Modifications in kubeadm scripts:
- Improves the initialization process of ironicendpoint bridge interface on the nodes.
- Introduces 60s time delay after the keepalived is started on the control plane nodes in order to avoid keepalived
  conflicting with other processes.
- Removes the custom keepalived monitoring script as it seems to be too agressive and restarts keepalived even when
  it is not needed + there should be no need for an external process to turn on/off keepalived if it is correctly
  deployed.
- Based on experience it seems that the time provided for calico to start up is not always engough so I
  moved the wait period from 3min to 10min and the frequency of the verification task from 20s to 10s.
- Implements the same 10min wait period for m3m verification as for calico verification.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -140,6 +140,9 @@ function launch_ironic() {
   pushd "${BMOPATH}"
 
     # Update Configmap parameters with correct urls
+    # Variable names inserted into the configmap might have different
+    # naming conventions than the dev-env e.g. PROVISIONING_IP and CIDR are
+    # called PROVISIONER_IP and CIDR in dev-env
     cat << EOF | sudo tee "${IRONIC_DATA_DIR}/ironic_bmo_configmap.env"
 HTTP_PORT=${HTTP_PORT}
 PROVISIONING_IP=${CLUSTER_BARE_METAL_PROVISIONER_IP}
--- a/ubuntu_bridge_network_configuration.sh
+++ b/ubuntu_bridge_network_configuration.sh
@@ -15,8 +15,10 @@ if [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
      # the IP address here.
      # Create a veth iterface peer.
      sudo ip link add ironicendpoint type veth peer name ironic-peer
-     # Create provisioning bridge.
-     sudo brctl addbr provisioning
+     # Create provisioning bridge, if the user allowed bridged provisioning network.
+     if [[ "${ENABLE_NATED_PROVISIONING_NETWORK:-false}" = "false" ]]; then
+         sudo brctl addbr provisioning
+     fi
      # sudo ifconfig provisioning 172.22.0.1 netmask 255.255.255.0 up
      # Use ip command. ifconfig commands are deprecated now.
      sudo ip link set provisioning up
--- a/vm-setup/roles/common/defaults/main.yml
+++ b/vm-setup/roles/common/defaults/main.yml
@@ -32,6 +32,13 @@ external_dhcp_v4_end: "{{ lookup('env', 'EXTERNAL_DHCP_V4_END')|default('', true
 external_dhcp_v6_start: "{{ lookup('env', 'EXTERNAL_DHCP_V6_START')|default('', true) }}"
 external_dhcp_v6_end: "{{ lookup('env', 'EXTERNAL_DHCP_V6_END')|default('', true) }}"
 
+provisioning_network_cidr_v4: "{{ lookup('env', 'PROVISIONING_SUBNET_V4')|default('', true) }}"
+provisioning_network_cidr_v6: "{{ lookup('env', 'PROVISIONING_SUBNET_V6')|default('', true) }}"
+provisioning_dhcp_v4_start: "{{ lookup('env', 'PROVISIONING_DHCP_V4_START')|default('', true) }}"
+provisioning_dhcp_v4_end: "{{ lookup('env', 'PROVISIONING_DHCP_V4_END')|default('', true) }}"
+provisioning_dhcp_v6_start: "{{ lookup('env', 'PROVISIONING_DHCP_V6_START')|default('', true) }}"
+provisioning_dhcp_v6_end: "{{ lookup('env', 'PROVISIONING_DHCP_V6_END')|default('', true) }}"
+
 # Set this to `false` if you don't want your vms
 # to have a VNC console available.
 enable_vnc_console: true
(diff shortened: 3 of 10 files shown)


Rozzii:
Waiting until these are merged:
https://github.com/metal3-io/cluster-api-provider-metal3/pull/885
https://github.com/metal3-io/metal3-dev-env/pull/1169

vm-setup/roles/common/defaults/main.yml
```diff
@@ -79,4 +106,11 @@ external_network:
           addr: "{% if external_network_cidr_v4|ipv4 != False %}127.0.0.1{% else %}::1{% endif %}"
       srvs: "{{dns_externalsrvs | default([])}}"
 
-networks: "{{ provisioning_network + external_network }}"
+# Provisioning network is bridged and external network is nated
+networks_mixed: "{{ provisioning_network_bridge + external_network }}"
+# Both networks are nated
+networks_nated: "{{ provisioning_network_nat + external_network }}"
+# Enable only nated networks
+networks_nat_only: "false"
```
lentzi90, line 114: Why not set it from env here directly?
```suggestion
networks_nat_only: "{{ lookup('env', 'ENABLE_NATED_PROVISIONING_NETWORK') | bool }}"
```
After that we could set `networks` to the correct value also so we don't have to do all the `set_fact` tasks

Rozzii: You are right, I will change it.

ubuntu_bridge_network_configuration.sh
```diff
@@ -15,8 +15,10 @@ if [ "$MANAGE_PRO_BRIDGE" == "y" ]; then
      # the IP address here.
      # Create a veth iterface peer.
      sudo ip link add ironicendpoint type veth peer name ironic-peer
-     # Create provisioning bridge.
-     sudo brctl addbr provisioning
+     # Create provisioning bridge, if the user allowed bridged provisioning network.
+     if [[ "${ENABLE_NATED_PROVISIONING_NETWORK:-false}" = "false" ]]; then
+       sudo brctl addbr provisioning
```
tuminoid, line 20: Indentation should be 4.

vm-setup/roles/common/tasks/main.yml
```diff
@@ -32,9 +49,9 @@
 
 - name: "Append extra networks when EXTRA_NETWORK_NAMES is configured"
   set_fact:
-    networks: "{{ networks + extra_networks}}"
+    networks: "{{ networks + extra_networks }}"
   when: generate_extra_networks
 
-- name: "Show networks data for debugging"
+- name: "Show networks data for debugging (commond role)"
```
tuminoid, line 55: `commond` role, typo?

tuminoid (commented):
Just minor nits.

vm-setup/roles/common/defaults/main.yml
Rozzii: Although that will introduce an additional export in the bash common.sh script or in one of the numbered bash scripts, but I guess as we define most of the defaults for env vars (if not all) in those scripts it is fine.

03_launch_mgmt_cluster.sh
```diff
@@ -140,6 +140,9 @@ function launch_ironic() {
   pushd "${BMOPATH}"
 
     # Update Configmap parameters with correct urls
+    # Vriable names inserted into the configmap might have different
```
lentzi90, line 143: nit
```suggestion
    # Variable names inserted into the configmap might have different
```

ubuntu_bridge_network_configuration.sh
Rozzii: Done.

vm-setup/roles/common/tasks/main.yml
Rozzii: Done.

03_launch_mgmt_cluster.sh
Rozzii: Done.

vm-setup/roles/common/defaults/main.yml
```diff
@@ -79,4 +106,11 @@ external_network:
...
 
-networks: "{{ provisioning_network + external_network }}"
+# Provisioning network is bridged and external network is nated
+networks_mixed: "{{ provisioning_network_bridge + external_network }}"
+# Both networks are nated
+networks_nated: "{{ provisioning_network_nat + external_network }}"
+# Enable only nated networks
+networks_nat_only: "{{ lookup('env', 'ENABLE_NATED_PROVISIONING_NETWORK') | bool }}"
+# Placeholder for the eventually selected network composition
+networks: ""
```
lentzi90, line 116: If we do it like this instead we don't need the `set_fact` tasks.
```suggestion
networks: "{{ networks_nat_only | ternary(networks_nated, networks_mixed) }}"
```

vm-setup/roles/common/tasks/main.yml
```diff
@@ -3,11 +3,25 @@
...
+- name: Set networks based on networks_mixed
+  set_fact:
+    networks: "{{ networks_mixed }}"
+  when: not networks_nat_only
+
+- name: Set networks based on networks_nated
+  set_fact:
+    networks: "{{ networks_nated }}"
+  when: networks_nat_only
+
```
lentzi90, line 24:
```suggestion
```

vm-setup/roles/common/defaults/main.yml
Rozzii: I am not sure I like these very abstracted  Ansible filters... for me it is more in line with the sequential execution of the tasks  to set these values when they are needed  but I see your point but I personally don't like it (no technical issue just personal preference).

lentzi90: I see. I find it much easier to understand what is going on when setting it like this directly instead of having to hunt down the `set_fact`s. The ternary operator is not very abstract even, it is just uncommon to spell it out like they do in ansible. Another name for it is "inline if" and a [more common way](https://en.wikipedia.org/wiki/Ternary_conditional_operator) to express it would be something like `a ? b : c`.

Would it be better if we use actual if/else inline? Like this:

```suggestion
networks: "{{ networks_nated if networks_nat_only else networks_mixed }}"
```

Rozzii: Okay, I can settle for the regular inline if that is fine for me too.

lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm

PR #1176, opened 2023-03-15
Unify ansible extra_var handling

Rozzii:
It has been noticed that a firewalld related ansible extra_var was passed to Ansible in json format instead of regular key=value format.

The root cause of the isse was that this variable was handled as bool type and directly used in ansible "when" conditions.

I beleive it is a less complex aproach if ansible arguments are all strings in bash, then if the variable would be used in a condition in ansible, it can be type casted via a type filter at the place of use thus the cotributors modifying the bash code have no need to one by one investigate all the use-cases af all possible extra_var of ansible.

It has been decided to unify the extra_var syntax in the bash scripts to follow the "key=value" format.

This commit also removes an unnecessary external subnet related extra_var as it was actually both redundant and acted as a noop.

Co-authored-by: Tuomo Tanskanen <tuomo.tanskanen@est.tech>
Co-authored-by: @tuminoid

### DIFF
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -42,14 +42,11 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -i vm-setup/inventory.ini \
     -b -v vm-setup/teardown-playbook.yml
 
-if [ "$USE_FIREWALLD" == "False" ]; then
- ANSIBLE_FORCE_COLOR=true ansible-playbook \
-    -e "{use_firewalld: $USE_FIREWALLD}" \
-    -e "external_subnet_v4: ${EXTERNAL_SUBNET_V4}" \
+ANSIBLE_FORCE_COLOR=true ansible-playbook \
+    -e "use_firewalld=${USE_FIREWALLD}" \
     -e "firewall_rule_state=absent" \
     -i vm-setup/inventory.ini \
     -b -v vm-setup/firewall.yml
-fi
 
 # There was a bug in this file, it may need to be recreated.
 if [[ $OS == "centos" || $OS == "rhel" ]]; then
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -172,9 +172,8 @@ EOF
 fi
 
 ANSIBLE_FORCE_COLOR=true ansible-playbook \
-    -e "{use_firewalld: $USE_FIREWALLD}" \
-    -e "external_subnet_v4: ${EXTERNAL_SUBNET_V4}" \
-    -e "provisioning_subnet: ${PROVISIONING_NETWORK}" \
+    -e "provisioning_subnet=${PROVISIONING_NETWORK}" \
+    -e "use_firewalld=${USE_FIREWALLD}" \
     -i vm-setup/inventory.ini \
     -b vm-setup/firewall.yml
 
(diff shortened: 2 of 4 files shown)


tuminoid (commented):
Thanks for fixing this. One off-topic-y nit.

host_cleanup.sh
```diff
@@ -42,10 +42,9 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -i vm-setup/inventory.ini \
     -b -v vm-setup/teardown-playbook.yml
 
-if [ "$USE_FIREWALLD" == "False" ]; then
+if [ "$USE_FIREWALLD" == "false" ]; then
```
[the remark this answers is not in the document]
Rozzii: done

host_cleanup.sh
```diff
@@ -42,10 +42,9 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -i vm-setup/inventory.ini \
     -b -v vm-setup/teardown-playbook.yml
 
-if [ "$USE_FIREWALLD" == "False" ]; then
+if [ "$USE_FIREWALLD" = "false" ]; then
  ANSIBLE_FORCE_COLOR=true ansible-playbook \
-    -e "{use_firewalld: $USE_FIREWALLD}" \
-    -e "external_subnet_v4: ${EXTERNAL_SUBNET_V4}" \
+    -e "use_firewalld=${USE_FIREWALLD}" \
```
mboukhalfa, line 47: this condition does not make sense for me if we are checking in ansible playboot use firewalld why we need to check here ?

mboukhalfa: I parameterized cleaning firewalld as well https://github.com/metal3-io/metal3-dev-env/pull/1179

Rozzii: Okay, your PR does other stuff too, so let's remove this firewall condition here and then you don't need to deal with @tuminoid 's comments related to the argument formatting.

tuminoid (approved):
/lgtm

host_cleanup.sh
```diff
@@ -42,14 +42,11 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
     -i vm-setup/inventory.ini \
     -b -v vm-setup/teardown-playbook.yml
 
-if [ "$USE_FIREWALLD" == "False" ]; then
- ANSIBLE_FORCE_COLOR=true ansible-playbook \
-    -e "{use_firewalld: $USE_FIREWALLD}" \
-    -e "external_subnet_v4: ${EXTERNAL_SUBNET_V4}" \
+ANSIBLE_FORCE_COLOR=true ansible-playbook \
```
mboukhalfa, line 45: I am afraid if we remove the condition it will re enable the firewalld instead of cleaning them not yet sure if we can use disabled with firewalld in the cleaning step

mboukhalfa: My question in the other PR if I enabled something was enabled by default does make sense to disable it in the cleaning step ?

Rozzii: if `use_firewalld` is true then the firewall rules would not be cleaned up correctly if it is nod true it doesn't matter anyways because then the iptable rules will be used.

mboukhalfa (approved):
Assuming that https://github.com/metal3-io/metal3-dev-env/pull/1179 will follow to fix disabling firewalld rules instead of re-applying then once again in the cleaning

kashifest (approved):

PR #1179, opened 2023-03-16
Add firewalld cleanup

mboukhalfa:
Currently, only the iptables rules could be configured to the `present` state during setup and `absent` during cleanup, while the firewalld rules were not being properly cleaned. To address this issue, we have introduced new parameters in `firewalld.yaml` that set the rules state to 'disabled' during the cleanup step.

### DIFF
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -12,8 +12,8 @@ remove_ironic_containers
 # Remove existing pod
 if [[ "${CONTAINER_RUNTIME}" == "podman" ]]; then
   for pod in ironic-pod infra-pod; do
-    if  sudo "${CONTAINER_RUNTIME}" pod exists "${pod}" ; then
-        sudo "${CONTAINER_RUNTIME}" pod rm "${pod}" -f
+    if sudo "${CONTAINER_RUNTIME}" pod exists "${pod}"; then
+      sudo "${CONTAINER_RUNTIME}" pod rm "${pod}" -f
     fi
   done
 fi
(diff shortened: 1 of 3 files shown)


host_cleanup.sh
```diff
@@ -31,40 +31,37 @@ if [ "${CAPM3_RUN_LOCAL}" = true ]; then
...
-if [ "$USE_FIREWALLD" == "False" ]; then
- ANSIBLE_FORCE_COLOR=true ansible-playbook \
-    -e "{use_firewalld: $USE_FIREWALLD}" \
-    -e "external_subnet_v4: ${EXTERNAL_SUBNET_V4}" \
-    -e "firewall_rule_state=absent" \
-    -i vm-setup/inventory.ini \
-    -b -v vm-setup/firewall.yml
-fi
+ANSIBLE_FORCE_COLOR=true ansible-playbook \
+  -e "{use_firewalld: $USE_FIREWALLD}" \
```
tuminoid, line 45: Remove use of JSON syntax, liike in Adam's patch? It is bad habit to mix syntaxes as we see on the next line.

host_cleanup.sh
```diff
@@ -31,40 +31,37 @@ if [ "${CAPM3_RUN_LOCAL}" = true ]; then
...
+  -e "external_subnet_v4: ${EXTERNAL_SUBNET_V4}" \
```
tuminoid, line 46: This is not JSON nor Bash format, and doesn't work. `external_subnet_v4` extra was also removed in Adam's patch, could it be removed here too?

tuminoid (changes requested):

Rozzii:
let's get merged mine first https://github.com/metal3-io/metal3-dev-env/pull/1176 that handles the formatting stuff and this firewall condition and after it is merged and you have re-based yours your PR can handle the firewall rule's (enable/disable) part.

mboukhalfa:
@elfosardo

host_cleanup.sh
mboukhalfa: I think these changes should be addressed only in one PR and then rebase the second PR once the first merged I do not want to replicate adams changes here else his PR will be meaningless ?

tuminoid: @Rozzii 's PR is now merged, please rebase.

mboukhalfa: rebased

tuminoid (approved):
Letting @Rozzii lgtm this, no objections from me.

furkatgofurov7 (approved):

tuminoid (approved):
/lgtm

PR #1181, opened 2023-03-17
Allow setting ansible verbosity for troubleshooting

elfosardo:


lib/common.sh
```diff
@@ -335,6 +335,11 @@ RESULT_STR=""
 # Avoid printing skipped Ansible tasks
 export ANSIBLE_DISPLAY_SKIPPED_HOSTS="no"
 
+# Set Ansible verbosity.
+# For more info please see:
+# https://docs.ansible.com/ansible/latest/reference_appendices/config.html#envvar-ANSIBLE_VERBOSITY
+export ANSIBLE_VERBOSITY=${ANSIBLE_VERBOSITY:-0}
```
dtantsur, line 341: Do you want to use this variable anyhow? Because otherwise this is no-op: any available variables will reach ansible anyway.

elfosardo: I'm going to cancel this PR
I did a test and it was not setting the  ANSIBLE_VERBOSITY correctly and I thought the variables set here were influencing that, but then I realized I made a typo.....

PR #1182, opened 2023-03-17
Fix redefinition of OS variable during krew install

elfosardo:
We should not redefine the OS variable during krew install because it breaks tests based on its original value assigned in common.sh Let's create and use a different variable instead with a smaller scope.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -96,9 +96,9 @@ source lib/releases.sh
 ## Install krew
 if ! kubectl krew > /dev/null 2>&1; then
   cd "$(mktemp -d)" &&
-  OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
+  KERNEL_OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
   ARCH="$(uname -m | sed -e 's/x86_64/amd64/' -e 's/\(arm\)\(64\)\?.*/\1\2/' -e 's/aarch64$/arm64/')" &&
-  KREW="krew-${OS}_${ARCH}" &&
+  KREW="krew-${KERNEL_OS}_${ARCH}" &&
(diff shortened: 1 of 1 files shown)


elfosardo:
@dtantsur @Rozzii @mboukhalfa @furkatgofurov7 this fix an issue that breaks execution on generic ubuntu VM because the OS variable is redefined as "linux" instead of being kept as "ubuntu" as defined in common.sh
the test on https://github.com/metal3-io/metal3-dev-env/blob/main/01_prepare_host.sh#L224-L228 does not work otherwise

dtantsur (approved):

01_prepare_host.sh
```diff
@@ -96,9 +96,9 @@ source lib/releases.sh
 ## Install krew
 if ! kubectl krew > /dev/null 2>&1; then
   cd "$(mktemp -d)" &&
-  OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
+  KERNEL_OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
```
dtantsur, line 99: I wonder if we ever expect anything other than Linux here. Like, what?

mboukhalfa:
/lgtm 
thanks for the fix

01_prepare_host.sh
tuminoid: As discussed while back about supported OS, we're currently limiting ourselves to Ubuntu + Centos/RHEL already, so there would be no harm of hard coding it to Linux, but on the other hand, this doesn't hurt anyone either.

jayofdoom:
Thanks for all the good work to enhance this so we can get tests in Ironic! I'm excited about Ironic and Metal3 working more closely together.

furkatgofurov7 (approved):
Thanks for the PR, @elfosardo 

/approve

PR #1192, opened 2023-03-29
Fix BMO branching and tags for different CAPM3 release branches

kashifest:
The idea of this PR is to make sure that BMO version used in different branch tests match with the BMO version imported in CAPM3 go modules. This PR also introduces keepalived image tag and syncs it with BMO image tag and makes sure the image tags for these two image are in sync with BMO release version.

PR #1195, opened 2023-03-31
Fix BMO local image building workflow

kashifest:
#1192 introduced a separate workflow for BMO local image building. But it seems too early to checkout to local branch because BMO repo has not been cloned yet. Reverting it to get the CI going.


lentzi90 (commented):
/approve
Let's get the CI back on track and then fix building the BMO image

smoshiur1237 (commented):
/lgtm

PR #1199, opened 2023-04-04
Allow custom source to build ironic-image from source

elfosardo:


elfosardo:
this is also being tested here -> https://review.opendev.org/c/openstack/ironic/+/879162

Rozzii:
/hold
We plan to propose the removal of image building from dev-env, I also think it is outside of the scope of the dev env, we also plan to decouple dev-env from CI.

Rozzii:
We have discussed that this can go forward, but note that the local image building feature might change in the future.

Sunnatillo (commented):
/lgtm

elfosardo:
@furkatgofurov7 @lentzi90 @kashifest please have a look when you get a chance, thanks!

02_configure_host.sh
```diff
@@ -245,7 +245,16 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
...
+    # NOTE(rpittau): to customize the source origin we need to copy the source code we
+    # want to use into the sources directory under the ironic-image repository.
+    for CODE_SOURCE_VAR in $(env | grep -E '^IRONIC_SOURCE=|^IRONIC_INSPECTOR_SOURCE=|^SUSHY_SOURCE=' | grep -o "^[^=]*"); do
+      CODE_SOURCE="${!CODE_SOURCE_VAR}"
+      SOURCE_DIR_DEST="${CODE_SOURCE##*/}"
+      cp -a "$CODE_SOURCE" "./sources/${SOURCE_DIR_DEST}"
+      CUSTOM_SOURCE_ARGS+="--build-arg ${CODE_SOURCE_VAR}=${SOURCE_DIR_DEST} "
+    done
+    #shellcheck disable=SC2086
+    sudo "${CONTAINER_RUNTIME}" build --build-arg INSTALL_TYPE=source ${CUSTOM_SOURCE_ARGS:-} -t "${!IMAGE_VAR}:latest" -t "${!IMAGE_VAR}:${IMAGE_GIT_HASH}_${IMAGE_DATE}" . -f ./Dockerfile
```
lentzi90, line 257: It is unnecessary here to disable shellcheck. We can instead do it properly with a bash array. Something like this should do it:

```bash
declare -a CUSTOM_SOURCE_ARGS
for CODE_SOURCE_VAR in $(env | ...); do
    ...
    CUSTOM_SOURCE_ARGS+=("--build-arg ...")
done
sudo "${CONTAINER_RUNTIME}" build ... "${CUSTOM_SOURCE_ARGS[@]}"
```

02_configure_host.sh
```diff
@@ -245,7 +245,16 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
 
   # Support building ironic-image from source
   if [[ "${IMAGE}" =~ "ironic" ]] && [[ ${IRONIC_FROM_SOURCE:-} == "true" ]]; then
-    sudo "${CONTAINER_RUNTIME}" build --build-arg INSTALL_TYPE=source -t "${!IMAGE_VAR}:latest" -t "${!IMAGE_VAR}:${IMAGE_GIT_HASH}_${IMAGE_DATE}" . -f ./Dockerfile
+    # NOTE(rpittau): to customize the source origin we need to copy the source code we
+    # want to use into the sources directory under the ironic-image repository.
+    for CODE_SOURCE_VAR in $(env | grep -E '^IRONIC_SOURCE=|^IRONIC_INSPECTOR_SOURCE=|^SUSHY_SOURCE=' | grep -o "^[^=]*"); do
+      CODE_SOURCE="${!CODE_SOURCE_VAR}"
+      SOURCE_DIR_DEST="${CODE_SOURCE##*/}"
+      cp -a "$CODE_SOURCE" "./sources/${SOURCE_DIR_DEST}"
```
lentzi90, line 253: nit for consistency
```suggestion
      cp -a "${CODE_SOURCE}" "./sources/${SOURCE_DIR_DEST}"
```

02_configure_host.sh
```diff
@@ -245,7 +245,16 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*") ; do
 
   # Support building ironic-image from source
   if [[ "${IMAGE}" =~ "ironic" ]] && [[ ${IRONIC_FROM_SOURCE:-} == "true" ]]; then
-    sudo "${CONTAINER_RUNTIME}" build --build-arg INSTALL_TYPE=source -t "${!IMAGE_VAR}:latest" -t "${!IMAGE_VAR}:${IMAGE_GIT_HASH}_${IMAGE_DATE}" . -f ./Dockerfile
+    # NOTE(rpittau): to customize the source origin we need to copy the source code we
+    # want to use into the sources directory under the ironic-image repository.
+    for CODE_SOURCE_VAR in $(env | grep -E '^IRONIC_SOURCE=|^IRONIC_INSPECTOR_SOURCE=|^SUSHY_SOURCE=' | grep -o "^[^=]*"); do
```
lentzi90, line 250: I think this would be far easier to understand if we skip the grep and just enumerate the variables directly. Maybe even skip the loop. It would be slightly more/repeated code but much "simpler".

Just a thought though. I am fine with it this way also and it is in line with what we do in other places in the repo.

02_configure_host.sh
elfosardo: have you tested it?
I tried that but it didn't work as it's still read as a single string, maybe I did something wrong?

02_configure_host.sh
elfosardo: thanks for the suggestion, maybe in a followup?

lentzi90: Sure! :slightly_smiling_face:

02_configure_host.sh
lentzi90: Yes, we use it in the 03 script. Here we create the array: https://github.com/metal3-io/metal3-dev-env/blob/3c9dcb188c05a559383257ebdacfa3897fbfee14/03_launch_mgmt_cluster.sh#L17-L28
And here we use it: https://github.com/metal3-io/metal3-dev-env/blob/3c9dcb188c05a559383257ebdacfa3897fbfee14/03_launch_mgmt_cluster.sh#L72
It must be an array though, otherwise it is always treated as a single string when quoted.

lentzi90: Never mind. I didn't consider the empty case! It would break if the array is empty, since you then end up with just `""`.

elfosardo:
@Rozzii @Sunnatillo when you have a moment please check this again, thanks!

Sunnatillo (commented):
/lgtm

PR #1206, opened 2023-04-11
Fix manifest image update and image tag for Keepalived

kashifest:
when a PR is going in BMO, keepalived image is also rebuilt locally it was not correctly updated in manifest. This PR would fix it. 

This PR also fixes a bug which was exporting keepalived image before the correct tag was determined.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -256,7 +256,6 @@ export IRONIC_IMAGE=${IRONIC_IMAGE:-"${CONTAINER_REGISTRY}/metal3-io/ironic:${IR
 export IRONIC_CLIENT_IMAGE=${IRONIC_CLIENT_IMAGE:-"${CONTAINER_REGISTRY}/metal3-io/ironic-client"}
 export IRONIC_DATA_DIR="$WORKING_DIR/ironic"
 export IRONIC_IMAGE_DIR="$IRONIC_DATA_DIR/html/images"
-export IRONIC_KEEPALIVED_IMAGE="${IRONIC_KEEPALIVED_IMAGE:-${CONTAINER_REGISTRY}/metal3-io/keepalived}"
 export IRONIC_NAMESPACE="${IRONIC_NAMESPACE:-baremetal-operator-system}"
 export NAMEPREFIX="${NAMEPREFIX:-baremetal-operator}"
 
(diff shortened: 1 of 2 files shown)


tuminoid (approved):
/lgtm

lentzi90 (commented):
Should we not do a separate if statement for this?

```bash
if [ -n "${IRONIC_KEEPALIVED_LOCAL_IMAGE:-}" ]; then
  update_component_image Keepalived "${IRONIC_KEEPALIVED_LOCAL_IMAGE}"
else
  update_component_image Keepalived "${IRONIC_KEEPALIVED_IMAGE}"
fi
```
I'm thinking it would be very confusing if someone is trying to test keepalived and not BMO if we have the same check for both.

kashifest:
> Should we not do a separate if statement for this?
> 
> ```shell
> if [ -n "${IRONIC_KEEPALIVED_LOCAL_IMAGE:-}" ]; then
>   update_component_image Keepalived "${IRONIC_KEEPALIVED_LOCAL_IMAGE}"
> else
>   update_component_image Keepalived "${IRONIC_KEEPALIVED_IMAGE}"
> fi
> ```
> 
> I'm thinking it would be very confusing if someone is trying to test keepalived and not BMO if we have the same check for both.

Good point, fixed.

lentzi90 (commented):
/approve

lib/common.sh
```diff
@@ -256,7 +256,6 @@ export IRONIC_IMAGE=${IRONIC_IMAGE:-"${CONTAINER_REGISTRY}/metal3-io/ironic:${IR
 export IRONIC_CLIENT_IMAGE=${IRONIC_CLIENT_IMAGE:-"${CONTAINER_REGISTRY}/metal3-io/ironic-client"}
 export IRONIC_DATA_DIR="$WORKING_DIR/ironic"
 export IRONIC_IMAGE_DIR="$IRONIC_DATA_DIR/html/images"
-export IRONIC_KEEPALIVED_IMAGE="${IRONIC_KEEPALIVED_IMAGE:-${CONTAINER_REGISTRY}/metal3-io/keepalived}"
```
kashifest, line 259: We don't need to export keepalived image here since after the correct tag's are set, keepalived image var is exported at line 291

kashifest:
@lentzi90 after your approval , I have pushed a new commit. Ideally the approve label should have been removed. But it is there. I am putting a hold for you to review again. Thanks. 
/hold

lib/common.sh
tuminoid: Good find.

tuminoid (approved):
/lgtm

lentzi90 (commented):
/approve
/hold cancel

PR #1208, opened 2023-04-12
Fix BMOBRANCH based on CAPM3 release branch. 

kashifest:
Previously this was being set from project-infra but only for main branch periodic tests. This makes sure BMOBRANCH is main not only for periodic tests but also for PR jobs. Related to https://github.com/metal3-io/project-infra/pull/507.

Also it does not make any sense to fetch BMORELEASE in lib/releases anymore since different CAPM3 branch are tested with different BMO releases. So the easiest and cleanest way would be to set it upfront based on the branch.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -174,7 +174,7 @@ EOF
   fi
 
   # Copy the generated configmap for ironic deployment
-  if [[ ${BMOBRANCH} == "v0.1.2" ]] || [[ ${BMOBRANCH} == "v0.1.1" ]]; then # BMORELEASE until v0.1.2 used the old path TODO(mboukhalfa) can be removed after new bmo release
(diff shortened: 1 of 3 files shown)


tuminoid (approved):
/lgtm

lentzi90 (commented):
I'm a bit afraid this will bite us at some point when we want to test with a specific `BMOBRANCH`.
How does it work with BMO PRs when we need to check that specific code?
I *think* it would still clone and use main with this change but I'm not sure.

Ref to code that does the cloning: https://github.com/metal3-io/metal3-dev-env/blob/8cd90a1c076feaed7302666316baa7c0b6f81671/02_configure_host.sh#L206

tuminoid (commented):
/lgtm

lentzi90 (commented):
/approve

/retitle Set BMOBRANCH to main by default

Sunnatillo (commented):
/lgtm

PR #1209, opened 2023-04-13
add lib/debug.sh

tuminoid:
Add debug library that can print out variables and track their values to help debug various problems in dev-env more efficiently.

### DIFF
--- a/lib/debug.sh
+++ b/lib/debug.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+#
+# usage:
+#
+# 1. source this debug.sh in other scripts, and run
+# "debug_store_vars <prefix text>"
+# to store current environment variables, and print those that have
+# changed since the previous call to the function
+#
+# 2. you can also print the currently stored variables with
+# "debug_print_stored_vars"
+#
+# 3. at the end of configuration and setup, call
+# "debug_print_all_vars"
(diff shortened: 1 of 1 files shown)


tuminoid:
/test-centos-e2e-integration-main

/test-ubuntu-integration-release-1-3

/test-ubuntu-integration-main
/test-centos-integration-main

Let's see what all of these tests are made of.

lentzi90 (commented):
Nice!
/approve

kashifest:
I dont see it being called from anywhere? So whats the plan?

tuminoid:
> I dont see it being called from anywhere? So whats the plan?

It is added as a library so it can be activated to aid in debugging. It is not actively called anywhere by default.

Sunnatillo:
Great job @tuminoid. Would it be better if we rename file name `debug_vars.sh` for specifying its usage? 
Otherwise LGTM.

kashifest:
/lgtm

> > Great job @tuminoid. Would it be better if we rename file name `debug_vars.sh` for specifying its usage? Otherwise LGTM.
> 
> It could be, but I doubt we will introduce `debug_something_else.sh` there, but rather add to the same `debug.sh` whatever debug funcs we come up with, so IMO this is better.

Agree

PR #1210, opened 2023-04-13
fix duplicate sourcing of network.sh

tuminoid:
Fix duplicate sourcing of network.sh and some spammy shellcheck ignores.


lentzi90 (commented):
/approve

PR #1211, opened 2023-04-13
Fix pulling container images

kashifest:
We dont need to manually check if the image exists and then pull the image. The current code is buggy. If images are prepulled in CI images, CRI can check if the digest is new and will pull newer image accordingly.


lib/common.sh
```diff
@@ -527,27 +527,14 @@ differs(){
...
 
-# If a given container with tag doesn't exist locally, pull it.
-# Otherwise, do nothing.
-# Helps conserve number of API calls to DockerHub to avoid hitting rate limit.
-#
 # Inputs:
 # - Full name of a Docker/podman/crictl image including tag
 #
-pull_container_image_if_missing() {
+pull_container_image() {
```
mboukhalfa, line 533: this should be called somewhere ?

kashifest: yes the file was left out from the commit.

tuminoid (dismissed):
This one was sneaky as it was hidden in CI image build, and caused very old images to be used. Good find!

/lgtm

Rozzii (dismissed):

01_prepare_host.sh
```diff
@@ -229,7 +229,7 @@ fi
 # Pulling all the images except any local image.
 for IMAGE_VAR in $(env | grep -v "_LOCAL_IMAGE=" | grep "_IMAGE=" | grep -o "^[^=]*") ; do
   IMAGE="${!IMAGE_VAR}"
-  pull_container_image_if_missing "${IMAGE}"
+  pull_container_image "${IMAGE}"
```
lentzi90, line 232: It is not worth having a separate function for this IMO. Just put the command directly and get rid of the function completely.
```suggestion
  sudo "${CONTAINER_RUNTIME}" pull "${IMAGE}"
```

lentzi90 (changes requested):

Sunnatillo (commented):
/lgtm

lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm

PR #1212, opened 2023-04-14
Remove leftover ISO images from the default pool

dtantsur:
The redfish emulator creates these. In case of a tear down, they will
not be deleted and may occupy a lot of disk space.


Sunnatillo:
/lgtm
/ test-ubuntu-e2e-integration-main

kashifest (approved):

PR #1215, opened 2023-04-17
Remove release-0.5 mentions in dev-env

mquhuy:
As title

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -139,10 +139,7 @@ export CAPM3_BASE_URL="${CAPM3_BASE_URL:-metal3-io/cluster-api-provider-metal3}"
 export CAPM3REPO="${CAPM3REPO:-https://github.com/${CAPM3_BASE_URL}}"
 export CAPM3RELEASEBRANCH="${CAPM3RELEASEBRANCH:-main}"
 
-if [[ "${CAPM3RELEASEBRANCH}" == "release-0.5" ]]; then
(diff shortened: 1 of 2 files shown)


Sunnatillo (dismissed):
/lgtm

adilGhaffarDev:
LGTM

kashifest (dismissed):

furkatgofurov7:
/hold

Do not we still need this as per: https://github.com/metal3-io/cluster-api-provider-metal3/blob/5ebb78cdbee8fb72ed3a6191d142f86eb4d635f1/README.md#upgrade-tests or the notes has to be updated if that is not accurate

Sunnatillo:
> /hold
> 
> Do not we still need this as per: https://github.com/metal3-io/cluster-api-provider-metal3/blob/5ebb78cdbee8fb72ed3a6191d142f86eb4d635f1/README.md#upgrade-tests or the notes has to be updated if that is not accurate

yes. But that part handled in capm3, we do not use dev env to initialize v1alpha5 cluster for upgrade tests.

furkatgofurov7:
> yes. But that part handled in capm3, we do not use dev env to initialize v1alpha5 cluster for upgrade tests. v1alpha5 is not supported. But we support upgrading from v1alpha5. It is ok to remove it from dev-env.

I am not sure I got that and think notes in CAPM3 has to be updated. If it is not supported then there should not be an upgrade test mentions in CAPM3.. 

/hold cancel

PR #1216, opened 2023-04-17
Add manifest collection after re-pivot

smoshiur1237:
This change will allow the manifest collection both before pivot and after re-pivot.


smoshiur1237:
//test-features-ubuntu

Sunnatillo (commented):
/lgtm

kashifest:
> /hold
> Is this going to collect manifest from target cluster after repivot or management cluster?

/hold cancel
Let’s take this in. In future we should probably take the manifests from target cluster right before repivot.

PR #1161, opened 2023-03-01
🐛 Workaround to fix github api limit for capi release

mboukhalfa:
A workaround to unlock the CI by increasing the api response limit

PR #1219, opened 2023-04-17
Simplify get latest releases function

Sunnatillo:
This PR simplifies get_latest_release function. Also reduces number of queries to github api.
Fixes "Failed to fetch CAPI release from Github" issue


lib/releases.sh
```diff
@@ -1,44 +1,26 @@
...
+  url="${1}?per_page=1000"
+  
+  # This gets the latest release as vx.y.z or vx.y.z-rc.0, including any version with a suffix starting with - , for example -rc.0
+  # The order is exactly as released in Github.
+  # Downside is that selecting official releases only isn't possible, while pre-release
+  # selection is possible given specific enough prefix, like v1.3.0-pre
+
   if [ -z "${GITHUB_TOKEN:-}" ]; then
-    last_page="$(curl -s -I "${1}" | grep '^link:' | sed -e 's/^link:.*page=//g' -e 's/>.*$//g')"
+    release_tag=$(curl -sL "${url}" | jq .[].name -r | egrep "^v1.4." -m 1)
```
tuminoid: egrep is deprecated, use `grep -E` instead.

lib/releases.sh
```diff
@@ -1,44 +1,26 @@
...
   else
-    last_page="$(curl -s -I "${1}" -H "Authorization: token $GITHUB_TOKEN" | grep '^link:' | sed -e 's/^link:.*page=//g' -e 's/>.*$//g')"
+    release_tag=$(curl -sL "${url}" -H "Authorization: token $GITHUB_TOKEN"  | jq .[].name -r | egrep "^v1.4." -m 1)
```
tuminoid: Curlys for GITHUB_TOKEN

lib/releases.sh
```diff
@@ -1,44 +1,26 @@
...
-    # selection is possible given specific enough prefix, like v1.3.0-pre
-    release_tag="$(echo "${release}" | jq -r "[.[].tag_name | select( startswith(\"${2:-}\"))] | .[0]")"
-    # if release tag found
-    if [[ "${release_tag}" != "null" ]]; then
-      break
-    fi
-
-  done
 
   set -x
```
tuminoid, line 18: `set -x` need to be removed as `set +x` was removed.

lib/releases.sh
```diff
@@ -1,44 +1,26 @@
...
   # if not found
   if [[ "${release_tag}" == "null" ]]; then
```
tuminoid, line 20: This null check probably doesn't work anymore, as `release_tag` is no longer coming from `select()` of jq which produced the `null`.

lib/releases.sh
Sunnatillo: done

lib/releases.sh
Sunnatillo: done

lib/releases.sh
Sunnatillo: removed

lib/releases.sh
```diff
@@ -1,44 +1,20 @@
...
+  url="${1}?per_page=1000"
+  
+  # This gets the latest release as vx.y.z or vx.y.z-rc.0, including any version with a suffix starting with - , for example -rc.0
+  # The order is exactly as released in Github.
+  # Downside is that selecting official releases only isn't possible, while pre-release
+  # selection is possible given specific enough prefix, like v1.3.0-pre
+
   if [ -z "${GITHUB_TOKEN:-}" ]; then
-    last_page="$(curl -s -I "${1}" | grep '^link:' | sed -e 's/^link:.*page=//g' -e 's/>.*$//g')"
+        release_tag=$(curl -sL "${url}" | jq .[].name -r | grep -E "${2:-}" -m 1)
```
tuminoid, line 13: You want to fail if `$2` is empty, since then the caller has not passed the release to search for.

Just add to top of the function:
`local release="${2:?no release given}"` and it'll fail nicely if its not passed, then use `${release}` in grep.

lib/releases.sh
```diff
@@ -1,44 +1,20 @@
...
   else
-    last_page="$(curl -s -I "${1}" -H "Authorization: token $GITHUB_TOKEN" | grep '^link:' | sed -e 's/^link:.*page=//g' -e 's/>.*$//g')"
+      release_tag=$(curl -sL "${url}" -H "Authorization: token $GITHUB_TOKEN"  | jq .[].name -r | grep -E "${2:-}" -m 1)
```
tuminoid, line 15: Curlys

mboukhalfa (changes requested):
Here you are removing the api pagination and keep checking only  the last 100 releases created! so the function will fail if the target release is not in the first 100 batch ? 
if the idea is just to limit the api request number you can increase the per_page to the max (i.e 100) while keeping the pagination since the current implementation will stop requesting the other pages if the target release found.
if you can guarantee that the target release will always be in the first api page with 100 elements you should make it clear in the function name for example :
`function get_latest_release_from_first_githubapi_page() {`
 and add error message if the release was not found stating not found in the first page !

lib/releases.sh
```diff
@@ -1,44 +1,27 @@
...
+  local release="${2:?no release given}"
+  
+  # This gets the latest release as vx.y.z or vx.y.z-rc.0, including any version with a suffix starting with - , for example -rc.0
+  # The order is exactly as released in Github.
+  # Downside is that selecting official releases only isn't possible, while pre-release
+  # selection is possible given specific enough prefix, like v1.3.0-pre
+
   if [ -z "${GITHUB_TOKEN:-}" ]; then
-    last_page="$(curl -s -I "${1}" | grep '^link:' | sed -e 's/^link:.*page=//g' -e 's/>.*$//g')"
+    release_tag=$(curl -sL "${url}" | jq .[].name -r | grep -E "${release}" -m 1)
```
tuminoid, line 20: Sorry for nitpicking this much, but please put `.[].name` in quotes for safer handling, and add the missing `^` in front of `${release}` so it matches releases from the start only as it should.

```suggestion
    release_tag=$(curl -sL "${url}" | jq ".[].name" -r | grep -E "^${release}" -m 1)
```

Sunnatillo:
> Here you are removing the api pagination and keep checking only the last 100 releases created! so the function will fail if the target release is not in the first 100 batch ? if the idea is just to limit the api request number you can increase the per_page to the max (i.e 100) while keeping the pagination since the current implementation will stop requesting the other pages if the target release found. if you can guarantee that the target release will always be in the first api page with 100 elements you should make it clear in the function name for example : `function get_latest_release_from_first_githubapi_page() {` and add error message if the release was not found stating not found in the first page !

CAPI  only maintain two latest minor release of current api version, and  latest minor release for older api version. We also follow CAPI. Considering newest releases will be always on top, I do not see necessity adding complications to function. IMHO, 100 latest release is good enough. I can add a note to function saying that it will check only latest 100 releases.

mboukhalfa:
> > Here you are removing the api pagination and keep checking only the last 100 releases created! so the function will fail if the target release is not in the first 100 batch ? if the idea is just to limit the api request number you can increase the per_page to the max (i.e 100) while keeping the pagination since the current implementation will stop requesting the other pages if the target release found. if you can guarantee that the target release will always be in the first api page with 100 elements you should make it clear in the function name for example : `function get_latest_release_from_first_githubapi_page() {` and add error message if the release was not found stating not found in the first page !
> 
> CAPI only maintain two latest minor release of current api version, and latest minor release for older api version. We also follow CAPI. Considering newest releases will be always on top, I do not see necessity adding complications to function. IMHO, 100 latest release is good enough. I can add a note to function saying that it will check only latest 100 releases.

Please note that although we no longer support v1alpha5 release we still use them (e.g in upgrade) and since there are no new patches they will keep going deep in bottom of the list as new releases are pushed on top not sure when but at some point they will get moved to the second page the proper behavior of this function is to go through the next pages if the release was not found on the first. I stand by keeping this behavior I remember when we hit this issue we increased the per_page to 100 as a workaround then we decided to implement proper api pagination use! We should not regress removing this implementation in the sake of limiting GitHub requests which we can do while keeping the same behavior.

Check this issue :
https://github.com/kubernetes-sigs/cluster-api/issues/8233

and previous PRs:
https://github.com/metal3-io/metal3-dev-env/pull/1168
workaround https://github.com/metal3-io/metal3-dev-env/pull/1161
slack discussion : https://kubernetes.slack.com/archives/C8TSNPY4T/p1678105939902579

lib/releases.sh
```diff
@@ -1,46 +1,60 @@
...
+  # This gets the latest release as vx.y.z or vx.y.z-rc.0, including any version with a suffix starting with - , for example -rc.0
+  # The order is exactly as released in Github.
+  # Downside is that selecting official releases only isn't possible, while pre-release
+  # selection is possible given specific enough prefix, like v1.3.0-pre
+  release_tag=$(echo "${response_body}" | jq ".[].name" -r | grep -E "${release}" -m 1)
+  
+  # If release_tag is not found in the first page(100 releases), this condition checks from second to last_page
+  # until release_tag is found
+  if [ -z "${release_tag:-}" ]; then
+    for current_page in $(seq 2 "$last_page"); do
```
tuminoid, line 36: Missing curlys.

lib/releases.sh
```diff
@@ -1,46 +1,60 @@
...
-  if [[ "${release_tag}" == "null" ]]; then
+  # This gets the latest release as vx.y.z or vx.y.z-rc.0, including any version with a suffix starting with - , for example -rc.0
+  # The order is exactly as released in Github.
+  # Downside is that selecting official releases only isn't possible, while pre-release
+  # selection is possible given specific enough prefix, like v1.3.0-pre
+  release_tag=$(echo "${response_body}" | jq ".[].name" -r | grep -E "${release}" -m 1)
+  
+  # If release_tag is not found in the first page(100 releases), this condition checks from second to last_page
+  # until release_tag is found
+  if [ -z "${release_tag:-}" ]; then
```
tuminoid, line 35: `:-` is not necessary, the variable is defined on previous line

lib/releases.sh
```diff
@@ -1,46 +1,60 @@
...
-    # The order is exactly as released in Github.
-    # Downside is that selecting official releases only isn't possible, while pre-release
-    # selection is possible given specific enough prefix, like v1.3.0-pre
-    release_tag="$(echo "${release}" | jq -r "[.[].tag_name | select( startswith(\"${2:-}\"))] | .[0]")"
-    # if release tag found
-    if [[ "${release_tag}" != "null" ]]; then
-      break
-    fi
+  if  [ -z "${GITHUB_TOKEN:-}" ]; then
+    response=$(curl -si "$url")
```
tuminoid, line 15: Curlys.

tuminoid (commented):
Nits.

lib/releases.sh
```diff
@@ -1,46 +1,60 @@
...
+  else
+    response=$(curl -si  "$url" -H "Authorization: token ${GITHUB_TOKEN}")
```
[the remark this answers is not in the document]
Sunnatillo: > Nits.

Addressed them all

tuminoid:

Come on Cleura....

lib/releases.sh
```diff
@@ -1,46 +1,59 @@
...
+    response=$(curl -si "${url}" -H "Authorization: token ${GITHUB_TOKEN}")
+  fi
 
-  done
+  # Divide response to headers and body
+  response_headers=$(echo "${response}" | awk 'BEGIN {RS="\r\n\r\n"} NR==1 {print}')
+  response_body=$(echo "${response}" | awk 'BEGIN {RS="\r\n\r\n"} NR==2 {print}')
+  
+  # get the last page of releases from headers
+  last_page=$(echo "${response_headers}" | grep '^link:' | sed -e 's/^link:.*page=//g' -e 's/>.*$//g')
```
mboukhalfa, line 25: something is printing the full response on the logs and you know how huge it is !

mboukhalfa: it has increased the logs size from around  `482 KB` to `73,059 KB` and Jenkins is struggling show full logs 💥

Sunnatillo: /hold 
until logs are silenced. 
Thank you for noticing it.

Sunnatillo:
/hold
until logs are silenced

kashifest (approved):

lentzi90 (commented):
/approve

PR #1226, opened 2023-05-10
Exclude install requirement script from CI tests

smoshiur1237:
As we are running 01_prepare_host script when building new image, CI run doesn't need to run the install script again during CI run. So this PR will move some parts of the 01_prepare_host script to 02_configure_host script and exclude the 01_prepare_host script from integration test and feature test in dev-env.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -77,9 +77,6 @@ ANSIBLE_FORCE_COLOR=true ansible-playbook \
   -i vm-setup/inventory.ini \
   -b vm-setup/install-package-playbook.yml
 
-# shellcheck disable=SC1091
-source lib/images.sh
-
 # Add usr/local/go/bin to the PATH environment variable
 GOBINARY="${GOBINARY:-/usr/local/go/bin}"
 if [[ ! "${PATH}" =~ .*(:|^)(${GOBINARY})(:|$).* ]]; then
@@ -92,7 +89,7 @@ source lib/releases.sh
 
 ## Install krew
 if ! kubectl krew > /dev/null 2>&1; then
-  cd "$(mktemp -d)" &&
+  pushd "$(mktemp -d)" &&
   KERNEL_OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
   ARCH="$(uname -m | sed -e 's/x86_64/amd64/' -e 's/\(arm\)\(64\)\?.*/\1\2/' -e 's/aarch64$/arm64/')" &&
   KREW="krew-${KERNEL_OS}_${ARCH}" &&
@@ -104,6 +101,7 @@ if ! kubectl krew > /dev/null 2>&1; then
   krew_path_bashrc="export PATH=${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
   # Add the line if it is not already there
   grep -qxF "${krew_path_bashrc}" ~/.bashrc || echo "${krew_path_bashrc}" >> ~/.bashrc
+  popd
 fi
 
 # Allow local non-root-user access to libvirt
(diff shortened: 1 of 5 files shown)


lentzi90 (commented):
This looks overall fine to me. My only concern is that if we move this to 02, does it mean we will download more in each test? The way it is now, we have already pulled the images (maybe old version, but still) when we build the metal3 VM image. That means we have less to download when the actual test runs, since it will only have to pull images that changed and potentially just the diff.

smoshiur1237:
> This looks overall fine to me. My only concern is that if we move this to 02, does it mean we will download more in each test? The way it is now, we have already pulled the images (maybe old version, but still) when we build the metal3 VM image. That means we have less to download when the actual test runs, since it will only have to pull images that changed and potentially just the diff.

The node image in not going to be pulled here. But it needs to pull other images such as bmo, ironic, sushy etc. 02 script is doing tagging of those image and yesterday it was failing without pulling those image from nordix registry. We cannot skip that.

lentzi90:
> The node image in not going to be pulled here. But it needs to pull other images such as bmo, ironic, sushy etc. 02 script is doing tagging of those image and yesterday it was failing without pulling those image from nordix registry. We cannot skip that.

Yes I'm talking about the container images. I think it may be good that we pull them so they are stored in the metal3 image. This change would make it so that they are not stored in the metal3 image I think. Or could it be that we also run to 02 script when we build the image?

smoshiur1237:
> > The node image in not going to be pulled here. But it needs to pull other images such as bmo, ironic, sushy etc. 02 script is doing tagging of those image and yesterday it was failing without pulling those image from nordix registry. We cannot skip that.
> 
> Yes I'm talking about the container images. I think it may be good that we pull them so they are stored in the metal3 image. This change would make it so that they are not stored in the metal3 image I think. Or could it be that we also run to 02 script when we build the image?

Will do the change as separate script for pulling all the images and call them in both 01 and 02 script. So in-case of mismatch of the version, it will pull the images.

tests/scripts/image_prepull.sh
```diff
@@ -0,0 +1,45 @@
...
+      IMAGE_BASE_NAME="${IMAGE_NAME%.*}"
+      export IMAGE_RAW_NAME="${IMAGE_BASE_NAME}-raw.img"
+    fi
+    if [[ "${IMAGE_SUFFIX}" = "bz2" ]] ; then
+        bunzip2 "${IMAGE_NAME}"
+        IMAGE_NAME="$(basename "${IMAGE_NAME}" .bz2)"
+        export IMAGE_NAME
+        IMAGE_BASE_NAME="${IMAGE_NAME%.*}"
+        export IMAGE_RAW_NAME="${IMAGE_BASE_NAME}-raw.img"
+    fi
```
lentzi90, line 27: Ideally I think we should move this to `lib/images.sh`. Then we can execute this script instead of sourcing and it is clearer separation between setting variables and downloading.

Makefile
```diff
@@ -1,5 +1,7 @@
 all: install_requirements configure_host launch_mgmt_cluster verify
 
+ci_config: configure_host launch_mgmt_cluster verify
```
lentzi90, line 3: I'm thinking maybe `run` or something similar would be easier to understand?

lentzi90 (commented):
Looks good so far! I added a couple of suggestions below.

tests/scripts/image_prepull.sh
Rozzii: This is a good idea, but I guess this might be a bit bigger work, the exported variables are probably reused in other places.
But I like the general idea to turn the images.sh into a regular script instead of this "lib" approach.

Makefile
Rozzii: or just "ci", If I phrase it in a sentence it makes sense "The job uses the CI Make target" .

tests/scripts/image_prepull.sh
Rozzii: If we turn the image.sh into a standalone thing then might as well we put there all the node image and container image building logic together.

smoshiur1237: I think sourcing is better option as it will run within the same process. If we just execute the file, it will run as separate process.

lentzi90: Ok let's do it as a followup! I can create an issue for it when this is merged

Makefile
smoshiur1237: I just make it as ci_run

tests/scripts/image_prepull.sh
smoshiur1237: We can do improvement or refactoring with separate PR.

Makefile
```diff
@@ -59,4 +61,4 @@ healthcheck_test:
 
 feature_tests: setup_env inspection_test remediation_test healthcheck_test cleanup_env pivoting_test node_reuse_test repivoting_test
 
-.PHONY: all install_requirements configure_host launch_mgmt_cluster clean delete_mgmt_cluster host_cleanup verify test lint
+.PHONY: all ci_config install_requirements configure_host launch_mgmt_cluster clean delete_mgmt_cluster host_cleanup verify test lint
```
lentzi90, line 64:
```suggestion
.PHONY: all ci_run install_requirements configure_host launch_mgmt_cluster clean delete_mgmt_cluster host_cleanup verify test lint
```

smoshiur1237: Just noticed it and changed

lentzi90 (commented):
/approve

smoshiur1237:
test-centos-integration-release-1-4

01_prepare_host.sh
```diff
@@ -190,116 +188,6 @@ podman)
...
-          --type network --config
-    fi
-}
-
-if [[ "${EPHEMERAL_CLUSTER}" = "minikube" ]]; then
-  init_minikube
-fi
+# pre-pull node and container images
+# shellcheck disable=SC1091
+source lib/image_prepull.sh
```
Rozzii, line 193: now that I think about it,  you can remove this as 02 will handle it anyways. It is fine in a separate PR.

Rozzii (commented):
On nit, it is fine if you fix this in the next PR. ( please ignore)

PR #1229, opened 2023-05-10
Fix CAPM3RELEASE to v1.4.99 for main branch

kashifest:
We need to fix this to a non-existent CAPM3 release version i.e. v1.4.99 in this case to make sure the local override created for main branch is not conflicting with an existing github release tag.


Sunnatillo (commented):
LGTM

lentzi90 (commented):
/approve

PR #1230, opened 2023-05-11
Changing defaults from md5 to sha256

adilGhaffarDev:
Changing defaults from md5 to sha256 to fix capm3 pr: https://github.com/metal3-io/cluster-api-provider-metal3/pull/975


kashifest (dismissed):

adilGhaffarDev:
/hold 
md5sum file is in the host image and it is not being replaced by sha256sum. That needs to be fixed in this PR first.

PR #1236, opened 2023-05-24
pin downloaded scripts and binaries to a sha digest

tuminoid:
For downloaded install scripts and readily-built binaries, we need to add a custom function to download them and verify the SHA in one go. This also means that when doing version update, we need to also change the SHA. It needs to be investigated if the renovate bot can handle that or if we should be disabling renovatebot for these binaries.

This PR adds versioning to many of the binaries, instead of using latest. Using latest is often a wrong choice anyways, for example kubectl has +/- 1 version skew allowed, and using the latest is having 2 versions skew.
We also add override variable `INSECURE_SKIP_DOWNLOAD_VERIFICATION` for testing purposes, which skips the verification of the digests, so dependency bumps can be tested without worrying about the digests.

This PR also has couple TODO:s for future:
- clusterctl is following CAPIRELEASE variable which is downloaded from Github, ie. it is not possible to pin it
- kubectl version skew fix
- ansible pip install

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -5,6 +5,15 @@ set -eux
 source lib/logging.sh
 # shellcheck disable=SC1091
 source lib/common.sh
+# shellcheck disable=SC1091
+source lib/releases.sh
+# shellcheck disable=SC1091
+source lib/download.sh
+# NOTE(fmuyassarov) Make sure to source before runnig install-package-playbook.yml
+# because there are some vars exported in network.sh and used by
+# install-package-playbook.yml.
+# shellcheck disable=SC1091
+source lib/network.sh
 
 if [[ "$(id -u)" -eq 0 ]]; then
   echo "Please run 'make' as a non-root user"
@@ -58,13 +67,9 @@ elif [[ "${OS}" = "centos" ]] || [[ "${OS}" = "rhel" ]]; then
   sudo ln -s /usr/bin/python3 /usr/bin/python || true
 fi
 
+# TODO: since ansible 8.0.0, pinning by digest is PITA, due additional ansible
+# dependencies, which would need to be pinned as well, so it is skipped for now
 sudo python -m pip install ansible=="${ANSIBLE_VERSION}"
-
-# NOTE(fmuyassarov) Make sure to source before runnig install-package-playbook.yml
-# because there are some vars exported in network.sh and used by
-# install-package-playbook.yml.
-# shellcheck disable=SC1091
-source lib/network.sh
 
 # Install requirements
 ansible-galaxy install -r vm-setup/requirements.yml
@@ -84,24 +89,10 @@ if [[ ! "${PATH}" =~ .*(:|^)(${GOBINARY})(:|$).* ]]; then
   export PATH=${PATH}:${GOBINARY}
 fi
 
-# shellcheck disable=SC1091
-source lib/releases.sh
 
 ## Install krew
 if ! kubectl krew > /dev/null 2>&1; then
-  pushd "$(mktemp -d)" &&
-  KERNEL_OS="$(uname | tr '[:upper:]' '[:lower:]')" &&
-  ARCH="$(uname -m | sed -e 's/x86_64/amd64/' -e 's/\(arm\)\(64\)\?.*/\1\2/' -e 's/aarch64$/arm64/')" &&
-  KREW="krew-${KERNEL_OS}_${ARCH}" &&
-  curl -fsSLO "https://github.com/kubernetes-sigs/krew/releases/latest/download/${KREW}.tar.gz" &&
-  tar zxvf "${KREW}.tar.gz" &&
-  rm -f "${KREW}.tar.gz" &&
-  ./"${KREW}" install krew
-  # Add krew to PATH by appending this line to .bashrc
-  krew_path_bashrc="export PATH=${KREW_ROOT:-$HOME/.krew}/bin:$PATH"
-  # Add the line if it is not already there
-  grep -qxF "${krew_path_bashrc}" ~/.bashrc || echo "${krew_path_bashrc}" >> ~/.bashrc
-  popd
+  download_and_install_krew
 fi
 
 # Allow local non-root-user access to libvirt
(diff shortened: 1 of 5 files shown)


tuminoid:
@lentzi90 @kashifest any objections doing it like this?

lib/download.sh
```diff
@@ -0,0 +1,128 @@
...
+    KIND_BINARY="kind"
+
+    wget_and_verify "${KIND_URL}" "${KIND_SHA256}" "${KIND_BINARY}"
+    chmod +x "${KIND_BINARY}"
+    sudo mv "${KIND_BINARY}" /usr/local/bin/
+}
+
+download_and_install_tilt()
+{
+    # TODO: see if there is actual releases instead of just git sha
```
lentzi90, line 86: Yes there are releases and it is possible to get the binaries directly for them. I do it like this in ansible: https://github.com/lentzi90/workstation/blob/8c1dc0e209311a8380683e652b46e2f42b132015/roles/workstation/tasks/tilt.yml#L24-L30

lib/download.sh
```diff
@@ -0,0 +1,128 @@
...
+    TILT_URL="https://raw.githubusercontent.com/tilt-dev/tilt/${TILT_VERSION}/scripts/install.sh"
+    TILT_SCRIPT="install.sh"
+
+    wget_and_verify "${TILT_URL}" "${TILT_SHA256}" "${TILT_SCRIPT}"
+    bash "${TILT_SCRIPT}"
+    rm "${TILT_SCRIPT}"
+}
+
+# FIXME: kubectl cannot be latest, use kubernetes version - 1
```
lentzi90, line 95: Was this already fixed or am I missing something? :thinking:

lentzi90 (commented):
This looks like a good approach to me :+1:

lib/download.sh
tuminoid: It was more of a note to myself, and yes, I've got the release variables defined now.

lib/download.sh
tuminoid: It is now taking same k8s version (1.27.1), but that is issue for upgrade etc, where the source is 1.25, which is beyond kubectl allowed version skew. So we should actually be installing kubectl 1.26 (hence "version - 1").

lentzi90: Ah got it. :+1:

tuminoid: I'll actually just leave it as TODO in the script, and fix is separately.

lentzi90 (commented):
I have been thinking about the different use cases on metal3-dev-env and have one concern about this. The versions are easily overridable, which is useful both for specific manual tests and also for multi-version CI, but the hashes are not.

Do you think it would make sense to allow the hashes to be unset for these cases? Or perhaps allow them to be overridden? Overriding should be fine for CI I think but it could quickly become very annoying for manual tests.
Yet another option would be to implement automatic fetching of the checksums, but I imagine that could be quite some work...

tuminoid:
> I have been thinking about the different use cases on metal3-dev-env and have one concern about this. The versions are easily overridable, which is useful both for specific manual tests and also for multi-version CI, but the hashes are not.
> 
> Do you think it would make sense to allow the hashes to be unset for these cases? Or perhaps allow them to be overridden? Overriding should be fine for CI I think but it could quickly become very annoying for manual tests. Yet another option would be to implement automatic fetching of the checksums, but I imagine that could be quite some work...

That is a fair point. Since the verification is consolidated to two functions, I could a check for something like `INSECURE_IGNORE_DIGESTS`, and it'll just print warnings instead of failing.

lentzi90:
Good idea!

tuminoid:
> Good idea!

Done. It is a bit lengthy variable name, but wanted to be explicit what it does.

lentzi90 (commented):
Looks good to me!
I would like to get input from @kashifest and @mboukhalfa before approving though. Do you think this cause issues with for example e2e or release branch tests?

tuminoid:
> Looks good to me! I would like to get input from @kashifest and @mboukhalfa before approving though. Do you think this cause issues with for example e2e or release branch tests?

It should not, since it just pins down the versions we are already using. The problematic ones like kubectl with version skew and clusterctl which is tied to downloaded CAPIRELEASE versions are not verified. Fixing those is left to some other PR.

We can and should wait for more comments to make sure nothing is missed.

01_prepare_host.sh
```diff
@@ -164,14 +137,16 @@ remove_ironic_containers
...
-  ;;
+  podman)
+    for pod in ironic-pod infra-pod; do
+      if  sudo "${CONTAINER_RUNTIME}" pod exists "${pod}" ; then
+          sudo "${CONTAINER_RUNTIME}" pod rm "${pod}" -f
+      fi
+      sudo "${CONTAINER_RUNTIME}" pod create -n "${pod}"
+    done
+    ;;
+  *)
```
kashifest, line 148: what is this line doing?

kashifest:
/approve
One question inline.
/test-ubuntu-e2e-feature-main

Rozzii:
This mostly affects 01 script, I think we don't run 01 in integration tests anymore.

Rozzii:
I have checked now, there is not PR test for image building as far as I can tell so I remove the hold, let's hope it won't break anything in image building.
/cancel hold

01_prepare_host.sh
tuminoid: It is fallback option for `case`, mandatory per shellcheck even if its just a fallthru like this one. We should error out if someone specifies `CONTAINER_RUNTIME` we don't support, but that is for another PR.

kashifest:
/hold
Some more tests are planned for next week.

tuminoid:
/test-centos-integration-release-1-4

This time including a test commit that removes all pre-installed binaries.

tuminoid:
> /test-ubuntu-e2e-integration-main /test-centos-integration-release-1-4

OK, now with test commit it can be seen that it works, removing the test commit again.

lentzi90 (commented):
/lgtm

kashifest:
> > /test-ubuntu-e2e-integration-main /test-centos-integration-release-1-4
> 
> OK, now with test commit it can be seen that it works, removing the test commit again.

Can you share the test run links with test commit here ?

tuminoid:
> Can you share the test run links with test commit here ?

How do you get the link after the fact?

kashifest:
> > Can you share the test run links with test commit here ?
> 
> How do you get the link after the fact?

Jenkins run?

lentzi90:
You can check the history in jenkins and find previous runs for the same PR: https://jenkins.nordix.org/job/metal3_metal3_dev_env_release-1-4_integration_test_centos/
I'm not sure which one is with the test commit here though

PR #1244, opened 2023-06-08
Fix running consecutive make test

mboukhalfa:
Currently when running `make test` 2 times consecutive the second run will fail because integration test expect BMH crs to pre-exist but as the first test is deleting them at the end the second will fail

This PR fix this issue by making integration test cleaning avoid deleting the bmh crs


tuminoid:
I'm running some tests on this.

lentzi90 (commented):
/approve
Hold until manual tests are done
/hold

tuminoid (commented):
/lgtm

Test passed twice in a row with no additional steps in between.

PR #1247, opened 2023-06-14
Fix package installation for ubuntu 20

Sunnatillo:
Fixes package installation for ubuntu 20 by creating missing folder /etc/apt/keyrings for gpg key storage

### DIFF
--- a/vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
+++ b/vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
@@ -55,6 +55,11 @@
 
     - name: Install docker
       block:
+        - name: Create /etc/apt/keyrings folder
+          file:
+            path: /etc/apt/keyrings
+            state: directory
+
         - name: Add Docker’s GPG key
           get_url:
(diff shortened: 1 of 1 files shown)


tuminoid:
For Ubuntu 20.04, `mkdir /etc/apt/keyrings` was sufficent fix, I'll test Ubuntu 18.04 as well. That'd be way simpler solution than this, and more futureproof as well.

/hold

tuminoid:
> For Ubuntu 20.04, `mkdir /etc/apt/keyrings` was sufficent fix, I'll test Ubuntu 18.04 as well. That'd be way simpler solution than this, and more futureproof as well.

18.04 can't handle `apt-get upgrade` at the start of 01 script, so it never even gets to the point where Docker keys are added. I feel we're wasting efforts supporting 18.04 at this point.

tuminoid (approved):
I tested this simple version on 22.04 and 20.04 and it works.

tuminoid (commented):
/lgtm

lentzi90 (commented):
/approve

PR #1251, opened 2023-06-20
move sourcing of libs after initial package install

tuminoid:
Many of the `lib/*.sh` actually run code requiring additional command line tooling, which is only installed after sourcing them. Running `make` on the vanilla Ubuntu fails currently. For this reason, we want to source `lib/releases.sh` and `lib/network.sh` after installing `jq` and `python` alternative, and `lib/download.sh` which needs `lib/common.sh`.

This is kind of first-aid to fix the issue. The real solution would be for the libs not to run any code when sourced, only create functions for later use.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -5,15 +5,6 @@ set -eux
 source lib/logging.sh
 # shellcheck disable=SC1091
 source lib/common.sh
-# shellcheck disable=SC1091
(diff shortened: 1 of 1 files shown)


Sunnatillo (commented):
LGTM

lentzi90 (commented):
/approve

PR #1253, opened 2023-07-05
Allow proxy connection when dividing response

fenggw-fnst:
If a proxy is used when trying to connect to the server, a response message with the content "HTTP/1.1 200 Connection established" will be returned in the first line, followed by the required header information in the second line, which will result in incorrect headers and body being obtained. This patch aims to address this issue.

### DIFF
--- a/lib/releases.sh
+++ b/lib/releases.sh
@@ -21,7 +21,12 @@ function get_latest_release() {
 
   # Divide response to headers and body
   response_headers=$(echo "${response}" | awk 'BEGIN {RS="\r\n\r\n"} NR==1 {print}')
-  response_body=$(echo "${response}" | awk 'BEGIN {RS="\r\n\r\n"} NR==2 {print}')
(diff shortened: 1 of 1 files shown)


lib/releases.sh
```diff
@@ -21,7 +21,12 @@ function get_latest_release() {
 
   # Divide response to headers and body
   response_headers=$(echo "${response}" | awk 'BEGIN {RS="\r\n\r\n"} NR==1 {print}')
-  response_body=$(echo "${response}" | awk 'BEGIN {RS="\r\n\r\n"} NR==2 {print}')
+  if echo "${response_headers}" | grep '^link:'; then
```
elfosardo, line 24: you probably want 'grep -q' here

fenggw-fnst: Thanks for the review, done.

Sunnatillo (commented):
/lgtm

furkatgofurov7 (commented):
LGTM

smoshiur1237 (approved):

PR #1255, opened 2023-07-18
Use MariaDB instead of SQLite

smoshiur1237:
We are having CI issue where ironic database was locked. Community suggested to use MariaDB instead of SQLite for ironic.


furkatgofurov7 (approved):
/approve

@smoshiur1237 feel free to unhold when ready

PR #1256, opened 2023-07-19
Fix use of mariadb install option for CAPM3 branch

smoshiur1237:
After enabling the use of mariadb variable, we are seeing issue with release 1.2 and 1.3 branch of CAPM3. Which doesn't have any use of the install mariadb options. So excluding the -m option for those CAPM3 branches.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -24,7 +24,9 @@ if [ "${IRONIC_BASIC_AUTH}" == "false" ]; then
   BMO_IRONIC_ARGS+=("-n")
 fi
 if [ "${IRONIC_USE_MARIADB:-true}" == "true" ]; then
-  BMO_IRONIC_ARGS+=("-m")
+  if [[ "${CAPM3BRANCH}" != "release-1.3" ]] && [[ "${CAPM3BRANCH}" != "release-1.2" ]]; then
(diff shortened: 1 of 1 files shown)


Rozzii:
Is this still relevant @smoshiur1237  ?

lentzi90:
I don't like this way of working around it. If `IRONIC_USE_MARIADB=true` then it should always use mariadb. If it fails because the release branch cannot work with mariadb then so be it. Overriding the variable based on the branch is too surprising in my opinion.
If we want an early exit that is fine, just print an error and exit, but don't run with sqlite when `IRONIC_USE_MARIADB=true`.

tuminoid:
> I don't like this way of working around it. If `IRONIC_USE_MARIADB=true` then it should always use mariadb. If it fails because the release branch cannot work with mariadb then so be it. Overriding the variable based on the branch is too surprising in my opinion. If we want an early exit that is fine, just print an error and exit, but don't run with sqlite when `IRONIC_USE_MARIADB=true`.

I agree with Lennart here. Ignoring user config is recipe for disaster later on. If we know it will not work, then we will detect that and print an error, and exit. This leaves no ambiquity to user what went wrong.

smoshiur1237:
I do agree with @lentzi90 and also as 1.2 and 1.3 is EOL. I can close this PR and if someone wants to take it forward, I am ok with that.

PR #1259, opened 2023-07-26
Allow proxy for downloading IPA

fenggw-fnst:
In environments that require a proxy, the script fails during the IPA download step. This patch addresses this issue by adding support for proxy.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -21,7 +21,7 @@ if [[ "${IPA_DOWNLOAD_ENABLED}" = "true" ]] || [[ ! -r "${IRONIC_DATA_DIR}/html/
         sudo "${CONTAINER_RUNTIME}" run --rm --net host --name ipa-downloader ${POD_NAME} \
           -e "IPA_BASEURI=${IPA_BASEURI:-}" \
           -v "${IRONIC_DATA_DIR}:/shared" "${IPA_DOWNLOADER_IMAGE}" \
-          /bin/bash -c "/usr/local/bin/get-resource.sh &> /dev/null" && s=0 && break || s=$?
+          /bin/bash -c "http_proxy=${http_proxy:-} https_proxy=${https_proxy:-} /usr/local/bin/get-resource.sh &> /dev/null" && s=0 && break || s=$?
     done
     (exit "${s}")
 fi


02_configure_host.sh
```diff
@@ -21,7 +21,7 @@ if [[ "${IPA_DOWNLOAD_ENABLED}" = "true" ]] || [[ ! -r "${IRONIC_DATA_DIR}/html/
         sudo "${CONTAINER_RUNTIME}" run --rm --net host --name ipa-downloader ${POD_NAME} \
           -e "IPA_BASEURI=${IPA_BASEURI:-}" \
           -v "${IRONIC_DATA_DIR}:/shared" "${IPA_DOWNLOADER_IMAGE}" \
-          /bin/bash -c "/usr/local/bin/get-resource.sh &> /dev/null" && s=0 && break || s=$?
+          /bin/bash -c "http_proxy=${http_proxy} https_proxy=${https_proxy} /usr/local/bin/get-resource.sh &> /dev/null" && s=0 && break || s=$?
```
elfosardo, line 24: you should probably assign a default for http_proxy and https_proxy to avoid  undefined variable error

elfosardo (changes requested):

02_configure_host.sh
fenggw-fnst: Thanks for the review, done.

Sunnatillo (commented):
/lgtm

Sunnatillo (approved):
/lgtm

02_configure_host.sh
```diff
@@ -21,7 +21,7 @@ if [[ "${IPA_DOWNLOAD_ENABLED}" = "true" ]] || [[ ! -r "${IRONIC_DATA_DIR}/html/
         sudo "${CONTAINER_RUNTIME}" run --rm --net host --name ipa-downloader ${POD_NAME} \
           -e "IPA_BASEURI=${IPA_BASEURI:-}" \
           -v "${IRONIC_DATA_DIR}:/shared" "${IPA_DOWNLOADER_IMAGE}" \
-          /bin/bash -c "/usr/local/bin/get-resource.sh &> /dev/null" && s=0 && break || s=$?
+          /bin/bash -c "http_proxy=${http_proxy:-} https_proxy=${https_proxy:-} /usr/local/bin/get-resource.sh &> /dev/null" && s=0 && break || s=$?
```
tuminoid, line 24: In case the variables contain space, this will break the syntax. Add quotes around the variables, and mind that the command is already once quoted.

tuminoid (changes requested):

02_configure_host.sh
dtantsur: What does this command even do? It seems to just set the variables to themselves. What I think is needed instead, is something like `-e HTTP_PROXY` passed to podman/docker (and needs handling HTTP_PROXY vs http_proxy).

fenggw-fnst:
IPA download is no longer needed (see https://github.com/metal3-io/metal3-dev-env/pull/1272), so this PR can be closed.

PR #1261, opened 2023-07-27
Fix incorrect EFI firmware reference in VM template

Rozzii:
Until this change a secure boot enabled EFI firmware was used even when secure boot was not requested. The old reference caused boot failures whit non secure bootable EFI compatoble boot media.

NOTE: there is a separate option already to enable secure EFI boot in the dev-env.


kashifest (approved):

smoshiur1237 (commented):
/lgtm

Issue #1264, opened 2023-07-31
minikube delete hangs after bumping to v1.31.1


kashifest:
It seems like after bumping minikube to [v1.31.1](https://github.com/metal3-io/metal3-dev-env/pull/1262) tests are stuck while deleting minikube. 

One example run is here:

https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3_daily_release-1-4_e2e_integration_test_centos/104/console

kashifest:
It seems even after the WA in https://github.com/metal3-io/metal3-dev-env/pull/1265 , there are cases where libvirt daemon is becoming unresponsive https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3_daily_main_e2e_feature_test_centos/192/console . This issue can be marked as a flake. There are open issues linking this to libvirt issue: https://bugzilla.redhat.com/show_bug.cgi?id=2213660 

/kind flake

mboukhalfa:
@kashifest it this issue assigned or fixed by someone already ?

kashifest:
This can be closed.

PR #1265, opened 2023-07-31
Fix minikube delete hanging issue

kashifest:
Restart libvirtd helps in local test to allow minikube delete to proceed. This is a workaround until real issue is found. Its a workaround for #1264


Sunnatillo (approved):
/lgtm

PR #1267, opened 2023-08-01
Download and install kvm2 driver with minikube

kashifest:
We should download the kvm2 driver whenever minikube is downloaded since minikube is dependant on the driver itself and current check would not download newer version of the driver in case newer minikube is downloaded in CI images. Keeping the old check in place though since that might be needed for vanilla images.


mboukhalfa:

Please add a brief description summarizing the problem being addressed. This will help reviewers and future contributors understand the changes better. If the discussion from the morning meeting can be summarized, that would be sufficient.

kashifest:
> > _ No description provided. _
> 
> Please add a brief description summarizing the problem being addressed. This will help reviewers and future contributors understand the changes better. If the discussion from the morning meeting can be summarized, that would be sufficient.

I am about to, checking the tests first, give me some time :)

tuminoid (commented):
Good finding!

/lgtm

PR #1160, opened 2023-02-28
Allow metal3-dev-env to run on a generic ubuntu VM

elfosardo:
Currently metal3-dev-env is configured to run on a preconfigured ubuntu host.
Some configuration parameters and options are given as assumed, but to run it
in a generic VM ( based on ubuntu focal or jammy), we need to update the
configuration to take into consideration a default clean OS configuration.

PR #1271, opened 2023-08-07
add python3-dev for libvirt-python building

tuminoid:
Vanilla Ubuntu image is missing python3-dev required to build and install libvirt-python. 01 script is failing on vanilla because of this.
Starting libvirt-python 9.5.0, released in July 2023, they state they need python development headers, which wasn't there in 9.4.0.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -16,7 +16,7 @@ if [[ "${OS}" = "ubuntu" ]]; then
   # make the data retrival more reliable
   sudo sh -c ' echo "Acquire::Retries \"10\";" > /etc/apt/apt.conf.d/80-retries '
   sudo apt-get update
-  sudo apt-get -y install python3-pip jq curl wget pkg-config bash-completion
+  sudo apt-get -y install python3-pip python3-dev jq curl wget pkg-config bash-completion
 
   # Set update-alternatives to python3
   if [[ "${DISTRO}" = "ubuntu18" ]]; then


kashifest:
so this https://github.com/metal3-io/metal3-dev-env/pull/1160 was not enough then? What if we add the package [here](https://github.com/metal3-io/metal3-dev-env/blob/main/vm-setup/roles/packages_installation/defaults/main.yml) or is it too late and the failure occurs earlier? This playbook is also called in 01 script.

Sunnatillo (commented):
/lgtm

tuminoid:
> so this #1160 was not enough then? What if we add the package [here](https://github.com/metal3-io/metal3-dev-env/blob/main/vm-setup/roles/packages_installation/defaults/main.yml) or is it too late and the failure occurs earlier? This playbook is also called in 01 script.

It might be that libvirt-python has its dependencies changed. We are not using versioned dependencies here yet it seems. Starting libvirt-python 9.5.0, released in July, they state they need python development headers, which wasn't there in 9.4.0.

tuminoid:
> so this #1160 was not enough then? What if we add the package [here](https://github.com/metal3-io/metal3-dev-env/blob/main/vm-setup/roles/packages_installation/defaults/main.yml) or is it too late and the failure occurs earlier? This playbook is also called in 01 script.

Is it guaranteed by Ansible that `pip3` section is executed only after all of the `ubuntu` packages are installed? If so, we can put it there as well.

PR #1280, opened 2023-08-29
fix KUBECTL_SHA256 to match KUBERNETES_VERSION

tuminoid:
When KUBERNETES_VERSION is changed, it changes the KUBECTL_SHA256 as well as the versions are coupled. There is TODO item to get rid of this coupling.

The issue is not visible until CI images are rebuilt as 01_script does not get executed during regular integration tests. Anyone using vanilla image will get the error though, like Ironic CI.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -315,9 +315,9 @@ else
   export EPHEMERAL_CLUSTER="minikube"
 fi
 
-# Kubectl version
+# Kubectl version (do not forget to update KUBECTL_SHA256 when changing KUBERNETES_VERSION!)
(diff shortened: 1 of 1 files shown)


tuminoid:
/override test-centos-integration-release-1-4 
/override test-ubuntu-e2e-integration-main 

01 script is skipped in integration tests, so it is meaningless to test this. Overriding to get it in faster to unblock Ironic CI.

kashifest (approved):

kashifest:
lets change the name of the var to reflect the k8s version it corresponds to in a followup PR

PR #1282, opened 2023-08-29
Add version check to kubectl installation

kashifest:
This PR fixes a bug in kubectl installation. Currently kubectl installation doesn't check the version and as a result even if the it is bumped and if in CI image the old binary exists, new installation will not happen. This PR adds the version check.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -124,7 +124,7 @@ else
   fi
 fi
 
-if ! command -v kubectl &>/dev/null; then
+if ! command -v kubectl &>/dev/null || [[ "$(kubectl version --client -o json|jq -r '.clientVersion.gitVersion')" != "${KUBECTL_VERSION}" ]]; then
   download_and_install_kubectl
 fi
 


lib/download.sh
```diff
@@ -126,6 +126,10 @@ download_and_install_kubectl()
     KUBECTL_PATH=$(whereis -b kubectl | cut -d ":" -f2 | awk '{print $1}')
     KUBECTL_PATH="${KUBECTL_PATH:-/usr/local/bin/kubectl}"
     KUBECTL_URL="https://storage.googleapis.com/kubernetes-release/release/${KUBECTL_VERSION}/bin/linux/amd64/kubectl"
+    
+    # Download the kubectl checksum file
+    curl -LO "https://dl.k8s.io/${KUBECTL_VERSION}/bin/linux/amd64/kubectl.sha256"
+    KUBECTL_SHA256=$(cat kubectl.sha256)
```
kashifest, line 132: @tuminoid is downloading sha256 from source ok according to security best practices or pinning is preferable?

tuminoid: Pinning is about preventing unwanted change. If someone replaces binary and checksum at upstream or in transit, we would never know. Pinning makes all changes go via PR, which is what we want.

Do not remove pinning.

tuminoid (changes requested):
Do not remove pinning.

lib/download.sh
kashifest: Well, for dev environment, pinning is not absolutely necessary IMO,  specially since sha256 in this case depends on the architecture the binary was built for. I dont see CAPI or other providers pinning kubectl sha.

tuminoid: It is not "dev", it is our CI as well. Also, CAPI doing something or not doing something doesn't mean it is the only right thing to do. Each download is part of our supply chain, which is increasingly attacked over various vectors.

kashifest: well, yes this is a part or CI, but I might have different understanding when it comes to supply chain since this dev-env is not part of it, I still think this is an overkill and forcing people to always use a specific architecture. anyways, I will keep the pinning

kashifest:
> Do not remove pinning.

Done

tuminoid (approved):
/lgtm

lib/download.sh
tuminoid: We have locked in architecture with the supported OS. We support amd64 Ubuntu and Centos for 01 script and make, and those can only be run in Linux amd64 systems due virtualization dependencies etc. Hence amd64 is hardcoded for those downloads. Developer using Mac won't be using those scripts to install the tools for themselves.

lentzi90 (commented):
/approve

PR #1283, opened 2023-08-30
add a helper script to print digests of downloaded artifacts

tuminoid:
For updating pinned binary downloads, getting the correct digest has been more work than it needs to be.

Workflow:
- update the version numbers in `lib/common.sh`
- run `./hack/print_checksums.sh` 
- update the sha256 in `lib/common.sh` as needed

### DIFF
--- a/hack/print_checksums.sh
+++ b/hack/print_checksums.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+# This is a helper for verifying/printing digests of downloaded, pinned
+# dependencies.
+
+set -eu
+
+METAL3_DEV_ENV_DIR="$(dirname "$(readlink -f "${0}")")/.."
+
+# shellcheck disable=SC1091
+. "${METAL3_DEV_ENV_DIR}/lib/common.sh"
+# shellcheck disable=SC1091
+. "${METAL3_DEV_ENV_DIR}/lib/download.sh"
+
+# this downloads and prints sha256 digests for all pinned dependencies
+_download_and_print_checksums
--- a/lib/download.sh
+++ b/lib/download.sh
@@ -1,14 +1,16 @@
 #!/usr/bin/env bash
 #
-# utils to download and verify binary downloads in shell scripts
+# Utils to download and verify binary downloads in shell scripts
 #
-# this expects lib/common.sh to be sourced
+# This expects lib/common.sh to be sourced where the variables are defined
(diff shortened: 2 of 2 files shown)


tuminoid:
/cc @kashifest 

Would this be good enough, or have ideas to make it even simpler?

tuminoid:
/hold
For opinions and improvements.

lentzi90 (commented):
/approve

The print for krew was a bit cryptic compared to the rest, but I don't think it is an issue.
```
info: sha256(/tmp/tmp.Zld3ATZ9Uz/krew-linux_amd64.tar.gz): 5df32eaa0e888a2566439c4ccb2ef3a3e6e89522f2f2126030171e2585585e4f
```

tuminoid:
> /approve
> 
> The print for krew was a bit cryptic compared to the rest, but I don't think it is an issue.
> 
> ```
> info: sha256(/tmp/tmp.Zld3ATZ9Uz/krew-linux_amd64.tar.gz): 5df32eaa0e888a2566439c4ccb2ef3a3e6e89522f2f2126030171e2585585e4f
> ```

True... I'll cut the leading directories off from the print.

kashifest:
lgtm

tuminoid:
OK, I changed the printout not to print the temporary path, and also it deletes the artifacts instead of leaving them in the working dir. I also added a `hack/print_checksums.sh` so you don't need to source stuff.

Unfortunately our dev-env `lib/common.sh` has idiotic passwordless sudo check, so unless you've configured passwordless sudo just the exact way it wants you to, you cannot run the script. Need to fix that in another PR.

lentzi90:
> Unfortunately our dev-env lib/common.sh has idiotic passwordless sudo check, so unless you've configured passwordless sudo just the exact way it wants you to, you cannot run the script. Need to fix that in another PR.

True, I'm always fighting this... :disappointed: I don't want passwordless sudo on my laptop.

tuminoid:
> > Unfortunately our dev-env lib/common.sh has idiotic passwordless sudo check, so unless you've configured passwordless sudo just the exact way it wants you to, you cannot run the script. Need to fix that in another PR.
> 
> True, I'm always fighting this... 😞 I don't want passwordless sudo on my laptop.

Created an issue to fix it: https://github.com/metal3-io/metal3-dev-env/issues/1288

PR #1289, opened 2023-09-01
make sha256 digests overridable for upgrade tests

tuminoid:
Make *_SHA256 variables take input.

In upgrade tests, we do not install default k8s version, but something older. Since kubectl is locked with the k8s version, it means it now installs older kubectl which then doesn't match the digest. This is now revealed as kubectl version check was recently fixed.

This allows us to input the SHA and VERSION as a pair from JJB, which fixes the upgrade tests.


lentzi90 (commented):
/approve

Sunnatillo (approved):
/lgtm

PR #1292, opened 2023-09-04
Bump minikube to v1.31.2

kashifest:
As the title says. Also trying to test if it solves the issue reported in  https://github.com/metal3-io/metal3-dev-env/issues/1264 by removing the extra libvirtd restart. Upstream issue https://github.com/kubernetes/minikube/issues/16975


Sunnatillo (approved):
/lgtm

lentzi90 (commented):
/approve

Do we need to run any extra tests to verify that it solves the hanging issue?
/hold

Sunnatillo:
> /approve
> 
> Do we need to run any extra tests to verify that it solves the hanging issue? /hold

as I remember it failed every time without restarting libvirt line. So it should be fine now.

kashifest:
> /approve
> 
> Do we need to run any extra tests to verify that it solves the hanging issue? /hold

I ran all the centos integration tests. Seems not to occur. Do you suggest any other tests ?

lentzi90:
> > /approve
> > Do we need to run any extra tests to verify that it solves the hanging issue? /hold
> 
> if I remember correctly, it failed every time without restarting libvirt. So it should be fine now.

Good, then we should be fine :slightly_smiling_face: 

/hold cancel

tuminoid (commented):
LGTM

Did you use the new helper script for the digests?

PR #1304, opened 2023-10-04
Add BMORELEASEBRANCH variable

kashifest:
This PR adds BMORELEASEBRANCH variable which gets the value from JJB. This PR also removes v01.X BMO version checks from code which is not tested/supported anymore. Also bumping bmo to v0.3.1 for release-1.4 test

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -177,11 +177,7 @@ EOF
   fi
 
   # Copy the generated configmap for ironic deployment
-  if [[ ${BMOBRANCH} == "v0.1.2" ]] || [[ ${BMOBRANCH} == "v0.1.1" ]]; then # BMOBRANCH until v0.1.2 used the old path TODO(mboukhalfa) can be removed once old releases are obsolete
-    cp "${IRONIC_DATA_DIR}/ironic_bmo_configmap.env" "${BMOPATH}/ironic-deployment/keepalived/ironic_bmo_configmap.env"
-  else
-    cp "${IRONIC_DATA_DIR}/ironic_bmo_configmap.env"  "${BMOPATH}/ironic-deployment/components/keepalived/ironic_bmo_configmap.env"
-  fi
+  cp "${IRONIC_DATA_DIR}/ironic_bmo_configmap.env"  "${BMOPATH}/ironic-deployment/components/keepalived/ironic_bmo_configmap.env"
 
   # Update manifests to use the correct images.
   # Note: Even though the manifests are not used for local deployment we need
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -144,6 +144,7 @@ export CAPM3PATH="${CAPM3PATH:-${M3PATH}/cluster-api-provider-metal3}"
 export CAPM3_BASE_URL="${CAPM3_BASE_URL:-metal3-io/cluster-api-provider-metal3}"
 export CAPM3REPO="${CAPM3REPO:-https://github.com/${CAPM3_BASE_URL}}"
 export CAPM3RELEASEBRANCH="${CAPM3RELEASEBRANCH:-main}"
+export BMORELEASEBRANCH="${BMORELEASEBRANCH:-main}"
 
 if [[ "${CAPM3RELEASEBRANCH}" == "release-1.3" ]]; then
   export CAPM3BRANCH="${CAPM3BRANCH:-release-1.3}"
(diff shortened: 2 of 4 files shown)


kashifest:
/hold
Until https://gerrit.nordix.org/c/infra/cicd/+/19538 and https://github.com/metal3-io/project-infra/pull/585 lands

lib/common.sh
```diff
@@ -268,23 +269,23 @@ if [[ "${CAPM3RELEASEBRANCH}" = "release-1.3" ]]; then
...
   export KEEPALIVED_TAG="v0.3.0"
-  export BMOBRANCH="${BMOBRANCH:-v0.3.0}"
+  export BMOBRANCH="${BMORELEASEBRANCH:-v0.3.0}"
 elif [[ "${CAPM3RELEASEBRANCH}" = "release-1.5" ]]; then
   export CAPM3_IMAGE=${CAPM3_IMAGE:-"${CONTAINER_REGISTRY}/metal3-io/cluster-api-provider-metal3:release-1.5"}
   export IPAM_IMAGE=${IPAM_IMAGE:-"${CONTAINER_REGISTRY}/metal3-io/ip-address-manager:release-1.5"}
   export BARE_METAL_OPERATOR_TAG="v0.4.0"
   export KEEPALIVED_TAG="v0.4.0"
-  export BMOBRANCH="${BMOBRANCH:-v0.4.0}"
+  export BMOBRANCH="${BMORELEASEBRANCH:-release-0.4}"
```
tuminoid, line 284: We don't have that branch yet btw. 

Should we leave it out of this one, and just use `v0.4.0` like we use `v0.3.0` above, and in another PR later on explicitly move to use release branches?

Btw, why we're using `v0.3.0` when we have `v0.3.1` out?

03_launch_mgmt_cluster.sh
tuminoid, line 1: Should something be also added to `lib/releases.sh`, which mentions BMORELEASEPATH on L68, but doesn't implement it?

tuminoid (commented):
Some questions

03_launch_mgmt_cluster.sh
kashifest: We can, but since we will be anyhow switching to BMO e2e in near future, I just went to specifying the BMO version explicitly as we do now currently in lib/common.sh based on CAPM3 version.

lib/common.sh
kashifest: I will put v0.3.1. Didn't notice it. The idea is to merge this once we have the branch. Or do you suggest to merge it with versions now. The jjb PR is also sending release branch name here instead of tag btw. So I have to modify that as well if we decide to merge first and then branch out and add the branch name.

tuminoid: I figured we'd do some groundwork now, and then switch to release branches for all supported releases at once, but we can do it this way too.

kashifest: bumped BMO

kashifest:
> /test-centos-integration-release-1-5 /test-ubuntu-e2e-integration-main

We would need to merge jjb and project-infra for this patch to work.

lib/common.sh
```diff
@@ -144,6 +144,7 @@ export CAPM3PATH="${CAPM3PATH:-${M3PATH}/cluster-api-provider-metal3}"
 export CAPM3_BASE_URL="${CAPM3_BASE_URL:-metal3-io/cluster-api-provider-metal3}"
 export CAPM3REPO="${CAPM3REPO:-https://github.com/${CAPM3_BASE_URL}}"
 export CAPM3RELEASEBRANCH="${CAPM3RELEASEBRANCH:-main}"
+export BMORELEASEBRANCH="${BMORELEASEBRANCH:-main}"
```
lentzi90, line 147: I'm not sure we should set this default. Below we set `BMOBRANCH` based on this variable or a default if unset, but since we set this already here, the default will never be used. Do we need to set it here?

lib/common.sh
```diff
@@ -268,23 +269,23 @@ if [[ "${CAPM3RELEASEBRANCH}" = "release-1.3" ]]; then
   export IPAM_IMAGE=${IPAM_IMAGE:-"${CONTAINER_REGISTRY}/metal3-io/ip-address-manager:release-1.3"}
   export BARE_METAL_OPERATOR_TAG="v0.2.0"
   export KEEPALIVED_TAG="v0.2.0"
-  export BMOBRANCH="${BMOBRANCH:-v0.2.0}"
+  export BMOBRANCH="${BMORELEASEBRANCH:-v0.2.0}"
```
lentzi90, line 272: This will never take the `v0.2.0` default since we set `BMORELEASEBRANCH` already above.

lib/common.sh
kashifest: good point, probably not

lib/common.sh
kashifest: Removed the previous default

lentzi90 (commented):
/approve

vars.md
```diff
@@ -29,6 +29,7 @@ assured that they are persisted.
 | TEST_MAX_TIME | Number of maximum verification or test retries | | 120 |
 | BMO_ROLLOUT_WAIT | Number of minutes(Until max 10m that is the default value of deployment.spec.progressDeadlineSeconds) to wait for BMO rollout | | 5 |
 | BMC_DRIVER | Set the BMC driver | "ipmi", "redfish", "redfish-virtualmedia" | "mixed" |
+| BMORELEASEBRANCH | BMO Release branch | "main", "release-1.4" | Set via jjb for CI, for local dev it gets default value based on CAPM3 branch |
```
tuminoid, line 32: Surely `release-1.4` is wrong for BMO.

tuminoid (commented):
/lgtm

PR #1305, opened 2023-10-05
fix the application of the kvm_tdp_mmu workaround

Rozzii:
At the time of writing the CI is experiencing broken pipe issues for a week. The issues are present for both Centos and Ubuntu. In the previous implementation the fix was applied in 01 script but that is not executed in most of the scripts currently. This commit moves the application of the fix into 02 script and enables it for both distros again.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -13,6 +13,18 @@ source lib/releases.sh
 # shellcheck disable=SC1091
 source lib/image_prepull.sh
 
+# (workaround) disable tdp_mmu to avoid
+# kernel crashes with  NULL pointer dereference
+# note(elfosardo): run this only if we have kvm support
+if grep -q vmx /proc/cpuinfo; then
+  sudo modprobe -r -a kvm_intel kvm
+  sudo modprobe kvm tdp_mmu=0
+  sudo modprobe -a kvm kvm_intel
+elif grep -q svm /proc/cpuinfo; then
+  sudo modprobe -r -a kvm_amd kvm
+  sudo modprobe kvm tdp_mmu=0
+  sudo modprobe -a kvm kvm_amd
+fi
 # Clean, copy and extract local IPA
 if [[ "${USE_LOCAL_IPA}" = "true" ]]; then
   sudo rm -f  "${IRONIC_DATA_DIR}/html/images/ironic-python-agent*"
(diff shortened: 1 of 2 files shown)


lentzi90 (commented):
/approve

adilGhaffarDev:
I was checking logs here: https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3_daily_main_e2e_feature_test_ubuntu/265/consoleFull and I can see 01 script running. Is it somehow being reset or not working in 01?

tuminoid:
I'm just wondering how is this supposed to change anything, or why is it suspected to be culprit in the first place?

Rozzii:
> I was checking logs here: https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3_daily_main_e2e_feature_test_ubuntu/265/consoleFull and I can see 01 script running. Is it somehow being reset or not working in 01?

I have hard time following when things are being re enabled, as of my last info these shouldn't ever run 01 but in any case I would like to re-enable this for all distros and all scenarios and let's see if it helps.

Rozzii:
> I'm just wondering how is this supposed to change anything, or why is it suspected to be culprit in the first place?

The symptoms shown back when this fix was made are exactly the same what we experience right now, I would like to re-enable this workaround to see if it helps or not and rule it out.

02_configure_host.sh
```diff
@@ -13,6 +13,18 @@ source lib/releases.sh
 # shellcheck disable=SC1091
 source lib/image_prepull.sh
 
+# (workaround) disable tdp_mmu to avoid
```
tuminoid, line 16: So earlier this was under 01, so moving this to 02 shouldn't make any difference, since we backed down from skipping 01 and like @adilGhaffarDev said, 01 is run always again. But this also moves it to top level, so it is no longer run only on Ubuntu 22, but Ubuntu 18, Ubuntu 20 and Centos. 

Since Ubuntu 22 is failing as well, I'm not sure why this would help... If someone wants to LGTM it, I don't object tho.

smoshiur1237: As we are getting this issue in both ubuntu and centos, this part of the code is valid for both.

Rozzii: yeah, I have forgot we again run 01, but yeah I would like to be sure this is being applied, the error is just too similar to this old issue (doesn't makes sense to me either....)

PR #1309, opened 2023-10-25
Quick fix for fetching github releases

kashifest:
Hardcoding the releases until we figure out whats wrong with the release fetching.


lentzi90:
I have regenerated the github token. Looks like that did the trick

kashifest:
Seems this patch is not needed, the issue was in token after all, closing this
/close

PR #1314, opened 2023-11-16
Run gpasswd with sudo

elfosardo:
It won't work properly in non root environments


tuminoid (commented):
/lgtm
/assign @smoshiur1237

smoshiur1237 (approved):
/approve

PR #1317, opened 2023-11-21
Correct command

cjeanner:
gpasswd does the same as the usermod call we have just before.

The actual command we want here is the `newgrp` since it will reload all the groups, like a new login.


tuminoid:
/ok-to-test

We know from previous PR that this isn’t actually verified in CI since the user is already in the group. I’ll verify this manually.

tuminoid (commented):
Vanilla Ubuntu VM -> run 01 script.

```
ubuntu@tuomo-ubuntu:~$ id
uid=1000(ubuntu) gid=1000(ubuntu) groups=1000(ubuntu),4(adm),20(dialout),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),119(netdev),120(lxd)
# no libvirt 
ubuntu@tuomo-ubuntu:~$ id ubuntu
uid=1000(ubuntu) gid=1000(ubuntu) groups=1000(ubuntu),4(adm),20(dialout),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),119(netdev),120(lxd),999(docker),122(libvirt)
# yes libvirt
ubuntu@tuomo-ubuntu:~$ newgrp
ubuntu@tuomo-ubuntu:~$ id
uid=1000(ubuntu) gid=1000(ubuntu) groups=1000(ubuntu),4(adm),20(dialout),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),119(netdev),120(lxd)
# still no libvirt in active session
```

Just verifying what is the expected effect of the `newgrp` call here? The `if id "${USER}"` part is returning libvirt on vanilla Ubuntu VM, while the group is not active on the session.

cjeanner:
hmm, newgrp is supposed to re-read the group and load the missing ones out of the the /etc/shadow and /etc/group:
https://www.man7.org/linux/man-pages/man1/newgrp.1.html

We may add the group name, but it means the user would be "removed" (in the current session) from all other groups - which may lead to some other issues...

Lemme do some tests in a local container..

cjeanner:
FUN TIME
so the manpage is wrong, or there's an issue in `newgrp`.. Or I'm too dumb to understand the manpage.

In any cases, it seems `newgrp libvirt` does the right thing for us.

cjeanner:
> > FUN TIME so the manpage is wrong, or there's an issue in `newgrp`.. Or I'm too dumb to understand the manpage.
> > In any cases, it seems `newgrp libvirt` does the right thing for us.
> 
> `newgrp libvirt` does change our current group to `libvirt`, so we go from:
> 
> ```
> ubuntu@tuomo-ubuntu:~$ id
> uid=1000(ubuntu) gid=1000(ubuntu) groups=1000(ubuntu),4(adm),20(dialout),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),119(netdev),120(lxd),122(libvirt),999(docker)
> ```
> 
> to
> 
> ```
> ubuntu@tuomo-ubuntu:~$ newgrp libvirt
> ubuntu@tuomo-ubuntu:~$ id
> uid=1000(ubuntu) gid=122(libvirt) groups=122(libvirt),4(adm),20(dialout),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev),119(netdev),120(lxd),999(docker),1000(ubuntu)
> ```
> 
> This leads to all files being created as `ubuntu:libvirt` instead of `ubuntu:ubuntu`.
right.. and there are "some" files created in that session

> 
> Secondly, as I mentioned `id ubuntu` and `id` as ubuntu user result in different output, which means the `newgrp` line isn't executed, if the user has been added to the group, but it isn't effective yet.

it's in the current session only - so if you open a new shell on the side, you won't see the group. The only real way to get a session wide update is basically to logout and login again... Maybe there should be a "stop" at this point, asking the user to logout/login again and restart the command?

> 
> For this, I think we should have separate `if` for the adding to the group, and separate if for activating the group for the session. WDYT?

It sounds... erk. Groups are a mess to manage, to be fair. We'd basically need to wrap any subshell with a `newgrp` call, and it will more than probably break something at some point, or get lost, or forgotten somewhere.

Also, in order to not change the "main" group, we may do this:
`newgrp libvirt; newgrp $USER` - but it means the user has a "main" group matching their name - I know it's the usual thing, but... May happen it's not the case:

```bash
[foo@5c9962be9c08 /]$ id
uid=1000(foo) gid=1000(foo) groups=1000(foo)
[foo@5c9962be9c08 /]$ newgrp - libvirt
[foo@5c9962be9c08 ~]$ id
uid=1000(foo) gid=999(libvirt) groups=999(libvirt),1000(foo)
[foo@5c9962be9c08 ~]$ newgrp foo
[foo@5c9962be9c08 ~]$ id
uid=1000(foo) gid=1000(foo) groups=1000(foo),999(libvirt)
```

In any cases : the previous thing that was reloading libvirtd.service "to reload groups" was wrong - so I'm wondering how it could work before? Maybe we just don't need all of that?

tuminoid:
It is bit of a mess since the CI is running all the tests on an image that has run `01` script during build, so when we run this script in a PR, the groups are already correct since we've "logged out and back in" at this point. So this issue is more or less only with users who use vanilla  images (or whatever custom base image).

I'll be off next few days, so 
/cc @lentzi90 @Rozzii 
to continue discussion on what we do here.

lentzi90:
As far as I understand, newgrp starts a subshell so we cannot really use it in a script like this. You would need something like

```bash
echo "Before newgrp"
newgrp libvirt <<EONG
echo "hello from within newgrp"
id
# Execute the rest in here...
EONG
```

which does not seem viable.

cjeanner:
> As far as I understand, newgrp starts a subshell so we cannot really use it in a script like this. You would need something like
> 
> ```shell
> echo "Before newgrp"
> newgrp libvirt <<EONG
> echo "hello from within newgrp"
> id
> # Execute the rest in here...
> EONG
> ```
> 
> which does not seem viable.

that's one way, yes. Though it really start the subshell, like you'd do as `su -l $USER`, but without the password request.

So... maybe the best thing is to remove that `newgrp` and be off with that? As said, restarting libvirtd has nothing to do with regards to the (non-libvirt) user groups anyway.... And the best way to get your groups reloaded properly is to logout and login again - so maybe we can set this:
```bash
if ! id "${USER}" | grep -q libvirt; then
  sudo usermod -a -G "libvirt" "${USER}"
  echo "Please logout and login again, and restart this process"
  exit 1
```

Ugly, and automation may suffer a bit (while.. not really? it's not working anyway - guess automation is doing like us: adding the packages/groups/others far, far beforehand), but at least it will ensure a clear status.. ?

lentzi90:
> Ugly, and automation may suffer a bit (while.. not really? it's not working anyway - guess automation is doing like us: adding the packages/groups/others far, far beforehand), but at least it will ensure a clear status.. ?

Yes exactly. In CI we first build a disk image where the 01 script has been executed (so the groups are set up already). Then when we run the tests we start a VM from this disk image, basically the equivalent of doing a reboot after the 01 script :smile:

cjeanner:
> > Ugly, and automation may suffer a bit (while.. not really? it's not working anyway - guess automation is doing like us: adding the packages/groups/others far, far beforehand), but at least it will ensure a clear status.. ?
> 
> Yes exactly. In CI we first build a disk image where the 01 script has been executed (so the groups are set up already). Then when we run the tests we start a VM from this disk image, basically the equivalent of doing a reboot after the 01 script 😄

sooo... shall I go that way? Or just remove that newgrp thing and be off with it?

cjeanner:
fair enough. new iteration coming - I'll do the exact same in the other dependent dev-scripts project.

lentzi90 (commented):
/approve

PR #1325, opened 2023-12-18
Use automatic firmware for UEFI

dtantsur:


tuminoid:
Can you add proper description for the change, if this not a test commit?

PR #1327, opened 2023-12-18
WIP test IRONIC_IMAGE env var

Rozzii:


tuminoid:
> /cc @kashif
Damn autocomplete... 
/cc @kashifest

kashifest:
I am bit lost what will this fix, is it the fix for the unser we talked offline? Can you add it in the description where the unset is done? How about keeping any value intact in between `make` and `make test`?

tuminoid:
> I am bit lost what will this fix, is it the fix for the unset we talked offline? 

Yes.

> Can you add it in the description where the unset is done? 

It's @Rozzii PR.

> How about keeping any value intact in between `make` and `make test`?

I'm working on it elsewhere, this was just another finding in the whole 02 mess.

Rozzii:
@kashifest @tuminoid  just one thing to keep in mind is that unsetting IRONIC_IMAGE is needed as otherwise we would have a conflicting upstream and localy built image

Rozzii:
/close 
in favor or @tuminoid 's more comprehensive solution

PR #1328, opened 2023-12-18
use locally built images in pivoting tests too

tuminoid:
Environment between "make" and "make test" is lost, so we are actually not using the locally built image in the pivoting tests. This has lead to false test results and breaking of integration.

- Add locally built images into config, so they are passed to make test
  and used in pivoting tests, where Ironic is moved to another cluster
- Add `set -u` to catch logic errors
- Change EXT/INT/PRO_IF checks from running the variable content to
  checking the variable content
- Rename some variables more logically, and also handle the LOCAL_IMAGE
  variables that have trailing / properly
- In case of ironic and cluster-api, don't use regex to match image
  name, as they end up building wrong images (like
  cluster-api-provider-metal3, or ironic-client)

NOTE: First commit, we cleanup the script, indentation, braces etc. Second commit, fix the actual issues. (Use the GH interface to) view them separately to help reviewing.

Verified this locally three ways:

- metal3-dev-env without this PR + broken local Ironic (before the iPXE config fix) = pass (old setup isn't capable of detecting the problem, this is expected)
- metal3-dev-env with patch + broken local Ironic (before the iPXE config fix) = fail (expected, it now detects the issues by using right images)
- metal3-dev-env with patch + fixed local Ironic = pass (as it should)

The CI runs in this PR do not actually test any of it, since it is using the pre-built Ironic image, since this PR is vs dev-env, the tested repo is not Ironic, and thus won't trigger building Ironic at all. It only tests that this doesn't break the flow.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -xe
+set -eux
 
 # shellcheck disable=SC1091
 source lib/logging.sh
@@ -12,6 +12,9 @@ source lib/releases.sh
 # pre-pull node and container images
 # shellcheck disable=SC1091
 source lib/image_prepull.sh
+
+# cleanup ci config file if it exists from earlier run
+rm -f "${CI_CONFIG_FILE}"
 
 # (workaround) disable tdp_mmu to avoid
 # kernel crashes with  NULL pointer dereference
@@ -188,7 +191,7 @@ EOF
         sudo nmcli con up provisioning
 
         # Need to pass the provision interface for bare metal
-        if [[ "${PRO_IF}" ]]; then
+        if [[ -n "${PRO_IF}" ]]; then
             sudo tee -a /etc/NetworkManager/system-connections/"${PRO_IF}".nmconnection <<EOF
 [connection]
 id=${PRO_IF}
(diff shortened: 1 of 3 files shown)


tuminoid:
OK, cleanup part works, now adding 2nd commit to address the actual issues.

tuminoid:
Verified this locally three ways:

- metal3-dev-env without this PR + broken local Ironic (before the iPXE config fix) = pass (old setup isn't capable of detecting the problem, this is expected)
- metal3-dev-env with patch + broken local Ironic (before the iPXE config fix) = fail (expected, it now detects the issues by using right images)
- metal3-dev-env with patch + fixed local Ironic = pass (as it should)

The CI runs in this PR do not actually test any of it, since it is using the pre-built Ironic image, since this PR is vs dev-env, the tested repo is not Ironic, and thus won't trigger building Ironic at all. It only tests that this doesn't break the flow.

PTAL.

lentzi90 (commented):
Great find, great cleanup and great fix!
/approve

mquhuy:
Thank you for this PR. Not only did it fix the issue, but the code gets much more error-proof and readable.
Just for perfectionism, I think some of changes in the second commit would be more fitted in the first (cleanup) one; but that's minor details, each of the commits makes senses standing alone.
LGTM

Sunnatillo (commented):
/lgtm

Thank you. It is a good change.

PR #1330, opened 2023-12-21
fix ironic building from source in PR test

tuminoid:
In a PR CI test, repo having the change and being tested is cloned as `/path/to/tested_repo`. This doesn't match `ronic-image` directory like it normally would, and causes Ironic image to not be built from soruce. Same issue with CAPI. Identify the local image via the prefix of the `XXX_LOCAL_IMAGE` variable rather than the variable content (repo path).

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -348,7 +348,7 @@ for IMAGE_VAR in $(env | grep "_LOCAL_IMAGE=" | grep -o "^[^=]*"); do
     IMAGE_DATE="$(date -u +%y_%j_%H%M)"
 
     # Support building ironic-image from source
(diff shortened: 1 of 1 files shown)


tuminoid:
/hold
until local testing finishes.

tuminoid:
/unhold
Worked as expected.
/cc @Sunnatillo @smoshiur1237 @lentzi90 @Rozzii

Rozzii:
Ironic from source doesn't mean that the image is being built from source, it means that ironic is installed from source not a package into the image. that is why it was checking the directory name because it copies the source code to the ironic-image directory, so I think there is a misunderstanding here.
/hold

Rozzii (changes requested):
/lgtm

Rozzii (approved):

lentzi90 (commented):
/approve

Sunnatillo (commented):
/lgtm

PR #1337, opened 2024-01-16
Bump minikube to v1.32.0

kashifest:
as the title says, also updating the related sha digest


tuminoid (commented):
No longer complaining about version skew between client+server :+1: 

/lgtm

lentzi90 (commented):
/approve

PR #1342, opened 2024-01-30
Create overlay in `deploy_tilt_env.sh` to make tilt setup work

maxrantil:
**Backstory**
I had some problems setting up the environment but the errors and fixes are presented here.
After running:
```bash
export BUILD_CAPM3_LOCALLY=true
export BUILD_BMO_LOCALLY=true
export BUILD_IPAM_LOCALLY=true
export BUILD_CAPI_LOCALLY=true

export IMAGE_OS=ubuntu
export EPHEMERAL_CLUSTER="tilt"
make
```
These problems occurred:

**Issue and Resolution Documentation**

1. **Initial Setup Problem:**
   - **Error**: `failed to create fsnotify watcher: too many open files`
   - **Resolution**: Increase the limits of `fs.inotify` parameters:
     ```bash
     sudo sysctl -w fs.inotify.max_user_watches=2099999999
     sudo sysctl -w fs.inotify.max_user_instances=2099999999
     sudo sysctl -w fs.inotify.max_queued_events=2099999999
     ```

2. **Subsequent Error:**
   - **Error**: 
     ```
     modprobe: WARNING: Module kvm_intel is in use.
     modprobe: WARNING: Module kvm is in use.
     ```
   - **Resolution**: Stop related services to release the modules:
     ```bash
     sudo systemctl stop libvirtd.socket
     sudo systemctl stop libvirtd-ro.socket
     sudo systemctl stop libvirtd-admin.socket
     sudo systemctl stop libvirtd
     sudo systemctl stop qemu-kvm
     ```

3. **Script Path Issue**: 
 The final fix was a simple correction in the script deploy_tilt_env.sh to include a missing directory.
 
 Documenting these steps may be useful for others who might face similar setup challenges.

### DIFF
--- a/tilt-setup/deploy_tilt_env.sh
+++ b/tilt-setup/deploy_tilt_env.sh
@@ -22,7 +22,8 @@ cat <<EOF > tilt-settings.json
   }
 }
 EOF
-sed -i 's/yaml = str(kustomizesub(context + "\/config"))/yaml = str(kustomizesub(context + "\/config\/tls"))/' Tiltfile
+REL_PATH_TO_DEV_ENV=$(realpath --relative-to="${BMOPATH}" "${SCRIPTDIR}")
+sed -i "s|yaml = str(kustomizesub(context + \"/config\"))|yaml = str(kustomizesub(\"${REL_PATH_TO_DEV_ENV}/config/overlays/tilt\"))|" Tiltfile
 make kind-reset
 kind create cluster --name capm3 --image="${KIND_NODE_IMAGE}"
 kubectl create namespace "${NAMESPACE}"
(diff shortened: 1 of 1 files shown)


mquhuy:
I think these errors you described can be seen in other scenarios as well, not just in Tilt.
Perhaps we could add a "Troubleshooting" section in our doc, just like what CAPI has: https://cluster-api.sigs.k8s.io/user/troubleshooting. Everything you wrote would fit nicely there.

peppi-lotta (commented):
Did you notice that in this same file on line 25 'config/tls' is used. Does that also need to be changed to 'config/components/tls'?

peppi-lotta:
Also if and when you get Tilt to set up correctly please mark this issue as resolved: https://github.com/metal3-io/baremetal-operator/issues/1508

maxrantil:
> Did you notice that in file 'deploy_tilt_env.sh' on line 25 'config/tls' is used. Does that also need to be changed to 'config/components/tls'?

@peppi-lotta Good catch! No I did not notice that before. It has to do with BMO but will need a change too.

tilt-setup/deploy_tilt_env.sh
```diff
@@ -76,7 +76,7 @@ spec:
         secret:
           secretName: ${IRONIC_SECRET_NAME}
 EOF
-tools/bin/kustomize build config/tls/
+tools/bin/kustomize build config/components/tls/
```
lentzi90, line 79: I don't think this is correct. The path points to a kustomize component, which is not meant to be built directly. It should be used by another kustomization. If you try to `kustomize build` this it just gives an empty output.
I think what we should do is either to use the `config/overlays/basic-auth_tls` overlay or create one here on the fly (this is pretty much what the script does already)

tilt-setup/deploy_tilt_env.sh
```diff
@@ -38,19 +39,38 @@ apply_bm_hosts "${NAMESPACE}"
...
+cat <<EOF > config/overlays/tilt/kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - ../../../${REL_PATH_TO_BMO}/config/namespace
+  - ../../../${REL_PATH_TO_BMO}/config/default
+
+components:
+  - ../../../${REL_PATH_TO_BMO}/config/components/tls
```
lentzi90, line 65: nit: Keep indent the same throughout
```suggestion
- ../../../${REL_PATH_TO_BMO}/config/namespace
- ../../../${REL_PATH_TO_BMO}/config/default

components:
- ../../../${REL_PATH_TO_BMO}/config/components/tls
```

lentzi90 (commented):
Looks like it fixes the current issue, so
/approve

maxrantil:
@mboukhalfa looking good now?

PR #1343, opened 2024-01-30
Bump golang to v1.21.6

kashifest:
as the title says

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -50,7 +50,7 @@
       "matchPackageNames": [
         "golang"
       ],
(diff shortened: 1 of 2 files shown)


smoshiur1237 (approved):
/approve

tuminoid (commented):
/lgtm
/hold
In case you need to coordinate with other PRs.

dtantsur:
Note for the future: changing the playbook has zero effect on the CI because 01_prepare_host is not run there (see https://github.com/metal3-io/metal3-dev-env/pull/1226).

PR #1346, opened 2024-02-05
Copy USE_IRONIC_INSPECTOR from BMO defaults

dtantsur:


03_launch_mgmt_cluster.sh
```diff
@@ -139,6 +139,8 @@ function update_images(){
 function launch_ironic() {
   pushd "${BMOPATH}"
 
+  inspector_default=$(grep USE_IRONIC_INSPECTOR "${BMOPATH}/ironic-deployment/default/ironic_bmo_configmap.env" || true)
```
kashifest, line 142: do we want to always keep it synced with whatever the value is in default configmap env or we should just take it from USER and put it in the ironic_bmo_configmap below

03_launch_mgmt_cluster.sh
```diff
@@ -158,6 +160,7 @@ IRONIC_FAST_TRACK=true
 RESTART_CONTAINER_CERTIFICATE_UPDATED="${RESTART_CONTAINER_CERTIFICATE_UPDATED}"
 IRONIC_RAMDISK_SSH_KEY=${SSH_PUB_KEY_CONTENT}
 IRONIC_USE_MARIADB=${IRONIC_USE_MARIADB:-false}
+${inspector_default}
```
kashifest, line 163: shouldn't we remove line 157 `IRONIC_INSPECTOR_ENDPOINT`

dtantsur: We cannot do it without breaking the current state of things (and the release branches).

03_launch_mgmt_cluster.sh
dtantsur: What do you mean, "from USER"?

kashifest: I mean we can set the default to `true` but overridable through env var if the user want it to be `false`. like the other vars, or is it so that we do not want to have the `false` value anymore, in that case why dont we just put it to `true` rightway by hardcoding it `USE_IRONIC_INSPECTOR=true` here instead of taking it from default kustomize path in BMO.

03_launch_mgmt_cluster.sh
kashifest: ah right !

03_launch_mgmt_cluster.sh
kashifest: something like this https://github.com/metal3-io/ironic-image/blob/3cbaa6d0aeac23ad543468e2ac1e383c439d7d3f/scripts/ironic-common.sh#L106

dtantsur: I'm still not sure what you mean by "user" in the case of centos CI job. We are the users, no?

In the end, I want to switch the default from `true` to `false`. I cannot just do it in BMO because the configuration here prevents that. I cannot just do it here or in ironic-image because the configuration in BMO will try to start Inspector. A variable won't help. Any other ideas?

kashifest: Yes user can be dev user like me/you or the CI etc. What I meant was simply to allow it to be configurable via a variable but as you said variable is not helping you. 

I guess I am not understanding the issue you described. What I understand is during the tests we don't use the `ironic_bmo_configmap.env` from BMO. As you know, we populate the vars for the configmap [here](https://github.com/metal3-io/metal3-dev-env/blob/main/03_launch_mgmt_cluster.sh#L146-L160) and we put the configmap in the `IRONIC_DATA_DIR` for the test and later use this configmap for ironic deployment [here](https://github.com/metal3-io/metal3-dev-env/blob/main/03_launch_mgmt_cluster.sh#L180). So my proposal was to simply add` USE_IRONIC_INSPECTOR=${USE_IRONIC_INSPECTOR:-true}` or `USE_IRONIC_INSPECTOR=true` in the configmap here rather than copying it from BMO dir. However, it seems my understanding here is inadequate.

Its not a big issue for me, I am ok to take it as it is.

tuminoid, line 142: It would be great to have it `^IRONIC_USE_INSPECTOR`, in case someone has a comment line with IRONIC_USE_INSPECTOR on it.

tuminoid (commented):
A nit, but should do the job.

tuminoid:
/cc @lentzi90 
To check it out as well.

lentzi90 (commented):
The dependencies between the repos is ugly but nothing new. I can live with this
/lgtm

PR #1347, opened 2024-02-06
Do not install NetworkManager-initscripts-updown anymore for el9

elfosardo:
We don't use legacy ifup/ifdown scripts anymore


kashifest (commented):
/lgtm

PR #1350, opened 2024-02-16
add make target for prepulling images

Sunnatillo:
Adds make target for prepulling images:

`make prepull_images`


smoshiur1237 (commented):
/approve

PR #1353, opened 2024-02-23
🌱 Write bmhs in separated files

mboukhalfa:
This PR add writing bmh files in separated files while keeping the old way of putting everything in one file because all the tests using dev-env expect it we can after merging this adapt them then remove the old function.

We need this changes because when creating multiple clusters in different namespaces each namespace need its own bmh applying everything from one file is not convenient we need to apply only bmh that we need in each namespace

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -225,6 +225,8 @@ EOF
 # Create the BMH CRs
 #
 function make_bm_hosts() {
+  mkdir -p "${WORKING_DIR}/bmhs"
+  i=0
   while read -r name address user password mac; do
     go run "${BMOPATH}"/cmd/make-bm-worker/main.go \
       -address "$address" \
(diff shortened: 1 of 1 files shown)


maxrantil:
LGTM

03_launch_mgmt_cluster.sh
```diff
@@ -236,13 +236,33 @@ function make_bm_hosts() {
   done
 }
 
+#
+# Create each bmh and secret in a separated file
+# TODO (mboukhalfa): removed the old make_bm_hosts when this adopted by all tests 
+#
+function make_bm_hosts_seprated() {
```
mquhuy, line 243: IMO the name `make_separated_bm_hosts()` might be more suitable. Anw, I think it's better to add a custom param `separate=${1:-false}`, instead of duplicating the function

mboukhalfa: Please check now in once function and saving in separated files and everything in the old file

mquhuy:
lgtm

PR #1354, opened 2024-02-27
Use CAPM3RELEASEBRANCH instead of CAPM3BRANCH for kustomization

kashifest:
for a PR in release-1.5 or release-1.4 branch, `CAPM3BRANCH` is exported as the branch commit hash, this causes the check to fail and the kustomization errors out as we see here https://github.com/metal3-io/cluster-api-provider-metal3/pull/1468. Using `CAPM3RELESEBRANCH` should solve this issue.


Sunnatillo (commented):
/lgtm

PR #1356, opened 2024-03-01
Remove inspector container

lentzi90:
We no longer deploy inspector so we should not try to get logs from it.

### DIFF
--- a/tests/roles/run_tests/tasks/move.yml
+++ b/tests/roles/run_tests/tasks/move.yml
@@ -9,7 +9,6 @@
           - vbmc
       ironic_containers:
           - ironic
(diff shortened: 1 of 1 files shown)


lentzi90:
I hope this is enough. There are still many places where we mention inspector but I think this is the only place where we fail if it is not there

kashifest:
e2e needs https://github.com/metal3-io/cluster-api-provider-metal3/pull/1481 unfortunately. So overriding the e2e tests:
/override test-ubuntu-e2e-integration-main

smoshiur1237 (commented):
/lgtm

PR #1358, opened 2024-03-06
Switch to deploy.go implementation from deploy.sh

maxrantil:
After [PR #1602](https://github.com/metal3-io/baremetal-operator/pull/1602) is merged, which changes the shell script to Go code, this PR is the next step.

/hold

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -27,9 +27,6 @@ if [ "${IRONIC_USE_MARIADB:-false}" == "true" ]; then
   BMO_IRONIC_ARGS+=("-m")
 fi
 
-sudo mkdir -p "${IRONIC_DATA_DIR}"
-sudo chown -R "${USER}:${USER}" "${IRONIC_DATA_DIR}"
-
 # shellcheck disable=SC1091
 source lib/ironic_tls_setup.sh
 # shellcheck disable=SC1091
@@ -39,13 +36,29 @@ source lib/ironic_basic_auth.sh
 # BMO and Ironic deployment functions
 # ------------------------------------
 
+deploy_ironic_and_or_bmo() {
+  export NAMEPREFIX=${NAMEPREFIX:-"baremetal-operator"}
+
+  IRONIC_DATA_DIR="${IRONIC_DATA_DIR:-/opt/metal3/ironic}"
+  IRONIC_AUTH_DIR="${IRONIC_AUTH_DIR:-${IRONIC_DATA_DIR}/auth}"
+
+  sudo mkdir -p "${IRONIC_DATA_DIR}"
+  sudo mkdir -p "${IRONIC_AUTH_DIR}"
+
+  sudo chown -R "${USER}:$(id -gn)" "${IRONIC_DATA_DIR}"
+  sudo chown "${USER}:$(id -gn)" "${IRONIC_AUTH_DIR}"
+
+  sudo chmod -R 755 "${IRONIC_DATA_DIR}"
+  sudo chmod 755 "${IRONIC_AUTH_DIR}"
+
+  "${BMOPATH}/tools/deploy-cli" "$@"
+}
+
 #
 # Create the BMO deployment (not used for CAPM3 v1a4 since BMO is bundeled there)
 #
 launch_baremetal_operator() {
   pushd "${BMOPATH}"
-
-  # Deploy BMO using deploy.sh script
 
 if [ "${EPHEMERAL_CLUSTER}" != "tilt" ]; then
   # Update container images to use local ones
@@ -73,8 +86,8 @@ EOF
     echo "DEPLOY_ISO_URL=${DEPLOY_ISO_URL}" | sudo tee -a "${BMOPATH}/config/default/ironic.env"
   fi
 
-  # Deploy BMO using deploy.sh script
-  "${BMOPATH}/tools/deploy.sh" -b "${BMO_IRONIC_ARGS[@]}"
+  # Deploy BMO using the deploy-cli executable
+	deploy_ironic_and_or_bmo -b "${BMO_IRONIC_ARGS[@]}"
 
   # If BMO should run locally, scale down the deployment and run BMO
   if [ "${BMO_RUN_LOCAL}" == "true" ]; then
@@ -210,8 +223,8 @@ EOF
     update_images
     ${RUN_LOCAL_IRONIC_SCRIPT}
   else
-    # Deploy Ironic using deploy.sh script
-    "${BMOPATH}/tools/deploy.sh" -i "${BMO_IRONIC_ARGS[@]}"
+    # Deploy Ironic using the deploy-cli executable
+    deploy_ironic_and_or_bmo -i "${BMO_IRONIC_ARGS[@]}"
   fi
   popd
 }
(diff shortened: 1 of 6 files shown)


03_launch_mgmt_cluster.sh
```diff
@@ -39,14 +36,31 @@ source lib/ironic_basic_auth.sh
 # BMO and Ironic deployment functions
 # ------------------------------------
 
+function deployIronicAndOrBMO() {
```
tuminoid, line 39:
```suggestion
deploy_ironic_and_or_bmo() {
```
No need to add `function`, and camel case -> snake case.

03_launch_mgmt_cluster.sh
```diff
@@ -39,14 +36,31 @@ source lib/ironic_basic_auth.sh
...
+  export NAMEPREFIX=${NAMEPREFIX:-"baremetal-operator"}
+
+  IRONIC_DATA_DIR="${IRONIC_DATA_DIR:-/opt/metal3/ironic/}"
+  IRONIC_AUTH_DIR="${IRONIC_AUTH_DIR:-${IRONIC_DATA_DIR}auth/}"
```
tuminoid, line 43:
```suggestion
  IRONIC_AUTH_DIR="${IRONIC_AUTH_DIR:-${IRONIC_DATA_DIR}/auth}"
```
Always expect directories not to have trailing slash. Shell doesn't care if you have multiple, but in case its missing, the AUTH_DIR would be invalid.

03_launch_mgmt_cluster.sh
```diff
@@ -39,14 +36,31 @@ source lib/ironic_basic_auth.sh
 # BMO and Ironic deployment functions
 # ------------------------------------
 
+function deployIronicAndOrBMO() {
+  export NAMEPREFIX=${NAMEPREFIX:-"baremetal-operator"}
+
+  IRONIC_DATA_DIR="${IRONIC_DATA_DIR:-/opt/metal3/ironic/}"
```
tuminoid, line 42:
```suggestion
  IRONIC_DATA_DIR="${IRONIC_DATA_DIR:-/opt/metal3/ironic}"
```
Don't include trailing slash by default.

lib/common.sh
```diff
@@ -139,8 +139,12 @@ export FORCE_REPO_UPDATE="${FORCE_REPO_UPDATE:-true}"
 
 export M3PATH="${M3PATH:-${GOPATH}/src/github.com/metal3-io}"
 export BMOPATH="${BMOPATH:-${M3PATH}/baremetal-operator}"
-export BMOREPO="${BMOREPO:-https://github.com/metal3-io/baremetal-operator.git}"
-export BMO_BASE_URL="${BMO_BASE_URL:-metal3-io/baremetal-operator}"
+# export BMOREPO="${BMOREPO:-https://github.com/metal3-io/baremetal-operator.git}"
```
tuminoid, line 142: Remove the commented out lines if they're not necessary.

lib/common.sh
```diff
@@ -171,7 +175,7 @@ export CAPIPATH="${CAPIPATH:-${M3PATH}/cluster-api}"
 export CAPI_BASE_URL="${CAPI_BASE_URL:-kubernetes-sigs/cluster-api}"
 export CAPIREPO="${CAPIREPO:-https://github.com/${CAPI_BASE_URL}}"
 
-export BMOCOMMIT="${BMOCOMMIT:-HEAD}"
+# export BMOCOMMIT="${BMOCOMMIT:-HEAD}"
```
tuminoid, line 178: Same

tilt-setup/deploy_tilt_env.sh
```diff
@@ -28,10 +28,24 @@ make kind-reset
 kind create cluster --name capm3 --image="${KIND_NODE_IMAGE}"
 kubectl create namespace "${NAMESPACE}"
 kubectl create namespace "${IRONIC_NAMESPACE}"
-mkdir -p "${HOME}/.config/cluster-api/overrides/infrastructure-metal3/${CAPM3RELEASE}"
+
+# rm -rf "${CAPI_CONFIG_FOLDER}"/overrides/infrastructure-metal3/"${CAPM3RELEASE}"
```
tuminoid, line 32: Same for all these

tilt-setup/deploy_tilt_env.sh
```diff
@@ -85,17 +99,30 @@ spec:
           - mountPath: /opt/metal3/auth/ironic
             name: ironic-credentials
             readOnly: true
+EOF
+
+if [ -n "${IRONICINSPECTOR_SECRET_NAME}" ]; then
```
tuminoid, line 104:
```suggestion
if [[ -n "${IRONICINSPECTOR_SECRET_NAME}" ]]; then
```
Let's not add new `[ .. ]` since we're using Bash.

03_launch_mgmt_cluster.sh
```diff
@@ -39,14 +36,31 @@ source lib/ironic_basic_auth.sh
...
+  sudo mkdir -p "${IRONIC_AUTH_DIR}"
+
+  sudo chown -R "${USER}:$(id -gn)" "${IRONIC_DATA_DIR}"
+  sudo chown "${USER}:$(id -gn)" "${IRONIC_AUTH_DIR}"
+
+  sudo chmod -R 755 "${IRONIC_DATA_DIR}"
+  sudo chmod 755 "${IRONIC_AUTH_DIR}"
+
+  go build -o "${BMOPATH}/tools/deploy" "${BMOPATH}/tools/deploy.go"
+  "${BMOPATH}/tools/deploy" "${1}" "${@:2}"
```
tuminoid, line 55: Whats the point of this.. I don't see any manipulation of the vars, so `"$@"` would do the same ?

maxrantil: Yes `"$@"` would do the same. I changed it.

maxrantil:
In the file `move_back.yml`, under section   `# Install BMO in Source cluster` the variable `KUBECTL_ARGS` were set to `KUBECTL_ARGS: "{{ KUBECTL_ARGS }}"` before I now took that away. This will pick up the `kube config` from the target cluster (`--kubeconfig=/tmp/kubeconfig-test1.yaml`) and I think this is what has been the issue why it wont pivot back correctly from target cluster to source cluter in our integration tests. In the same file, under section `Install Ironic in Source cluster`, we don't set the variable `KUBECTL_ARGS`. Does anybody have any input on this? Maybe @mboukhalfa ?

lib/common.sh
maxrantil: I have kept the old lines because I thought that I need to revert to them after the testing is done. I am now testing with a PR that is not yet merged as you might notice.

mboukhalfa:
> In the file `move_back.yml`, under section `# Install BMO in Source cluster` the variable `KUBECTL_ARGS` were set to `KUBECTL_ARGS: "{{ KUBECTL_ARGS }}"` before I now took that away. This will pick up the `kube config` from the target cluster (`--kubeconfig=/tmp/kubeconfig-test1.yaml`) and I think this is what has been the issue why it wont pivot back correctly from target cluster to source cluter in our integration tests. In the same file, under section `Install Ironic in Source cluster`, we don't set the variable `KUBECTL_ARGS`. Does anybody have any input on this? Maybe @mboukhalfa ?

Good catch that was a tricky one to discover the BMO deployment was not getting deleted from the sources  and in the `move_back.yml` we were applying this in the wrong cluster because of this `KUBECTL_ARGS` this issue has been fixed in #1373

lib/common.sh
tuminoid: If you need to test with BMO from your branch, put the required config to `config_example.sh` as that is always sourced, then you don't need to mix test config to actual changes.

maxrantil: Gotcha! Thanks for the explanation.

tuminoid, line 142: Dev settings here.
/hold

PR #1361, opened 2024-03-07
Fix tilt setup script after the removal of ironic-inspector

maxrantil:
A bug appeared after the removal of the ironic-inspector. The ironic-inspector-secret was used for ironic and vice versa. This fix will address that issue and also remove the ironic-inspector from our tilt-setup.

### DIFF
--- a/tilt-setup/deploy_tilt_env.sh
+++ b/tilt-setup/deploy_tilt_env.sh
@@ -30,8 +30,12 @@ kubectl create namespace "${NAMESPACE}"
 kubectl create namespace "${IRONIC_NAMESPACE}"
 mkdir -p "${HOME}/.config/cluster-api/overrides/infrastructure-metal3/${CAPM3RELEASE}"
 make tilt-up &
-# wait for cert-manager to be ready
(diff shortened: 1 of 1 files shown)


kashifest (commented):
wondering if this would work with older release branches now. Shall we add `if..else ` instead

maxrantil:
@kashifest What conditions can I use for the if statement? I changed it back to only fix the bug now. I think we can keep the ironic-inspector code if older branches still use it. I didn't think about that before.

maxrantil:
My bad, we need the `if, else` statement, and I noticed I can check on `IRONIC_INSPECTOR_URL` for that.

maxrantil:
I tried removing `IRONIC_INSPECTOR_URL": "${IRONIC_INSPECTOR_URL}`, from the tilt-settings.json file with an if-statement from the script. But that was not straight-forward. Removing it is only a matter of having a neater tilt-settings.json file. For now, this approach should be compatible with both old and new branches.

lentzi90 (commented):
/lgtm

PR #1365, opened 2024-03-12
Cleanup BMH CRs in between multiple make and make cleans

kashifest:
we should cleanup the bmh CRs in between multiple runs.


Sunnatillo (commented):
/lgtm

maxrantil:
LGMT

PR #1373, opened 2024-03-25
🐛 Fix BMO Deployment Deletion and Installation Issues during Pivoting

mboukhalfa:
Co-authored-by: Max Rantil <max.rantil@est.tech>

BMO deployment was not getting deleted from the source cluster in the ansible test this PR fix it.

### DIFF
--- a/tests/roles/run_tests/tasks/move.yml
+++ b/tests/roles/run_tests/tasks/move.yml
@@ -65,13 +65,6 @@
       namespace: "{{ IRONIC_NAMESPACE }}"
     when: EPHEMERAL_CLUSTER == "minikube"
   
-  - name: Remove BMO deployment from source cluster
-    kubernetes.core.k8s:
-      name: "{{ NAMEPREFIX }}"
-      kind: Deployment
-      state: absent
-      namespace: "{{ IRONIC_NAMESPACE }}"
-
   - name: Label baremetalhost CRD to pivot.
     shell: "kubectl label --overwrite crds baremetalhosts.metal3.io {{ item }}"
     with_items:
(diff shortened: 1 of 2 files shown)


smoshiur1237 (commented):
/lgtm

tuminoid:
My question is how this has worked before, if this is the fix? What broke it?

maxrantil:
@tuminoid It was not working correctly earlier. Like @mboukhalfa says in the fist comment, the cluster was never deleted in the source cluster and the pivot back from target cluster to source was not working correctly either because of the variable `KUBECTL_ARGS`. So the test passed because it thought it had pivot back but it had never left.

kashifest:
> @tuminoid It was not working correctly earlier. Like @mboukhalfa says in the fist comment, the cluster was never deleted in the source cluster and the pivot back from target cluster to source was not working correctly either because of the variable `KUBECTL_ARGS`. So the test passed because it thought it had pivot back but it had never left. Check this out: [#1358 (comment)](https://github.com/metal3-io/metal3-dev-env/pull/1358#issuecomment-2018549963)

Cluster was never deleted in source cluster ???

tuminoid:
> @tuminoid It was not working correctly earlier. Like @mboukhalfa says in the fist comment, the cluster was never deleted in the source cluster and the pivot back from target cluster to source was not working correctly either because of the variable `KUBECTL_ARGS`. So the test passed because it thought it had pivot back but it had never left. Check this out: [#1358 (comment)](https://github.com/metal3-io/metal3-dev-env/pull/1358#issuecomment-2018549963)

So it never worked and tests didn't actually test it. Nice....

kashifest:
Ah, so BMO was never removed from source and hence it was working, got it.

kashifest:
> 
> So it never worked and tests didn't actually test it. Nice....

Pivot was working because BMO never got deleted from source , so even when pivot back was installing the BMO in targer wrongly again, the pivot back worked since BMO was already there, its a comedy of errors.

PR #1374, opened 2024-03-26
Echo error string in stderr instead of stdout

kashifest:
Function `get_latest_release()` is called as a subshell and since the echo is in stdout in case an error has occurred, this error string is exported wrongly as the release versions in variables.This in turn fools the empty check and causes the `lib/releases.sh` not to error out. The test moves on and fails elsewhere with a confusing error output.

### DIFF
--- a/lib/releases.sh
+++ b/lib/releases.sh
@@ -58,7 +58,7 @@ function get_latest_release() {
 
   # if release_tag is not found
   if [ -z "${release_tag}" ]; then
-    echo "Error: release is not found from ${release_path}"
+    echo "Error: release is not found from ${release_path}" >&2
     exit 1
   else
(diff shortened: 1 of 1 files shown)


tuminoid (commented):
/lgtm

kashifest:
As you can see, we are lucky to hit the issue right away and now the `lib/releases.sh` is failing right away without moving forward
https://jenkins.nordix.org/job/metal3_metal3_dev_env_main_e2e_integration_test_ubuntu/318/
https://jenkins.nordix.org/job/metal3_metal3_dev_env_release-1-6_integration_test_centos/72/

smoshiur1237 (approved):
/approve

PR #1395, opened 2024-05-02
Fix kustomize v5.4.1 sha value

Sunnatillo:
Fixes kustomize checksum value in dev env


lib/common.sh
```diff
@@ -343,7 +343,7 @@ export KREW_SHA256="${KREW_SHA256:-5df32eaa0e888a2566439c4ccb2ef3a3e6e89522f2f21
 
 # Kustomize version
 export KUSTOMIZE_VERSION="${KUSTOMIZE_VERSION:-v5.4.1}"
-export KUSTOMIZE_SHA256="${KUSTOMIZE_SHA256:-29e4c003b5b22fd017e0da0f71ce0a4ceb7bfd1dec338f0b505b341f087499c8}"
+export KUSTOMIZE_SHA256="${KUSTOMIZE_SHA256:-3d659a80398658d4fec4ec4ca184b432afa1d86451a60be63ca6e14311fc1c42}"
```
kashifest, line 346: wait, how did the PR went in if the sha was not matching ?

Sunnatillo: That is a wonder for me as well

Sunnatillo: Quicly checked the tests run on kustomize uplift PR. 
download_and_install_kustomize function did not run on those tests.

kashifest: I know now whats the issue, we only check in those functions only if the binary exists, we dont check if the binary has the intended version or not, it was the same with minikube which I fixed at some point, I will send a fix for kustomize. 
/approve

smoshiur1237:
Overriding the following required tests, They were already passed on this PR.
/override metal3-ubuntu-e2e-integration-test-main

smoshiur1237 (approved):
/lgtm

PR #1400, opened 2024-05-13
speed up git cloning in dev-env

Rozzii:
This commit:
  - Aims to speed up the git cloning operations in dev-env by making the git clone operations shallow with depth 1
  - Implements a condition to avoid redundant commit checkout
  - Decreases number of checkout operations by one

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -104,14 +104,17 @@ clone_repo() {
   if [[ -d "${REPO_PATH}" ]] && [[ "${FORCE_REPO_UPDATE}" = "true" ]]; then
     rm -rf "${REPO_PATH}"
   fi
-  if [[ ! -d "${REPO_PATH}" ]] ; then
+  if [[ ! -d "${REPO_PATH}" ]]; then
     pushd "${M3PATH}" || exit
-    git clone "${REPO_URL}" "${REPO_PATH}"
-    popd || exit
-    pushd "${REPO_PATH}" || exit
-    git checkout "${REPO_BRANCH}"
-    git checkout "${REPO_COMMIT}"
-    git pull -r || true
+    if [[ "${REPO_COMMIT}" = "HEAD" ]]; then
+        git clone --depth 1 --branch "${REPO_BRANCH}" "${REPO_URL}" \
+            "${REPO_PATH}"
+    else
+        git clone --branch "${REPO_BRANCH}" "${REPO_URL}" "${REPO_PATH}"
+        pushd "${REPO_PATH}" || exit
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/approve

lib/common.sh
```diff
@@ -104,15 +104,16 @@ clone_repo() {
...
-  if [[ ! -d "${REPO_PATH}" ]] ; then
+  if [[ ! -d "${REPO_PATH}" ]]; then
     pushd "${M3PATH}" || exit
-    git clone "${REPO_URL}" "${REPO_PATH}"
-    popd || exit
-    pushd "${REPO_PATH}" || exit
-    git checkout "${REPO_BRANCH}"
-    git checkout "${REPO_COMMIT}"
-    git pull -r || true
+    git clone --depth 1 --branch "${REPO_BRANCH}" "${REPO_URL}" "${REPO_PATH}"
```
mquhuy, line 109: I worry this may cause us pain down the road. How can we make sure that cloning with depth 1 will be sufficient for every repo in the future?

IMO the way to improve speed would be to reduce the needs of cloning the repos in the first place.

Rozzii: In CI environments depth 1 should be enough.
If a user runs dev-env as a local sandbox then it is possible to substitute the automatically cloned repos with manually cloned repos thus if a user needs additional history it can be provided already.  

But please correct me if I have missed something. I can't think of a use-case at the moment that we can't handle with this structure.

lib/common.sh
```diff
@@ -104,15 +104,16 @@ clone_repo() {
...
     popd || exit
+    if [[ "${REPO_COMMIT}" != "HEAD" ]]; then
+        pushd "${REPO_PATH}" || exit
+        git checkout "${REPO_COMMIT}"
```
tuminoid, line 113: This will fail with `--depth 1`. With depth 1, it does not know any other commit than HEAD.

```
git clone --depth 1 --branch main https://github.com/metal3-io/project-infra project-infra-1
# it doesn't know HEAD~2 for example
❯ git checkout HEAD~2
error: pathspec 'HEAD~2' did not match any file(s) known to git
# it doesn't know a sha that is couple commits in history
❯ git checkout 7c8282e6f015246068044cd6b17d66eb4ab2a7db
fatal: reference is not a tree: 7c8282e6f015246068044cd6b17d66eb4ab2a7db
```

/hold

Rozzii: thanks, now it should be better

tuminoid (commented):
/lgtm

PR #1404, opened 2024-05-20
Make minikube the default ephemeral cluster

kashifest:
This PR makes minikube the default ephemeral cluster irrespective of container runtime or image OS. The idea is to run Ubuntu tests only for the time being in different repos until CI is stable. Centos e2es are failing more often. Making minikube ephemeral cluster makes sure that we test ironic as a k8s deployment instead of local containers.


kashifest:
/hold
lets test it properly first.

mboukhalfa:
Can you create an issue to keep track and add TODOs once centos back to normal

kashifest:
> Can you create an issue to keep track and add TODOs once centos back to normal

https://github.com/metal3-io/metal3-dev-env/issues/1405

kashifest:
> /hold We must check e2e tests https://github.com/metal3-io/cluster-api-provider-metal3/blob/95ac037d51099c05255993c53fbb068ec73bf996/scripts/environment.sh#L45

I am actually hardcoding it here , so exporting elsewhere shouldn't be an issue, we also have an export in project-infra.

kashifest:
> I am just thinking about when e2e export that in their environment var should not trigger any conflict later ? If I understand correctly `EPHEMERAL_CLUSTER="minikube"` will be exported in devenv but for e2e still sees `EPHEMERAL_CLUSTER="kind"` is there any unwanted behavior when pivoting back or something ?

In that case lets run it for safety here,
/test metal3-dev-env-integration-test-ubuntu-release-1-6
/test metal3-ubuntu-e2e-integration-test-main

kashifest:
You were right @mboukhalfa 

e2e wont pass with just only this I can see it is now trying to fetch ironic local container logs which it shouldnt , so e2e export is coming into play and needs to change as well if we go this way :

```
STEP: Fetch logs for container ironic @ 05/22/24 07:02:31.24
10:02:39  2024/05/22 07:02:31 exit status 1
10:02:39  Could not open /home/metal3ci/metal3/test/e2e/junit.e2e_suite.1.xml:
10:02:39  open /home/metal3ci/metal3/test/e2e/junit.e2e_suite.1.xml: no such file or directory
```

kashifest:
/override metal3-ubuntu-e2e-integration-test-main  , it would need a separate patch
/override metal3-centos-e2e-integration-test-release-1-7  centos is failing for unrelated reasons

kashifest:
@dtantsur @elfosardo PTAL here, after this change, plan is to run only ubuntu based tests as required in PRs, until we have a solution for centos based jobs. After this change, Ubuntu will run minikube as ephemeral cluster and run ironic as a k8s deployment (not as local containers as it is currently in kind cluster)

Sunnatillo:
/hold

Do we still need this change?

kashifest:
> /hold
> 
> Do we still need this change?

W can discuss on it, I still think this is a better option given that:
1. We can reduce the tests being run
2. Not being dependant on CentOS packages which breaks more often compared to Ubuntu

kashifest:
> We can discuss on it, I still think this is a better option given that:
> 
> 1. We can reduce the tests being run
> 2. Not being dependant on CentOS packages which breaks more often compared to Ubuntu

@Rozzii @lentzi90 @elfosardo how do you folks feel about this ?

lib/common.sh
```diff
@@ -330,11 +330,7 @@ export KUBERNETES_BINARIES_VERSION="${KUBERNETES_BINARIES_VERSION:-${KUBERNETES_
 export KUBERNETES_BINARIES_CONFIG_VERSION=${KUBERNETES_BINARIES_CONFIG_VERSION:-"v0.15.1"}
 
 # Ephemeral Cluster
-if [[ "${CONTAINER_RUNTIME}" = "docker" ]]; then
-  export EPHEMERAL_CLUSTER=${EPHEMERAL_CLUSTER:-"kind"}
-else
-  export EPHEMERAL_CLUSTER="minikube"
-fi
+export EPHEMERAL_CLUSTER="minikube"
```
lentzi90, line 333:
```suggestion
export EPHEMERAL_CLUSTER=${EPHEMERAL_CLUSTER:-"minikube"}
```

lentzi90 (commented):
I see no issues switching to use minikube throughout CI. However, I think we should keep it configurable.

lib/common.sh
kashifest: ow yeah, this I did intentionally here for the test because it gets dictated from different places like project-infra. CAPM3 etc

Rozzii:
I was already proponent of this change and I still support it.
/lgtm

kashifest:
I am guessing no one has any objections on this one, I will push an update on this PR and then we can take it in

dtantsur:
> test ironic as a k8s deployment instead of local containers.

You sold it to me here 👍🏽

elfosardo:
definitely in favor of this!

tuminoid:
If we keep it configurable and mean to support it, we need some periodic test at minimum to verify it keeps working. It is already sometimes breaking after we stopped requiring both ubuntu and centos e2e to be run on PRs.

Rozzii:
@kashifest I think we can cancel the hold

tuminoid:
> If we keep it configurable and mean to support it, we need some periodic test at minimum to verify it keeps working. It is already sometimes breaking after we stopped requiring both ubuntu and centos e2e to be run on PRs.

Added https://github.com/metal3-io/metal3-dev-env/issues/1418 for that. It must be implemented asap, if we merge this and plan to keep local ironic support, otherwise the feature will rot fast and be unsupportable.

kashifest:
I might need to re-consider this PR.
/remove-lifecycle stale

tuminoid:
> > If we keep it configurable and mean to support it, we need some periodic test at minimum to verify it keeps working. It is already sometimes breaking after we stopped requiring both ubuntu and centos e2e to be run on PRs.
> 
> Added #1418 for that. It must be implemented asap, if we merge this and plan to keep local ironic support, otherwise the feature will rot fast and be unsupportable.

> I might need to re-consider this PR. /remove-lifecycle stale

Its been 6+ months since these comments. @kashifest whats is the latest on this PR? IIRC we're more or less agreeing on adding the local ironic script support in IRSO, but the rot issue stands until that and proper IRSO integration happens.

kashifest:
> Its been 6+ months since these comments. @kashifest whats is the latest on this PR? IIRC we're more or less agreeing on adding the local ironic script support in IRSO, but the rot issue stands until that and proper IRSO integration happens.

No update as you can see. I plan to work on it soon but its not on top of my priority list.

kashifest:
We are still juggling in between kind and minikube in our CI. minikube is not as stable as we would wish. I will focus on this PR again in 2 months or so. Meanwhile if anyone is willing to take this forward, feel free to do so.

Rozzii:
I think based on the current state of minikube we can let this issue get closed.

PR #1415, opened 2024-06-07
tie BMO commits for CAPM3RELEASEBRANCH variables

Rozzii:
This commit:
  - ties specific commit hashes of the BMO release tags for each pre-defined CAPM3RELEASEBRANCH variable in the lib/common.sh

The reason is that in case BMO release branch HEAD is used together with a specific BMO container image release tag in the same deployment, there is a chance that a e.g. some patch could introduce CRD misalignment between the unchanged  container image and the moving release branch head thus causing BMO operations or even BMO bootstrapping to fail.


tuminoid:
Alternatively, we could use `release-0.x` tag of the container image and Git branch both?

lib/common.sh
```diff
@@ -269,19 +269,22 @@ if [[ "${CAPM3RELEASEBRANCH}" = "release-1.5" ]]; then
   export KEEPALIVED_TAG="v0.4.2"
   export IRONIC_TAG="v23.1.0"
   export BMOBRANCH="${BMORELEASEBRANCH:-release-0.4}"
+  export BMOBCOMMIT="e92d770b2e8d14276c3798e6e803eec91e2e9f3c" # tag v0.4.2
```
adilGhaffarDev, line 272:
```suggestion
  export BMOCOMMIT="e92d770b2e8d14276c3798e6e803eec91e2e9f3c" # tag v0.4.2
```

Rozzii: thanks done

Rozzii:
> Alternatively, we could use `release-0.x` tag of the container image and Git branch both?

In capm3 e2e we could overwrite these as that is , but in dev-env imo the idea is to deploy the supported releases by default so in dev-env I would go the way I did.

kashifest (commented):
/approve

tuminoid (commented):
/lgtm

PR #1419, opened 2024-06-13
Install ansible in a virtual environment

dtantsur:
Installing packages globally with pip is not recommended since such
packages may conflict with the system. Using a virtual environment is
the approach we have employed in Bifrost for a long time, let's do the
same here.

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -16,7 +16,7 @@ if [[ "${OS}" = "ubuntu" ]]; then
   # make the data retrival more reliable
   sudo sh -c ' echo "Acquire::Retries \"10\";" > /etc/apt/apt.conf.d/80-retries '
   sudo apt-get update
-  sudo apt-get -y install python3-pip python3-dev jq curl wget pkg-config bash-completion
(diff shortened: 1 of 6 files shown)


tuminoid:
/test metal3-dev-env-integration-test-centos-main 
Python installations are somewhat different, so let's run centos as well.

dtantsur:
> Python installations are somewhat different, so let's run centos as well.

Yeah, I wanted one job to pass before I occupy the CI with the rest :)

tuminoid:
> > Python installations are somewhat different, so let's run centos as well.
> 
> Yeah, I wanted one job to pass before I occupy the CI with the rest :)

Ubuntu progressed, so ... :) 

Centos will probably fail tho, there seems to be again some issue with the new cloud + Centos, but it'll probably pass far enough to tell if venv worked or not.

tuminoid:
Don't bother with Centos right now, we're debugging Centos main as all centos e2e are failing now, as VMs are not coming up and hence BMHs are not ever inspected.

tuminoid:
/override metal3-dev-env-integration-test-centos-main 
Not required even, skip.

tuminoid (commented):
/lgtm

PR #1421, opened 2024-06-13
Enable the use of local node images

huutomerkki:
This change enables the usage of local node images by copying them over to IRONIC_IMAGE_DIR from IMAGE_LOCATION/IMAGE_NAME.

### DIFF
--- a/lib/image_prepull.sh
+++ b/lib/image_prepull.sh
@@ -7,9 +7,15 @@ source lib/images.sh
 
 mkdir -p "${IRONIC_IMAGE_DIR}"
 pushd "${IRONIC_IMAGE_DIR}"
-# Downloading image if it does not exist locally
+
 if [[ ! -f "${IMAGE_NAME}" ]]; then
-    wget --no-verbose --no-check-certificate "${IMAGE_LOCATION}/${IMAGE_NAME}"
+    if [[ -f "${IMAGE_LOCATION}/${IMAGE_NAME}" ]]; then
+      # Copy local image  
+      cp "${IMAGE_LOCATION}/${IMAGE_NAME}" .
+    else
+      # Downloading image if it does not exist locally
+      wget --no-verbose --no-check-certificate "${IMAGE_LOCATION}/${IMAGE_NAME}"
+    fi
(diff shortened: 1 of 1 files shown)


smoshiur1237 (commented):
/approve

lib/image_prepull.sh
```diff
@@ -7,9 +7,15 @@ source lib/images.sh
...
-# Downloading image if it does not exist locally
+
 if [[ ! -f "${IMAGE_NAME}" ]]; then
-    wget --no-verbose --no-check-certificate "${IMAGE_LOCATION}/${IMAGE_NAME}"
+    if [[ -f "${IMAGE_LOCATION}/${IMAGE_NAME}" ]]; then
+      # Copy local image  
+      cp "${IMAGE_LOCATION}/${IMAGE_NAME}" .
+    else
+      # Downloading image if it does not exist locally
+      wget --no-verbose --no-check-certificate "${IMAGE_LOCATION}/${IMAGE_NAME}"
```
mboukhalfa, line 17: I am bit confused if the image exist in the location than we copy it to the current path else we download it to the location path ! should not this be downloaded to the current path ?

huutomerkki: In the wget, the IMAGE_LOCATION is the URL of the artifactory, and we do in fact dowload it to current directory.

tuminoid: I*m also confused. We download it to current directory on L17, why would image ever be in the directory checked in L12? Wget has option to create dir structure from URL that'd match L12, but L17 certainly wouldn't then put the image.

huutomerkki: In the case of using a local image, the IMAGE_LOCATION is a filepath and the image needs to be copied over.

mboukhalfa:
/lgtm 
@huutomerkki thinking run tests ?

lib/image_prepull.sh
tuminoid: Still not following the code thanks to overloading IMAGE_LOCATION variable, but I guess it is issue with the code itself and not this change. I'll refactor it some day.

PR #1422, opened 2024-06-13
Remove CentOS Stream 8 configuration leftovers

elfosardo:
CentOS Stream 8 is EoL and we don't really support it. Removing what's left from the configuration.


vm-setup/roles/packages_installation/tasks/main.yml
```diff
@@ -46,24 +46,17 @@
...
-- name: Install packages on CentOS/RHEL8
+- name: Install packages on CentOS/RHEL
   block:
     - name: Install packages on CentOS/RHEL
       package:
           name: "{{ packages.centos.common.packages }}"
           state: present
-    - name: Install packages on CentOS8/RHEL8
-      package:
-          name: "{{ packages.centos.el8.packages }}"
```
dtantsur, line 57: I suspect we can now consolidate the el8 and el9 lists (at least until el10 is a thing, which is relatively soon lol)

elfosardo: that's what blocked me from doing it

dtantsur: yeah, but we still don't need el8 packages any more

vm-setup/roles/packages_installation/tasks/main.yml
```diff
@@ -46,24 +46,17 @@
...
-          state: present
-      when: ansible_distribution_version == "8"
     - name: Install packages on CentOS9/RHEL9
       package:
           name: "{{ packages.centos.el9.packages }}"
           state: present
       when: ansible_distribution_version == "9"
-    - name: Perform CentOS/RHEL8 required configurations
-      include_tasks: centos_required_packages.yml
```
dtantsur, line 66: Can we remove this file now?

Actually, this step is not conditional on version 8, maybe it's still needed?

vm-setup/roles/packages_installation/tasks/main.yml
elfosardo: yep, that's what I forgot, thanks

vm-setup/roles/packages_installation/tasks/main.yml
elfosardo: I thought also it was needed by other centos versions but it looks like only CS8 uses it
I'll remove it

dtantsur (approved):

tuminoid:

This is not relevant really, skip.

PR #1427, opened 2024-06-17
Use Python from the Ansible venv when importing Jinja2

dtantsur:
Nothing installs Jinja2, we just expect Ansible to pull it in. Now that
Ansible is installed in a venv, use Jinja2 from there.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -409,7 +409,7 @@ function launch_cluster_api_provider_metal3() {
 # -------------
 
 function render_j2_config () {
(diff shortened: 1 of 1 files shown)


tuminoid:
Might also need to fix `lib/network.sh` lines 16, 27, and 67 to get correct python used, even if those lines do not explode right now.

tuminoid:
> Might also need to fix `lib/network.sh` lines 16, 27, and 67 to get correct python used, even if those lines do not explode right now.

L67 call is prior setting up venv, so it would not even work actually. Let's skip, module is standard anyways.

tuminoid (commented):
/lgtm

smoshiur1237 (commented):
/approve

PR #1428, opened 2024-06-18
restore centos required packages task

tuminoid:
It was removed in [Centos8 cleanup](https://github.com/metal3-io/metal3-dev-env/pull/1422), but it applies to Centos 9 too.

Without this, we don't have podman etc so it cannot work.


tuminoid:
FYI @elfosardo @dtantsur 
/cc @smoshiur1237

PR #1433, opened 2024-06-19
[WIP] PoC Hostnetworkless Ironic with VirtualMedia Using NodePorts

mboukhalfa:
:warning: Do not merge this is only to demonstrate a PoC

This is a small PoC part of the short focus discussion https://github.com/metal3-io/baremetal-operator/discussions/1739. It demonstrate running Ironic without hostNetwork, limited to the virtualMedia use case, using a nodePort service. 

## Changes Needed

1. **Edit Ironic Deployment:**
   - Remove `hostNetwork: true`.
   - Remove `dnsmasq` container.
   - Remove security restrictions to allow root access for debugging inside the containers.

2. **Add NodePort Service:**
   - Map the following ports:
     ```
     - ironic
       30085:6385
     - inspector
       30050:5050
     - httpd
       30080:30080
     ```
   - Changed the `httpd` internal port because the `HTTP_PORT` variable from the dev-env is used in both internal and external configurations, which would require more changes to configure only the external port.

3. **Edit Ironic ConfigMap:**
   - Remove `PROVISIONING_IP` so that the `runironic` script uses the pod's IP from the `eth0` interface.
   - Add external Ironic IPs to be published to external components like IPA.

4. **Provisioning Networks:**
  - Remove `keepalived` since it cannot access the `ironicendpoint` bridge and instead use manual commands to configure the bridge with the Ironic external IP.
   - Manually remove the IP from minikube and add it to the CP node when pivoting.

/hold


zaneb:
See https://github.com/metal3-io/ironic-standalone-operator/issues/21#issuecomment-2208165986

PR #1437, opened 2024-07-01
remove hardcoded KUBECTL_SHA256 pinning with downloaded SHA check

tuminoid:
Kubectl is not our production binary, so loosen the pinning by changing from hardcoded kubectl pinning to verifying versus downloaded sha256. KUBECTL_SHA256 is passed thru JJB -> project-infra -> CAPM3 -> dev-env and it makes it super difficult to test new k8s versions in sane way, especially when we would actually need to have one version skew to for upgrade scenarios.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -345,11 +345,10 @@ else
   export EPHEMERAL_CLUSTER="minikube"
 fi
 
-# Kubectl version (do not forget to update KUBECTL_SHA256 when changing KUBERNETES_VERSION!)
+# Kubectl version - SHA256 is downloaded and verified
 export KUBECTL_VERSION="${KUBECTL_VERSION:-${KUBERNETES_BINARIES_VERSION}}"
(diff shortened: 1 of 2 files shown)


tuminoid:

Setting up a dev-env is successful test enough for this, as we can see from the logs if it downloaded it or not, and if it matched or not.

tuminoid:
/test metal3-centos-e2e-integration-test-release-1-7 

Let's also see if we can pass older release test (we should, as we just ignore anything passed to us)

tuminoid:
/unhold

Works as expected. It ignores the KUBECTL_SHA256 if passed to it and download the sha and verified it regardless of the version of kubectl.

/cc @Rozzii @kashifest

tuminoid:
/retitle remove hardcoded KUBECTL_SHA256 pinning with downloaded SHA check

kashifest (commented):
/approve

PR #1434, opened 2024-06-19
Cleanup dev env network interfaces

Sunnatillo:
This PR adds scripts to remove network interfaces added by dev env in host cleanup step.
Fixes: #1436

PR #1439, opened 2024-07-05
source lib/network.sh in host_cleanup.sh

tuminoid:
Source lib/network.sh in host_cleanup.sh to get the value of BARE_METAL_PROVISONER_INTERFACE, which is now empty and deletion fails, which causes issues in BML.

Also, fix the formatting issues shellcheck strict complains about.

Cleanup of https://github.com/metal3-io/metal3-dev-env/pull/1434


Rozzii (commented):
/lgtm

PR #1441, opened 2024-07-10
:seedling: Verbose "Image not found" error, if local image doesn't exist or the url is missing the http part

mquhuy:
In `lib/image_prepull.sh`, if the local image doesn't exist, but IMAGE_LOCATION isn't pointing to a remote location, then the script should verbose the image, instead of trying to download.


Sunnatillo (commented):
/lgtm

Rozzii (commented):
/approve

PR #1442, opened 2024-07-25
try with only virt

sgoveas:


mboukhalfa:

I would like to get some description ?  Are you experimenting or there is some need for this changes ? I do have clear idea about the difference between the two definition

dtantsur:
>  Are you experimenting or there is some need for this changes ? I do have clear idea about the difference between the two definition

I expect the final version of the PR to contain an explanation, but tl;dr the versioned machine types can disappear in future versions of libvirt, and apparently this one is no longer available. Just "virt" means "the latest version".

Rozzii (commented):
/lgtm

PR #1449, opened 2024-08-14
Silence fetch logs and fetch manifests

Sunnatillo:
Silence fetch logs and fetch manifests as they are too noise in the jenkins logs, which makes  debugging difficult.

### DIFF
--- a/tests/scripts/fetch_manifests.sh
+++ b/tests/scripts/fetch_manifests.sh
@@ -48,6 +48,8 @@ manifests=(
   m3datatemplate
 )
 
+set +x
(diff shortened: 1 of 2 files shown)


tuminoid (commented):
While this might hide some issues sometimes, I agree it is annoying that the log collection is often most of the log output.

/approve

Sunnatillo:

one test would be enough for this small change.

Rozzii (commented):
/lgtm
I don't really mind the logs but I don't have strong opinion on this.

PR #1458, opened 2024-09-18
Allowing downloadable checksums and different checksum formats

huutomerkki:
This PR changes most hardcoded checksums in this repo to downloadable ones. The exception is krew as there is no checksum available.  Support for checksums provided in a checksums.txt file is also added. 

Also renaming some variables so that if need comes, it'll be easier to add different checksum formats.


Rozzii (commented):
/lgtm

lib/download.sh
```diff
@@ -51,20 +55,31 @@ wget_and_verify()
...
     wget "${args[@]}"
 
-    if [[ "${sha256}" =~ https ]]; then
-        sha256="$(curl -SsL "${sha256}")"
+    if [[ "${checksum}" =~ https ]]; then
+        checksum="$(curl -SsL "${checksum}")"
+    fi
+
+    if [[ "${checksum}" =~ ${dlname,,} ]]; then
+        while read -r line
```
tuminoid, line 63:
```suggestion
        while read -r line; do
```

lib/download.sh
```diff
@@ -51,20 +55,31 @@ wget_and_verify()
...
+        do
```
tuminoid, line 64:
```suggestion
```

tuminoid (commented):
A minor style nit, otherwise looks solid improvement.

Also,
/hold
for a sec to let me test the various scenarios. This seems to handle both direct and multi-arch files, so expecting a quick pass.

lib/download.sh
```diff
@@ -142,6 +159,7 @@ download_and_install_kind()
 download_and_install_tilt()
 {
     TILT_URL="https://raw.githubusercontent.com/tilt-dev/tilt/${TILT_VERSION}/scripts/install.sh"
+    TILT_SHA256="${TILT_SHA256:-https://github.com/tilt-dev/tilt/releases/download/${TILT_VERSION}/checksums.txt}"
```
tuminoid, line 162:
```suggestion
    TILT_SHA256="${TILT_SHA256:-https://github.com/tilt-dev/tilt/releases/download/${TILT_VERSION}/checksums.txt}"
```

lib/download.sh
```diff
@@ -128,6 +144,7 @@ download_and_install_kvm2_driver()
 download_and_install_kind()
 {
     KIND_URL="https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-$(uname)-amd64"
+    KIND_SHA256="${KIND_SHA256:-https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-$(uname)-amd64.sha256sum}"
```
tuminoid, line 147:
```suggestion
    KIND_SHA256="${KIND_SHA256:-https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-$(uname)-amd64.sha256sum}"
```

lib/download.sh
```diff
@@ -51,20 +55,30 @@ wget_and_verify()
     fi
     wget "${args[@]}"
 
-    if [[ "${sha256}" =~ https ]]; then
-        sha256="$(curl -SsL "${sha256}")"
+    if [[ "${checksum}" =~ https ]]; then
+        checksum="$(curl -SsL "${checksum}")"
+    fi
+
+    if [[ "${checksum}" =~ ${dlname,,} ]]; then
```
tuminoid, line 62: This does not work for Tilt. 

1. We download Tilt's installation script, and the checksum is for release tarball. It cannot find install.sh on the checksum file, and skips it.
2. It does not return failure for this case, and happily goes and installs TIlt anyways.

We either need to change the download itself to the release tarball or use hardcoded SHA for the script. Using hardcoded SHA wouldn't be too bad, as it is the same script regardless of arch.

tuminoid (commented):
Whitespace nits + actual issues with Tilt verification.

lib/download.sh
huutomerkki: Ah, I had missed that we use a different install method for tilt. I can change it back to the pinned SHA for now, but changing the installation method shouldn't be too hard either if we want that instead.

lib/download.sh
```diff
@@ -142,7 +159,7 @@ download_and_install_kind()
 download_and_install_tilt()
 {
     TILT_URL="https://raw.githubusercontent.com/tilt-dev/tilt/${TILT_VERSION}/scripts/install.sh"
-    TILT_SCRIPT="install.sh"
+   TILT_SCRIPT="install.sh"
```
tuminoid, line 162:
```suggestion
    TILT_SCRIPT="install.sh"
```

tuminoid (commented):
One whitespace mishap, but this is good to merge otherwise.

/approve
/hold for the whitespace fix

PR #1474, opened 2024-11-25
:seedling: Get lates release with goproxy

peppi-lotta:


lib/releases.sh
```diff
@@ -1,5 +1,51 @@
 #!/bin/bash
 
+# Requires parameters url and version. An optional parameter can be given to
+# exclude some versions. 
+# Example usage:
+# get_latest_release_from_goproxy "https://proxy.golang.org/sigs.k8s.io/cluster-api/@v/list" "v1.8." "beta|rc|pre|alpha"
+function get_latest_release_from_goproxy() {
+
```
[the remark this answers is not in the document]
peppi-lotta: `get_latest_release_from_goproxy ` is as concise as I know how to make it.  The headers need to be removed to allow the `sort` to work correctly. With out removal of the headers  the the output looks like this and we can't get the latest release from it.
```
curl -si https://proxy.golang.org/sigs.k8s.io/cluster-api/@v/list | sort -rV

x-xss-protection: 0
x-frame-options: SAMEORIGIN
x-content-type-options: nosniff
vary: Sec-Fetch-Site
vary: Sec-Fetch-Mode
vary: Sec-Fetch-Dest
v1.9.0-beta.1
v1.9.0-beta.0
v1.8.5
```

The original way is currently left as a fail safe and that is why the file has so much code in it. We agreed on this with @adilGhaffarDev.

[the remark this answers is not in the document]
tuminoid: Yes, fallback is fine. The comment above I wrote has single-liner which is basically achieving everything the function is supposed to do.

lib/releases.sh
```diff
@@ -65,56 +101,55 @@ function get_latest_release() {
...
-  export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.8.")}"
+  export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release_from_goproxy "${CAPIGOPROXY}" "v1.8.")}"
 fi
 
 CAPIBRANCH="${CAPIBRANCH:-${CAPIRELEASE}}"
 
 # On first iteration, jq might not be installed
 if [[ "$CAPIRELEASE" == "" ]]; then
-  command -v jq &>/dev/null && echo "Failed to fetch CAPI release from Github" && exit 1
+  command -v jq &>/dev/null && echo "Failed to fetch CAPI release from Github"
```
adilGhaffarDev, line 128:
```suggestion
  echo "Failed to fetch CAPI release from GOPROXY, now trying to fetch with Github"
```
we don't need `command -v jq &>/dev/null` , not sure why this was here before.

lib/releases.sh
```diff
@@ -65,56 +101,55 @@ function get_latest_release() {
...
+    export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.8.")}"
+  else
+    # 1.9.99 points to the head of the main branch of CAPM3. Local override for CAPM3 is created for this version.
+    export CAPM3RELEASE="v1.9.99"
+    export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.8.")}"
+  fi
+
 fi
 
 if [[ "$CAPM3RELEASE" == "" ]]; then
```
adilGhaffarDev, line 153: let's check CAPIRELEASE too here. or maybe in a separate if check. But we need to check if we are able to find CAPIRELEASE.

lib/releases.sh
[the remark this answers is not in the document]
adilGhaffarDev: I agree with @tuminoid . I think it's looking good now. I'm fine with keeping it as a parameterized function rather than reducing it to a one-liner; it makes it more reusable.

lib/releases.sh
```diff
@@ -65,56 +101,58 @@ function get_latest_release() {
...
 
 # On first iteration, jq might not be installed
 if [[ "$CAPIRELEASE" == "" ]]; then
-  command -v jq &>/dev/null && echo "Failed to fetch CAPI release from Github" && exit 1
+  echo "Failed to fetch CAPI release from GOPROXY, now trying to fetch with Github"
+  # CAPM3, CAPI and BMO release path
+  CAPIRELEASEPATH="{https://api.github.com/repos/${CAPI_BASE_URL:-kubernetes-sigs/cluster-api}/releases}"
+
+  # CAPM3, CAPI and BMO releases
+  if [ "${CAPM3RELEASEBRANCH}" = "release-1.6" ]; then
```
tuminoid, line 133:
```suggestion
  if [[ "${CAPM3RELEASEBRANCH}" = "release-1.6" ]]; then
```

lib/releases.sh
```diff
@@ -65,56 +101,58 @@ function get_latest_release() {
...
+    # 1.6.99 points to the head of the release-1.6 branch. Local override for CAPM3 is created for this version.
+    export CAPM3RELEASE="v1.6.99"
+    export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.6.")}"
+  elif [ "${CAPM3RELEASEBRANCH}" = "release-1.7" ]; then
```
tuminoid, line 137:
```suggestion
  elif [[ "${CAPM3RELEASEBRANCH}" = "release-1.7" ]]; then
```

03_launch_mgmt_cluster.sh
```diff
@@ -10,6 +10,7 @@ source lib/releases.sh
 # shellcheck disable=SC1091
 source lib/network.sh
 
+export CAPI_CONFIG_FOLDER="${HOME}/.cluster-api"
```
adilGhaffarDev, line 13: according to this condition https://github.com/metal3-io/metal3-dev-env/pull/1474/files#diff-417b79700310fc30a2ae54bd699206962d6a97cfe30018a66ba6854312acdf9dL114 
CAPI_CONFIG_FOLDER should be set like this:
```
  # Default CAPI_CONFIG_FOLDER to $HOME/.config folder if XDG_CONFIG_HOME not set
  CONFIG_FOLDER="${XDG_CONFIG_HOME:-$HOME/.config}"
  export CAPI_CONFIG_FOLDER="${CONFIG_FOLDER}/cluster-api"
```

lib/releases.sh
```diff
@@ -65,56 +101,58 @@ function get_latest_release() {
...
+  else
+    # 1.9.99 points to the head of the main branch of CAPM3. Local override for CAPM3 is created for this version.
+    export CAPM3RELEASE="v1.9.99"
+    export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.8.")}"
+  fi
 fi
 
-if [[ "$CAPM3RELEASE" == "" ]]; then
-  command -v jq &>/dev/null && echo "Failed to fetch CAPM3 release from Github" && exit 1
+if [[ "$CAPIRELEASE" == "" ]]; then
```
tuminoid, line 152:
```suggestion
if [[ -z "${CAPIRELEASE}" ]]; then
```

lib/releases.sh
```diff
@@ -65,56 +101,58 @@ function get_latest_release() {
...
+  echo "Failed to fetch CAPI release from Github" && exit 1
```
tuminoid, line 153:
```suggestion
  echo "Failed to fetch CAPI release from Github" 
  exit 1
```

lib/releases.sh
```diff
@@ -65,56 +101,58 @@ function get_latest_release() {
...
-version_string="${CAPIRELEASE#v}"
-IFS='.' read -r _ minor _ <<< "$version_string"
-
-if [ "$minor" -lt 5 ]; then
-  export CAPI_CONFIG_FOLDER="${HOME}/.cluster-api"
-else
-  # Default CAPI_CONFIG_FOLDER to $HOME/.config folder if XDG_CONFIG_HOME not set
-  CONFIG_FOLDER="${XDG_CONFIG_HOME:-$HOME/.config}"
-  export CAPI_CONFIG_FOLDER="${CONFIG_FOLDER}/cluster-api"
+if [[ "$CAPM3RELEASE" == "" ]]; then
```
tuminoid, line 156:
```suggestion
if [[ -z "${CAPM3RELEASE}" ]]; then
```

lib/releases.sh
```diff
@@ -65,56 +101,58 @@ function get_latest_release() {
...
+  echo "Failed to fetch CAPM3 release from Github" && exit 1
```
tuminoid, line 157:
```suggestion
  echo "Failed to fetch CAPM3 release from Github"
  exit 1
```

lib/releases.sh
```diff
@@ -1,5 +1,41 @@
 #!/bin/bash
 
+# Requires parameters url and version. An optional parameter can be given to
+# exclude some versions. 
+# Example usage:
+# get_latest_release_from_goproxy "https://proxy.golang.org/sigs.k8s.io/cluster-api/@v/list" "v1.8." "beta|rc|pre|alpha"
+get_latest_release_from_goproxy() {
+  # fail when release_path is not passed
```
tuminoid, line 8: this comment this not needed
```suggestion
```

lib/releases.sh
```diff
@@ -1,5 +1,41 @@
...
+  local proxuUrl="${1:?no release path is given}"
+
+  # fail when release is not passed
```
tuminoid, line 11: nor this
```suggestion
```

lib/releases.sh
```diff
@@ -1,5 +1,41 @@
...
+  local release="${2:?no release given}"
+
```
tuminoid, line 13: Remove empty lines between variable declarations.

lib/releases.sh
```diff
@@ -65,56 +101,58 @@ function get_latest_release() {
...
   # 1.9.99 points to the head of the main branch of CAPM3. Local override for CAPM3 is created for this version.
   export CAPM3RELEASE="v1.9.99"
-  export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.8.")}"
+  export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release_from_goproxy "${CAPIGOPROXY}" "v1.8.")}"
 fi
 
 CAPIBRANCH="${CAPIBRANCH:-${CAPIRELEASE}}"
 
 # On first iteration, jq might not be installed
 if [[ "$CAPIRELEASE" == "" ]]; then
```
tuminoid, line 127:
```suggestion
if [[ -z "${CAPIRELEASE}" ]]; then
```

03_launch_mgmt_cluster.sh
```diff
@@ -10,6 +10,7 @@ source lib/releases.sh
 # shellcheck disable=SC1091
 source lib/network.sh
 
+export CAPI_CONFIG_FOLDER="${CONFIG_FOLDER}/cluster-api"
```
tuminoid, line 13: The tests fail to `mkdir -p /cluster-api`  which tells us CONFIG_FOLDER is empty. We don't run `set -eu`, only` set -xe` so it fails to that.

tuminoid (commented):
/approve

03_launch_mgmt_cluster.sh
```diff
@@ -10,6 +10,9 @@ source lib/releases.sh
 # shellcheck disable=SC1091
 source lib/network.sh
 
+# Default CAPI_CONFIG_FOLDER to $HOME/.config folder if XDG_CONFIG_HOME not set
+CONFIG_FOLDER="${XDG_CONFIG_HOME:-$HOME/.config}"
```
dtantsur, line 14: nit: "directory" is a more common word in the Linux world

lib/releases.sh
```diff
@@ -65,56 +90,60 @@ function get_latest_release() {
...
-# CAPM3, CAPI and BMO release path
-CAPIRELEASEPATH="{https://api.github.com/repos/${CAPI_BASE_URL:-kubernetes-sigs/cluster-api}/releases}"
-
+CAPIGOPROXY="https://proxy.golang.org/sigs.k8s.io/cluster-api/@v/list"
 # CAPM3, CAPI and BMO releases
-if [ "${CAPM3RELEASEBRANCH}" = "release-1.6" ]; then
+if [[ "${CAPM3RELEASEBRANCH}" = "release-1.6" ]]; then
   # 1.6.99 points to the head of the release-1.6 branch. Local override for CAPM3 is created for this version.
   export CAPM3RELEASE="v1.6.99"
-  export CAPIRELEASE="${CAPIRELEASE:-$(get_latest_release "${CAPIRELEASEPATH}" "v1.6.")}"
```
dtantsur, line 75: nit: would be great to avoid duplication by setting e.g. CAPI_RELEASE_PREFIX="v1.6." and then only setting CAPIRELEASE once after the "if" block

lib/releases.sh
```diff
@@ -1,5 +1,30 @@
 #!/bin/bash
 
+# Requires parameters url and version. An optional parameter can be given to
+# exclude some versions. 
+# Example usage:
+# get_latest_release_from_goproxy "https://proxy.golang.org/sigs.k8s.io/cluster-api/@v/list" "v1.8." "beta|rc|pre|alpha"
+get_latest_release_from_goproxy() {
```
dtantsur, line 7: It does not look like get_latest_release is used anywhere else, so you can probably replace the old function instead of creating a new one with a new name

Rozzii (commented):
/lgtm

PR #1475, opened 2024-11-26
cleanup and reformat bash files at repositor root

tuminoid:
It is time to make dev-env more consistent as we see plenty of changes happening here.

Make all root level bash files look the same. Same indent, same curlys, same nice bash syntax all around.

0* scripts will be fixed in follow-up, and libs after that.

### DIFF
--- a/host_cleanup.sh
+++ b/host_cleanup.sh
@@ -15,25 +15,26 @@ remove_ironic_containers
 sudo "${CONTAINER_RUNTIME}" rm -f fake-ipa 2>/dev/null
 
 # Remove existing pod
-if [[ "${CONTAINER_RUNTIME}" == "podman" ]]; then
-  for pod in ironic-pod infra-pod; do
-    if sudo "${CONTAINER_RUNTIME}" pod exists "${pod}"; then
-      sudo "${CONTAINER_RUNTIME}" pod rm "${pod}" -f
-    fi
-  done
+if [[ "${CONTAINER_RUNTIME}" = "podman" ]]; then
+    for pod in ironic-pod infra-pod; do
+        if sudo "${CONTAINER_RUNTIME}" pod exists "${pod}"; then
+           sudo "${CONTAINER_RUNTIME}" pod rm "${pod}" -f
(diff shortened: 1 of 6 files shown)


tuminoid:
/test metal3-dev-env-integration-test-ubuntu-main metal3-centos-e2e-integration-test-release-1-8
Unrelated fail.

dtantsur:
In an ideal world, we would rewrite this all in something like ansible.. we're not in an ideal world, and the change looks good so far.

tuminoid:
> In an ideal world, we would rewrite this all in something like ansible.. we're not in an ideal world, and the change looks good so far.

Yes, there are ideas how to improve (in my ideal world it is not ansible though :laughing:), but since we're seeing lots of changes coming here, it'll make reviewing easier to have all these bash monsters at least looking alike.

PTAL.
/cc @dtantsur @lentzi90 @smoshiur1237

smoshiur1237:
Changes are looking good. I will check one more time

smoshiur1237 (commented):
Thanks for the improvement. 
/approve

host_cleanup.sh
```diff
@@ -55,26 +56,26 @@ ANSIBLE_FORCE_COLOR=true "${ANSIBLE}-playbook" \
...
-    sudo nmcli con delete "${BARE_METAL_PROVISIONER_INTERFACE}"
-    sudo nmcli con delete provisioning
-  fi
-  # External net should have been cleaned already at this stage, but we double
-  # check as leaving it around causes issues when the host is rebooted
-  if [[ "${MANAGE_EXT_BRIDGE}" == "y" ]]; then
-    sudo nmcli con delete external || true
-  fi
+    sudo rm -rf /etc/NetworkManager/conf.d/dnsmasq.conf
+    if [[  "${MANAGE_PRO_BRIDGE}" == "y" ]]; then
```
dtantsur, line 60: nit: redundant space after [[

dtantsur:
I don't think I have any rights here but

/lgtm

PR #1480, opened 2024-12-10
Wait for local ironic to be up

lentzi90:
This ensures that ironic is ready before proceeding when running ironic locally. It is specifically needed when SKIP_APPLY_BMH is set to true. In this case we will otherwise continue straight to 04_verify.sh with a high probability of finding the ironic endpoint unavailable.

When SKIP_APPLY_BMH is NOT true, we will in any case implicitly wait for ironic, since the BMHs won't become ready until ironic does.

This should fix the clusterctl upgrade job: https://jenkins.nordix.org/view/Metal3%20Periodic/job/metal3-periodic-e2e-clusterctl-upgrade-test-main/

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -521,6 +521,26 @@ iterate(){
   return "${TMP_RET_CODE}"
 }
 
+#
+# Retry a command until it runs successfully or exceeds the maximum retries
+#
+# Inputs:
+# - the command to run
+#
+retry()
+{
+    local retries=10
+    local i
+    for i in $(seq 1 "${retries}"); do
+        if "${@}"; then
+            return 0
+        fi
+        echo "Retrying... ${i}/${retries}"
+        sleep 5
+    done
+    return 1
+}
+
 
 #
 # Check the return code
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -232,6 +232,9 @@ EOF
     if [[ "${EPHEMERAL_CLUSTER}" != "minikube" ]]; then
         update_images
         ${RUN_LOCAL_IRONIC_SCRIPT}
+        # Wait for ironic to become ready
+        echo "Waiting for Ironic to become ready"
+        retry sudo "${CONTAINER_RUNTIME}" exec ironic /bin/ironic-readiness
     else
         # Deploy Ironic using deploy.sh script
         "${BMOPATH}/tools/deploy.sh" -i "${BMO_IRONIC_ARGS[@]}"


lib/common.sh
```diff
@@ -497,8 +497,10 @@ iterate(){
   local RUNS=0
   local COMMAND="$*"
   local TMP_RET TMP_RET_CODE
+  set +e
```
lentzi90, line 500: Allowing errors here since we would otherwise immediately exit `iterate` and the whole point of the function is to retry until successful.

tuminoid: Technically in library code you should never just set or unset any of the flags, without reading the exact value and then restoring that. Here we just assume we have -e, and toggle it back and forth, but we might not actually have it enabled in the first place.

That is one reason why we should maybe do it in caller side.

lib/common.sh
```diff
@@ -497,8 +497,10 @@ iterate(){
...
   TMP_RET="$(${COMMAND})"
```
tuminoid, line 501: This is also horrible way of checking the success of a command, when we care only about 0 or non-zero.

lib/common.sh
```diff
@@ -497,8 +497,10 @@ iterate(){
...
   TMP_RET_CODE="$?"
+  set -e
 
   until [[ "${TMP_RET_CODE}" = 0 ]] || [[ "${SKIP_RETRIES}" = true ]]
   do
```
tuminoid, line 506:
```suggestion
  until "${COMMAND}" || [[ "${SKIP_RETRIES}" = true ]]; do
```
and deleting all TMP_RET stuff would be much cleaner.

lib/common.sh
```diff
@@ -512,9 +514,11 @@ iterate(){
       break
     fi
     sleep "${TEST_TIME_INTERVAL}"
+    set +e
```
tuminoid, line 517: Also, why is it duplicated here?

lib/common.sh
```diff
@@ -512,9 +514,11 @@ iterate(){
...
     # shellcheck disable=SC2068
     TMP_RET="$(${COMMAND})"
     TMP_RET_CODE="$?"
+    set -e
   done
   FAILS="$((FAILS+TMP_RET_CODE))"
```
tuminoid, line 523: Were counting failures but not actually using it for anything, also also counting it based on the return value which can be anything else than 0 or 1 depending on the command?

tuminoid (commented):
Sorry, this review went a bit sideways about the code you're touching, but it has a point in there too.

lentzi90:
Thanks @tuminoid ! Let's see how it works!
/test metal3-e2e-clusterctl-upgrade-test-main

tuminoid:
> Thanks @tuminoid ! Let's see how it works! /test metal3-e2e-clusterctl-upgrade-test-main

Looks much cleaner now, but the original logic was quite weird so let's indeed see what hidden assumptions we break with this :)

lentzi90:
Hmm it does not look good.
Now it went through all retries and then continued without failing:

```
01:09:20.821  + sleep 10
01:09:31.868  + 'sudo docker exec ironic /bin/ironic-readiness'
01:09:31.868  lib/common.sh: line 500: sudo docker exec ironic /bin/ironic-readiness: No such file or directory
01:09:31.868  + [[ false = true ]]
01:09:31.868  + [[ 239 = \0 ]]
01:09:31.868  + RUNS=240
01:09:31.868  + [[ 240 = \2\4\0 ]]
01:09:31.868  + break
01:09:31.868  + set -e
```

In the verify script is also has issues:

```
01:26:22.365  lib/common.sh: line 500: check_k8s_entity deployments capm3-system:capm3-controller-manager   capi-system:capi-controller-manager   capi-kubeadm-bootstrap-system:capi-kubeadm-bootstrap-controller-manager   capi-kubeadm-control-plane-system:capi-kubeadm-control-plane-controller-manager   baremetal-operator-system:baremetal-operator-controller-manager: command not found
```

Seems like it doesn't find files/commands. Not sure what to make of that.

tuminoid:
> In the verify script is also has issues:
> 
> ```
> 01:26:22.365  lib/common.sh: line 500: check_k8s_entity deployments capm3-system:capm3-controller-manager   capi-system:capi-controller-manager   capi-kubeadm-bootstrap-system:capi-kubeadm-bootstrap-controller-manager   capi-kubeadm-control-plane-system:capi-kubeadm-control-plane-controller-manager   baremetal-operator-system:baremetal-operator-controller-manager: command not found
> ```
> 
> Seems like it doesn't find files/commands. Not sure what to make of that.

That seems like quotation issue, it is executing the whole thing as a command instead of executing the function with some arguments.

lib/common.sh
```diff
@@ -496,12 +496,8 @@ list_nodes() {
 iterate(){
   local RUNS=0
   local COMMAND="$*"
-  local TMP_RET TMP_RET_CODE
-  TMP_RET="$(${COMMAND})"
-  TMP_RET_CODE="$?"
 
-  until [[ "${TMP_RET_CODE}" = 0 ]] || [[ "${SKIP_RETRIES}" = true ]]
-  do
+  until "${COMMAND}" || [[ "${SKIP_RETRIES}" = true ]]; do
```
tuminoid, line 500:
```suggestion
  until "$@" || [[ "${SKIP_RETRIES}" = true ]]; do
```
$* makes it a single string, which probably the subshell expansion "rescued". Use "$@" directly.

lentzi90:
Oh my bad. Unbound variable `COMMAND`. How did shellcheck not catch that? :thinking:

tuminoid:
> Oh my bad. Unbound variable `COMMAND`. How did shellcheck not catch that? 🤔

While cleaning up I've been adding `-u` and making difference between local and global variables, which are often overlapping etc making it impossible for shellcheck to know whats in scope.

lentzi90:
Oh it is again quotation I think. `"capm3-controller-manager   capi-system"` :facepalm:

lentzi90:
I'm not sure how to make this pass. It is failing in the 04 script now on the `check_k8s_entity` and it looks like some refactoring may be needed to get it going. :disappointed:

lentzi90:
It is doing some counting that I don't understand
```
01:27:12.754  FAIL - Replica sets with label cluster.x-k8s.io/provider=infrastructure-metal3 created
01:27:12.754         expected 2   cluster.x-k8s.io/provider, got 2
```

tuminoid:
> It is doing some counting that I don't understand
> 
> ```
> 01:27:12.754  FAIL - Replica sets with label cluster.x-k8s.io/provider=infrastructure-metal3 created
> 01:27:12.754         expected 2   cluster.x-k8s.io/provider, got 2
> ```

I'm literally laughing at the same piece of code right now. It counts failures, but uses global variables and it adds, not the number of actual failures, but the return code which may be something else than 0 or 1 to the number, leading to completely misleading results. And of course it uses bunch of functions from lib/common.sh that share the same variables and message vars globally, leading to more chaos.

At this rate, I'm rewriting half of lib/common.sh to sort out 04 ... :(

lentzi90:
Alright, let's see how this goes. Now I circumvented the `iterate` and implemented a new `retry` function instead. Perhaps we can "migrate" over the rest and get rid of `iterate` later?
/test metal3-e2e-clusterctl-upgrade-test-main

lentzi90:
Looks like it is working
```
00:16:04.049  + for i in $(seq 1 "${retries}")
00:16:04.049  + sudo docker exec ironic /bin/ironic-readiness
00:16:04.049  exec /bin/ironic-readiness: exec format error
00:16:04.049  + echo 'Retrying... 1/10'
00:16:04.049  Retrying... 1/10
00:16:04.049  + sleep 5
00:16:07.966  + for i in $(seq 1 "${retries}")
00:16:07.966  + sudo docker exec ironic /bin/ironic-readiness
00:16:07.966  curl: (7) Couldn't connect to server
00:16:07.966  + echo 'Retrying... 2/10'
00:16:07.966  Retrying... 2/10
00:16:07.966  + sleep 5
00:16:12.874  + for i in $(seq 1 "${retries}")
00:16:12.874  + sudo docker exec ironic /bin/ironic-readiness
00:16:12.874  {"name": "OpenStack Ironic API", "description": "Ironic is an OpenStack project which enables the provision and management of baremetal machines.", "default_version": {"id": "v1", "links": [{"href": "https://172.22.0.2:6385/v1/", "rel": "self"}], "status": "CURRENT", "min_version": "1.1", "version": "1.95"}, "versions": [{"id": "v1", "links": [{"href": "https://172.22.0.2:6385/v1/", "rel": "self"}], "status": "CURRENT", "min_version": "1.1", "version": "1.95"}]}+ return 0
```

I am now very skeptical of that 04 script though... It is saying "OK - Fetch Baremetalhosts" even though this test does NOT apply BMHs until later... :rofl: 
Some checks are deliberately skipped in this case. That I can see from the code. But I don't know what good these checks are at this point.

lentzi90:
Network flake...

tuminoid:
> Oh it is again quotation I think. `"capm3-controller-manager capi-system"` 🤦

It is hitting this again. :cry:

lentzi90:
> It is hitting this again. 😢

Where do you see it? Maybe you saw the previous run?

/test metal3-centos-e2e-integration-test-release-1-8

tuminoid:
> > It is hitting this again. 😢
> 
> Where do you see it? Maybe you saw the previous run?

Well it was there on the latest failure, but now all the failures are gone, so .... :confused:

tuminoid (commented):
/approve
/lgtm

Let's get this in for the nightly periodics.

PR #1483, opened 2024-12-12
launch irso only we we actually use it

tuminoid:
Currently IRSO is launched even we don't want IRSO controlled Ironic. Thanks to a manifest mismatch now, it is just crashing around. Move launching IRSO behind "if USE_IRSO" check.


lentzi90 (commented):
/approve

lentzi90:
Please ignore, just generating some data for prow
/test shellcheck

Rozzii (commented):
/lgtm

PR #1487, opened 2024-12-20
🌱 Check the default value of CAPM3RELEASE before setting it

adilGhaffarDev:
Check the default value of CAPM3RELEASE before setting it, we will set this value in CAPM3 for e2e tests
Part of https://github.com/metal3-io/cluster-api-provider-metal3/issues/2181


lib/releases.sh
```diff
@@ -37,10 +37,10 @@ fi
 # Fetch CAPI version that coresponds to CAPM3_RELEASE_PREFIX release version
 if [[ "${CAPM3_RELEASE_PREFIX}" =~ ^(1\.6|1\.7|1\.8|1\.9)$ ]]; then
   export CAPM3RELEASE="v${CAPM3_RELEASE_PREFIX}.99"
-  CAPI_RELEASE_PREFIX="v${CAPM3_RELEASE_PREFIX}."
+  export CAPI_RELEASE_PREFIX="v${CAPM3_RELEASE_PREFIX}."
```
tuminoid, line 40: Why do we have to export it, if it wasn't exported before?

lib/releases.sh
```diff
@@ -37,10 +37,10 @@ fi
...
 else
-  export CAPM3RELEASE="v1.10.99"
-  CAPI_RELEASE_PREFIX="v1.9."
+  export CAPM3RELEASE="${CAPM3RELEASE:-"v1.10.99"}"
+  export CAPI_RELEASE_PREFIX="${CAPI_RELEASE_PREFIX:-"v1.9."}"
```
tuminoid, line 43: same here?

tuminoid (commented):
Nits/questions.

lib/releases.sh
```diff
@@ -39,8 +39,8 @@ if [[ "${CAPM3_RELEASE_PREFIX}" =~ ^(1\.6|1\.7|1\.8|1\.9)$ ]]; then
   export CAPM3RELEASE="v${CAPM3_RELEASE_PREFIX}.99"
   CAPI_RELEASE_PREFIX="v${CAPM3_RELEASE_PREFIX}."
 else
-  export CAPM3RELEASE="v1.10.99"
-  CAPI_RELEASE_PREFIX="v1.9."
+  export CAPM3RELEASE="${CAPM3RELEASE:-"v1.10.99"}"
+  CAPI_RELEASE_PREFIX="${CAPI_RELEASE_PREFIX:-"v1.9."}"
```
kashifest, line 43: the version prefix mismatch between CAPM3 and CAPI would be very confusing though, although I know currently theres no other way

tuminoid (commented):
/approve

lib/releases.sh
adilGhaffarDev: v1.10.99 is the main branch of CAPM3, I think we should start testing CAPI main instead of latest(v1.9) with CAPM3 main. But let's discuss it further and do it in other PR.

kashifest (commented):
/lgtm

PR #1494, opened 2025-01-20
Update renovate.json config

kashifest:
Renovate is quiet for few months now. The `"enabled": "true"` setting is using a string instead of boolean. Changing it to boolean to see if it triggers something. Also updating the assignee and the reviewers section
Signed-off-by: Kashif Khan <kashif.khan@est.tech>

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,7 +2,7 @@
   "extends": [
     "config:base"
   ],
-  "enabled": "true",
+  "enabled": true,
   "baseBranches": [
     "main"
   ],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -114,17 +114,20 @@
     }
   ],
   "assignees": [
-    "kashifest"
-  ],
-  "assigneesSampleSize": 1,
-  "reviewers": [
+    "elfosardo",
```
tuminoid, line 117: Should we drop assignees/reviewers completely, since we use Blunderbuss for reviews these days? Renovate's lists were added when we did not have blunderbuss...

renovate.json
```diff
@@ -2,7 +2,7 @@
   "extends": [
     "config:base"
   ],
-  "enabled": "true",
+  "enabled": true,
```
tuminoid, line 5: Not sure it matters, since it worked with that before, but doesn't hurt to correct it. We can see from other boolean values that they're not quoted.

renovate.json
kashifest: oh yes, true  . Removing it

renovate.json
kashifest: Yes

renovate.json
kashifest: Done

tuminoid (commented):
/approve

kashifest:
/override metal3-centos-e2e-integration-test-release-1-9 metal3-dev-env-integration-test-ubuntu-main

Issue #1499, opened 2025-01-24
CloudProviderEnabled feature in CAPM3 broke dev-env tests


tuminoid:
We merged https://github.com/metal3-io/cluster-api-provider-metal3/pull/2108 yesterday, and it seems we have conflicting values in dev-env now.

```text
fatal: [localhost]: FAILED! => {"changed": false, "cmd": "kubectl apply -f \"/home/metal3ci/tested_repo/tests/roles/run_tests/files/manifests/v1beta1_cluster_centos.yaml\" -n \"metal3\"", "delta": "0:00:00.256850", "end": "2025-01-23 22:59:05.092458", "msg": "non-zero return code", "rc": 1, "start": "2025-01-23 22:59:04.835608", "stderr": "The Metal3Cluster \"test1\" is invalid: spec.cloudProviderEnabled: Invalid value: true: cloudProviderEnabled conflicts the value of noCloudProvider", "stderr_lines": ["The Metal3Cluster \"test1\" is invalid: spec.cloudProviderEnabled: Invalid value: true: cloudProviderEnabled conflicts the value of noCloudProvider"], "stdout": "cluster.cluster.x-k8s.io/test1 created\nippool.ipam.metal3.io/provisioning-pool created\nippool.ipam.metal3.io/externalv4-pool created", "stdout_lines": ["cluster.cluster.x-k8s.io/test1 created", "ippool.ipam.metal3.io/provisioning-pool created", "ippool.ipam.metal3.io/externalv4-pool created"]}
```

/kind bug
/triage accepted
/assign @adilGhaffarDev

adilGhaffarDev:
pr for the fix: https://github.com/metal3-io/cluster-api-provider-metal3/pull/2276

tuminoid:
Fix merged, thank you @adilGhaffarDev for debugging.

/close

PR #1500, opened 2025-01-24
[wip]🐛 Remove noCloudProvider and add cloudProviderEnabled

adilGhaffarDev:
Remove noCloudProvider and add cloudProviderEnabled.
testing
/hold

Fixes https://github.com/metal3-io/metal3-dev-env/issues/1499

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -151,10 +151,11 @@ export BMO_BASE_URL="${BMO_BASE_URL:-metal3-io/baremetal-operator}"
 
 export RUN_LOCAL_IRONIC_SCRIPT="${BMOPATH}/tools/run_local_ironic.sh"
 
+# Nordix:fix-noccloudprovider/adil
 export CAPM3PATH="${CAPM3PATH:-${M3PATH}/cluster-api-provider-metal3}"
-export CAPM3_BASE_URL="${CAPM3_BASE_URL:-metal3-io/cluster-api-provider-metal3}"
-export CAPM3REPO="${CAPM3REPO:-https://github.com/${CAPM3_BASE_URL}}"
-export CAPM3RELEASEBRANCH="${CAPM3RELEASEBRANCH:-main}"
+export CAPM3_BASE_URL="Nordix/cluster-api-provider-metal3"
+export CAPM3REPO="https://github.com/Nordix/cluster-api-provider-metal3"
+export CAPM3RELEASEBRANCH="fix-noccloudprovider/adil"
 
 export USE_IRSO="${USE_IRSO:-false}"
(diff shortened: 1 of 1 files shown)


tests/roles/run_tests/templates/cluster-template-cluster.yaml
```diff
@@ -27,7 +27,7 @@ spec:
   controlPlaneEndpoint:
     host: ${ CLUSTER_APIENDPOINT_HOST }
     port: ${ CLUSTER_APIENDPOINT_PORT }
-  noCloudProvider: true
+  cloudProviderEnabled: false
```
tuminoid, line 30: This however does not change the logic. The error message is about conflicting values. This should not happen when only one value is provided right?

tuminoid (commented):
Question:

tests/roles/run_tests/templates/cluster-template-cluster.yaml
adilGhaffarDev: yes, that is what I am testing, not sure why its conflicting.  I am in the VM where the test is running to check what are the final values of noCloudProvider and cloudProviderEnabled when we apply it. Lets see

adilGhaffarDev: setting only cloudProviderEnabled works fine. I will run periodic ci test again, I need to check what is happening in the failure case.

adilGhaffarDev: in the generated yaml file that is being applied I only see noCloudProvider:
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
```
kind: Metal3Cluster
metadata:
  name: test1
  namespace: metal3
spec:
  controlPlaneEndpoint:
    host: 192.168.111.249
    port: 6443
  noCloudProvider: true
```
I have no clue why it says there is a conflict; something is wrong. I am running tests locally to debug more.

adilGhaffarDev: I have figured it out, it is because we are setting // +kubebuilder:default=true for cloudProviderEnabled, in validation webhook it fails because it gets default value of cloudProviderEnabled instead of nil. 
We can fix this by adding defaulting webhook, and removing // +kubebuilder:default=true , I am working on it. 
One weird thing is this does not happen in tests here : https://github.com/metal3-io/cluster-api-provider-metal3/blob/7aaf7b896c7939776198e9a3d19f2ec5a963c2b0/api/v1beta1/metal3cluster_webhook_test.go#L121 ,

adilGhaffarDev:
/close
it was used for testing, fix is not needed in dev-env but in CAPM3

PR #1464, opened 2024-10-28
🌱 Add tls certs to fakeIPA config

mboukhalfa:
After merging https://github.com/metal3-io/utility-images/pull/26 we can set TLS certs path in fakeIPA config to enable TLS

This needs to wait for sushy-tools image and release with the following patch :https://review.opendev.org/c/openstack/sushy-tools/+/933551

/hold

PR #1501, opened 2025-01-28
Revert "🌱 Add tls certs to fakeIPA config"

mquhuy:
This PR Reverts metal3-io/metal3-dev-env#1464. It has been causing the test to be extremely slow, temporarily disable it until the problem is investigated.


tuminoid (commented):
/approve

PR #1504, opened 2025-02-17
:bug: Pin IPA version

peppi-lotta:
Pin IPA version
Related PR in BMO: https://github.com/metal3-io/baremetal-operator/pull/2264


lentzi90 (commented):
/approve

Sunnatillo (commented):
/lgtm
Thank you.

elfosardo (changes requested):
we should not really point to stable branches but to the corresponding bugfix ones based on the ironic releases

README.md
```diff
@@ -177,11 +177,11 @@ The following table describes which branches are tested for different test trigg
...
-| ----------- | ------------ | -----------  | --------------- | -------------- | ---------- |
-| main        | main         | main         | main            | latest         | latest     |
-| release-1-9 | release-1.9  | release-1.9  | release-0.9     | v0.9.0         | v27.0.0    |
-| release-1-8 | release-1.8  | release-1.8  | release-0.8     | v0.8.0         | v26.0.1    |
-| release-1-7 | release-1.7  | release-1.7  | release-0.6     | v0.6.2         | v24.1.2    |
+| test suffix | CAPM3 branch | IPAM branch  | BMO branch/tag  | Keepalived tag | Ironic tag | IPA branch    |
+| ----------- | ------------ | -----------  | --------------- | -------------- | ---------- | ------------- |
+| main        | main         | main         | main            | latest         | latest     | master        |
+| release-1-9 | release-1.9  | release-1.9  | release-0.9     | v0.9.0         | v27.0.0    | stable/2024.2 |
+| release-1-8 | release-1.8  | release-1.8  | release-0.8     | v0.8.0         | v26.0.1    | stable/2024.2 |
```
elfosardo, line 184: for example this should be "bugfix/10.0"

peppi-lotta: I have added "bugfix/10.0" when ironic is v27. What is the correct branch for IPA when ironic is v26?

Rozzii: or maybe 9.14 IPA for for v26.0.1? I am not sure either if we would not go with the stable then is 9.13  or 9.14 would fit better.
9.13 is the closest to the release of 26.0.0 but there have been a few patches on bugfix/26.00 Ironic branch even as yung as a month old.

elfosardo: bugfx/9.13 is the correct one for ironic 26, there is no bugfix/9.14
9.14 is based on stable/2024.2, which is more recent and could potentally have incompatible code with ironic 26
the bugfix branches are cut at the same time for ironic and IPA, so ideally we would use them in couple

elfosardo: @peppi-lotta ironic-mage 26.0.1 is based on ironic bugfix/26.0
you can check the ironic-image repo to see the correspondency between ironic-image release and ironic version

peppi-lotta: Thank you for confirming what is the correct branch to use here.

README.md
```diff
@@ -177,11 +177,11 @@ The following table describes which branches are tested for different test trigg
...
-| main        | main         | main         | main            | latest         | latest     |
-| release-1-9 | release-1.9  | release-1.9  | release-0.9     | v0.9.0         | v27.0.0    |
-| release-1-8 | release-1.8  | release-1.8  | release-0.8     | v0.8.0         | v26.0.1    |
-| release-1-7 | release-1.7  | release-1.7  | release-0.6     | v0.6.2         | v24.1.2    |
+| test suffix | CAPM3 branch | IPAM branch  | BMO branch/tag  | Keepalived tag | Ironic tag | IPA branch    |
+| ----------- | ------------ | -----------  | --------------- | -------------- | ---------- | ------------- |
+| main        | main         | main         | main            | latest         | latest     | master        |
+| release-1-9 | release-1.9  | release-1.9  | release-0.9     | v0.9.0         | v27.0.0    | bugfix/10.0   |
+| release-1-8 | release-1.8  | release-1.8  | release-0.8     | v0.8.0         | v26.0.1    | stable/2024.2 |
+| release-1-7 | release-1.7  | release-1.7  | release-0.6     | v0.6.2         | v24.1.2    | stable/2024.2 |
```
elfosardo, line 185: this should be 2024.1 :)

lib/common.sh
```diff
@@ -288,6 +288,7 @@ if [[ "${CAPM3RELEASEBRANCH}" = "release-1.7" ]]; then
   export BARE_METAL_OPERATOR_TAG="v0.6.2"
   export KEEPALIVED_TAG="v0.6.2"
   export IRONIC_TAG="v24.1.2"
+  export IPA_BRANCH="stable/2024.1"
```
Rozzii, line 291: Please take into consideration that the IPA is not build from source in dev-env rather pulled from here https://tarballs.opendev.org/openstack/ironic-python-agent/. Please make sure that you specify the "branch" in a way that it is understandable for the IPA downloader.

I would recommend not specifying branches rather full  file names that will be used here https://github.com/metal3-io/ironic-ipa-downloader/blob/main/get-resource.sh#L29 and this way the whole IPA pinning would be more explicit and easier to understand and track.

Rozzii (dismissed):

lib/common.sh
peppi-lotta: All right I get it. I need to write the file name that is based on a branch name instead of the actual branch name. I think the `get-resource.sh` needs a little fixing before this will work because:

The ipa-downloader is not pulling from https://tarballs.opendev.org/openstack/ironic-python-agent/ but https://tarballs.opendev.org/openstack/ironic-python-agent/dib/

The second location currently doesn't have the bugfixes I'm trying to add here. I can override the `IPA_BASEURI` but the files are not named the same way in these two location. The first location doesn't have "IPA_FLAVOR" in the file name and  `get-resource.sh` script seems to be always adding that to the [file name](https://github.com/metal3-io/ironic-ipa-downloader/blob/5ceaad40909c39edc097a78c4622a791e8eee92c/get-resource.sh#L29).

peppi-lotta:
I made this small [PR in ipa-downloader](https://github.com/metal3-io/ironic-ipa-downloader/pull/64) that I think should merge before I make further changes in this PR.

Rozzii (commented):
/approve

Rozzii (commented):
Okay this was already approved so I will lgtm
/lgtm

PR #1507, opened 2025-03-06
:bug: Fix writing a provider to clusterctl.yaml to append not overwrite

peppi-lotta:
`> ` over writes a file `>>`  appends to file. This was over writing some values set in CAPM3 `ci-script.sh`. This will fix e2e test fails related ipreuse :) 

Also this has minor syntax improvement: begin all function the same way.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -492,13 +492,12 @@ update_component_image()
 #
 # Update the clusterctl deployment files to use local repositories
 #
-patch_clusterctl(){
-
+patch_clusterctl()
+{
     pushd "${CAPM3PATH}"
-    mkdir -p "${CAPI_CONFIG_DIR}"
-    touch "${CAPI_CONFIG_DIR}"/clusterctl.yaml
 
-    cat << EOF > "${CAPI_CONFIG_DIR}"/clusterctl.yaml
+    mkdir -p "${CAPI_CONFIG_DIR}"
+    cat << EOF >> "${CAPI_CONFIG_DIR}"/clusterctl.yaml
 providers:
 - name: metal3ipam
   url: https://github.com/metal3-io/ip-address-manager/releases/${IPAMRELEASE}/ipam-components.yaml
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/approve

03_launch_mgmt_cluster.sh
```diff
@@ -498,7 +498,7 @@ patch_clusterctl(){
     mkdir -p "${CAPI_CONFIG_DIR}"
     touch "${CAPI_CONFIG_DIR}"/clusterctl.yaml
 
-    cat << EOF > "${CAPI_CONFIG_DIR}"/clusterctl.yaml
+    cat << EOF >> "${CAPI_CONFIG_DIR}"/clusterctl.yaml
```
tuminoid, line 501: On the lines above L498-499, it is supposedly creating the directory and touching empty file there. Both actions are non-destructive though, so there COULD already be a file (coming from Ansible task I presume).

Do you know if the file is guaranteed to exist, and we could remove the mkdir+touch to make it clearer?

adilGhaffarDev: I don't think we can guarantee this file, its coming from capm3 ci-e2e.sh script but only for some tests.

peppi-lotta: At least `make clean ` [removes](https://github.com/metal3-io/metal3-dev-env/blob/63c7904d38a542b67f9c954d72a023959758f096/host_cleanup.sh#L88) the `.config/cluster-api` where that `clusterctl.yaml` is located. So it is not guaranteed to exist. I think I can remove the L499 but keep L498. I think `>>` will create the file if its missing but won't create directories that are missing in the path.

tuminoid: Thanks. `>>` definitely creates the file if its missing, so let's get rid of that at least.

tuminoid (commented):
/lgtm

Please update the description that the clusterctl.yaml might be populated by ansible for some tests already.

PR #1509, opened 2025-03-10
Change ipam provider name from metal3ipam to metal3

peppi-lotta:
Change provider name from metal3ipam to metal3.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -499,7 +499,7 @@ patch_clusterctl()
     mkdir -p "${CAPI_CONFIG_DIR}"
     cat << EOF >> "${CAPI_CONFIG_DIR}"/clusterctl.yaml
 providers:
-- name: metal3ipam
+- name: metal3
   url: https://github.com/metal3-io/ip-address-manager/releases/${IPAMRELEASE}/ipam-components.yaml
   type: IPAMProvider
 EOF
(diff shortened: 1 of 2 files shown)


tuminoid:
/retitle Change ipam provider name from metal3ipam to metal3

lentzi90 (commented):
Makes sense!
/approve

03_launch_mgmt_cluster.sh
```diff
@@ -499,7 +499,7 @@ patch_clusterctl()
     mkdir -p "${CAPI_CONFIG_DIR}"
     cat << EOF >> "${CAPI_CONFIG_DIR}"/clusterctl.yaml
 providers:
-- name: metal3ipam
+- name: metal3
```
adilGhaffarDev, line 502: We already have `InfrastructureProvider` with the name `metal3`, is it ok to have 2 providers of different types having same name? Have you checked CAPI documentation?

lentzi90: Maybe best to do a quick test? Create the config file as here and then `clusterctl init ...` with both CAPM3 and IPAM. Can you do it @peppi-lotta ?

peppi-lotta: As you can see the [metal3-dev-env-integration-test-ubuntu-main](https://jenkins.nordix.org/view/Metal3/job/metal3-dev-env-integration-test-ubuntu-main/243/) test passed. This means it was able to deploy both ipam and infrastructure providers with name `metal3`. 

We want to get this IPAM as an official provider eventually [here](https://github.com/kubernetes-sigs/cluster-api/blob/bfb2649fba918e05d6012dbb3599543017554898/cmd/clusterctl/client/config/providers_client.go#L103). IPAM providers and infrasturcture providers have their own lists so I don't see there would be a problem. 

At least nutanix has both infrastucture and Ipam providers by the same name in this file: https://github.com/kubernetes-sigs/cluster-api/blob/main/cmd/clusterctl/client/config/providers_client.go

lentzi90: Excellent, thanks!

PR #1510, opened 2025-03-19
enable tpm simulation for virtual machines

Rozzii:
This commit:
 - Appends the libvirt VM definition xml templates to include tpm simulation.

This commit is needed to enable the testing of use-cases where tpm features e.g. disk encryption, remote secure boot attestation, random hash generation, measured boot etc are used.

### DIFF
--- a/vm-setup/roles/libvirt/templates/baremetalvm.xml.j2
+++ b/vm-setup/roles/libvirt/templates/baremetalvm.xml.j2
@@ -44,6 +44,9 @@
   <on_reboot>restart</on_reboot>
   <on_crash>restart</on_crash>
   <devices>
+    <tpm model="tpm-tis">
+      <backend type="emulator" version="2.0"/>
+    </tpm>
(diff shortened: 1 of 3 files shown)


kashifest (commented):
/approve

Rozzii:
looks like the swtpm package is missing from the ubuntu based ci image so I will add it.

tuminoid (commented):
/lgtm

Rozzii:
I am checking, even though it runs the swtpm install QEMU complains that swtpm ins not in $PATH

Rozzii:
I will take a look when I have some time, but it is strange that it works on centos but not on ubuntu when I have wrote this on ubuntu.

Rozzii:
> > /remove-lifecycle stale I will handle this, it is not urgent yet.
> 
> Btw, I have TPM stuff in my Secure boot POC PR too, which is verified to work, we can take that from there too?

I will take your's if I can't make the "bult in" work.

PR #1513, opened 2025-03-19
add option to toggle dev-env host package updates

Rozzii:
This PR:
 - Introduces `nodep` make target to skip redundant dependency installation
 - Moves container runtime cleanup to 02 config script
 - Moves the GO related bashrc check to 02 script

This commit is needed because on persistent hosts (e.g. developer's
workstation) it might be undesirable to update the local system packages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 all: install_requirements configure_host launch_mgmt_cluster verify
 
+nodep: configure_host launch_mgmt_cluster verify
+
 ci_run: configure_host launch_mgmt_cluster verify
 
 install_requirements:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,5 +1,7 @@
 all: install_requirements configure_host launch_mgmt_cluster verify
 
+nodep: configure_host launch_mgmt_cluster verify
```
Rozzii, line 3: This is right now the same as the ci_run but I intend to keep them separate because they might diverge in the future.

smoshiur1237 (commented):
/lgtm

PR #1520, opened 2025-04-16
Use Ansible 8 in CentOS Stream 9

elfosardo:


kashifest (commented):
/approve

lib/common.sh
```diff
@@ -399,11 +399,11 @@ export TILT_SHA256="${TILT_SHA256:-b30ebbba68d4fd04f8afa11efc439515241dbcc2582ea
...
-# Ansible 7.0.0 or newer requires python 3.10+
+# Ansible 7.0.0 or newer requires python 3.9+
 # TODO: Ansible pinning
 if [[ "${DISTRO}" = "ubuntu24" ]]; then
     export ANSIBLE_VERSION="${ANSIBLE_VERSION:-10.6.0}"
-elif [[ "${DISTRO}" = "ubuntu22" ]]; then
+elif [[ "${DISTRO}" = "ubuntu22" ]] || [[ "${DISTRO}" = "centos9" ]]; then
     export ANSIBLE_VERSION="${ANSIBLE_VERSION:-8.0.0}"
 else
     export ANSIBLE_VERSION="${ANSIBLE_VERSION:-6.7.0}"
```
kashifest, line 409: is there any more use case for this older version i.e. 6.7.0?

elfosardo: AFAIK nothing that we currently support
I did not remove it just because  I didn't want to include the change here, but I think it can be removed

PR #1528, opened 2025-05-05
Add redfish SSL support

slintes:
Hello,

we are experimenting with using fence agents on cluster nodes. `fence_redfish` only supports `https` connections, but it's not possible to configure sushy to provide a SSL connection yet.

After discusion on this PR, it was decided to not make SSL support optional, but always use it, because that's what real world BMCs look like. Since the certificate is self signed, the `redfish_verify_ca` option is added and set to `False`

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -391,9 +391,11 @@ make_bm_hosts()
     mkdir -p "${WORKING_DIR}/bmhs"
 
     local i=0
-    while read -r name address user password mac; do
+    while read -r name address user password mac verify_ca; do
+        disableCertificateVerification=$([ "${verify_ca}" == "False" ] && echo -n "true" || echo -n "false")
         go run "${BMOPATH}"/cmd/make-bm-worker/main.go \
             -address "${address}" \
+            -disableCertificateVerification "${disableCertificateVerification}" \
             -password "${password}" \
             -user "${user}" \
             -boot-mac "${mac}" \
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -393,12 +393,12 @@ make_bm_hosts()
     local i=0
     while read -r name address user password mac verify_ca; do
         go run "${BMOPATH}"/cmd/make-bm-worker/main.go \
-            -address "${address}" \
-            -disableCertificateVerification "$(get_disableCertificateVerification_from_verify_ca "${verify_ca}")" \
-            -password "${password}" \
-            -user "${user}" \
-            -boot-mac "${mac}" \
-            -boot-mode "${BOOT_MODE}" \
+            -address="${address}" \
+            -disableCertificateVerification="$(get_disableCertificateVerification_from_verify_ca "${verify_ca}")" \
+            -password="${password}" \
+            -user="${user}" \
+            -boot-mac="${mac}" \
+            -boot-mode="${BOOT_MODE}" \
             "${name}" | \
             tee "${WORKING_DIR}/bmhs/node_${i}.yaml" >> "${WORKING_DIR}/bmhosts_crs.yaml"
         i=$((i+1))
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -392,10 +392,9 @@ make_bm_hosts()
 
     local i=0
     while read -r name address user password mac verify_ca; do
-        disableCertificateVerification=$([ "${verify_ca}" == "False" ] && echo -n "true" || echo -n "false")
         go run "${BMOPATH}"/cmd/make-bm-worker/main.go \
             -address "${address}" \
-            -disableCertificateVerification "${disableCertificateVerification}" \
+            -disableCertificateVerification "$(get_disableCertificateVerification_from_verify_ca "${verify_ca}")" \
             -password "${password}" \
             -user "${user}" \
             -boot-mac "${mac}" \
--- a/03_launch_mgmt_cluster_pre_1_10.sh
+++ b/03_launch_mgmt_cluster_pre_1_10.sh
@@ -385,10 +385,9 @@ make_bm_hosts()
 
     local i=0
     while read -r name address user password mac verify_ca; do
-        disableCertificateVerification=$([ "${verify_ca}" == "False" ] && echo -n "true" || echo -n "false")
         go run "${BMOPATH}"/cmd/make-bm-worker/main.go \
             -address "${address}" \
-            -disableCertificateVerification "${disableCertificateVerification}" \
+            -disableCertificateVerification "$(get_disableCertificateVerification_from_verify_ca "${verify_ca}")" \
             -password "${password}" \
             -user "${user}" \
             -boot-mac "${mac}" \
--- a/03_launch_mgmt_cluster_pre_1_10.sh
+++ b/03_launch_mgmt_cluster_pre_1_10.sh
@@ -386,12 +386,12 @@ make_bm_hosts()
     local i=0
     while read -r name address user password mac verify_ca; do
         go run "${BMOPATH}"/cmd/make-bm-worker/main.go \
-            -address "${address}" \
-            -disableCertificateVerification "$(get_disableCertificateVerification_from_verify_ca "${verify_ca}")" \
-            -password "${password}" \
-            -user "${user}" \
-            -boot-mac "${mac}" \
-            -boot-mode "${BOOT_MODE}" \
+            -address="${address}" \
+            -disableCertificateVerification="$(get_disableCertificateVerification_from_verify_ca "${verify_ca}")" \
+            -password="${password}" \
+            -user="${user}" \
+            -boot-mac="${mac}" \
+            -boot-mode="${BOOT_MODE}" \
             "${name}" | \
             tee "${WORKING_DIR}/bmhs/node_${i}.yaml" >> "${WORKING_DIR}/bmhosts_crs.yaml"
         i=$((i+1))
--- a/03_launch_mgmt_cluster_pre_1_10.sh
+++ b/03_launch_mgmt_cluster_pre_1_10.sh
@@ -384,9 +384,11 @@ make_bm_hosts()
     mkdir -p "${WORKING_DIR}/bmhs"
 
     local i=0
-    while read -r name address user password mac; do
+    while read -r name address user password mac verify_ca; do
+        disableCertificateVerification=$([ "${verify_ca}" == "False" ] && echo -n "true" || echo -n "false")
         go run "${BMOPATH}"/cmd/make-bm-worker/main.go \
             -address "${address}" \
+            -disableCertificateVerification "${disableCertificateVerification}" \
             -password "${password}" \
             -user "${user}" \
             -boot-mac "${mac}" \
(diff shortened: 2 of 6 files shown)


vm-setup/roles/common/defaults/main.yml
```diff
@@ -114,3 +114,6 @@ networks_nated: "{{ provisioning_network_nat + external_network }}"
 networks_nat_only: "{{ lookup('env', 'ENABLE_NATED_PROVISIONING_NETWORK') | bool }}"
 # Placeholder for the eventually selected network composition
 networks: "{{ networks_nated if networks_nat_only else networks_mixed }}"
+
+# Create a SSL certificate and configure sushy to use it
+sushy_use_ssl: "False"
```
dtantsur, line 119: Is there any reason not to use TLS by default?

(btw, I'd rather call it "tls" than "ssl")

dtantsur: Most real-world BMC's use TLS, and in most cases we need to disable the validation, unfortunately. I'd just switch the whole thing to TLS by default (bonus: no need to figure out the right True/true values)

03_launch_mgmt_cluster.sh
```diff
@@ -391,9 +391,11 @@ make_bm_hosts()
     mkdir -p "${WORKING_DIR}/bmhs"
 
     local i=0
-    while read -r name address user password mac; do
+    while read -r name address user password mac verify_ca; do
+        disableCertificateVerification=$([ "${verify_ca}" == "False" ] && echo -n "true" || echo -n "false")
```
tuminoid, line 395: Bash doesn't have proper reliable ternary operator, please use regular if.

03_launch_mgmt_cluster_pre_1_10.sh
```diff
@@ -384,9 +384,11 @@ make_bm_hosts()
     mkdir -p "${WORKING_DIR}/bmhs"
 
     local i=0
-    while read -r name address user password mac; do
+    while read -r name address user password mac verify_ca; do
+        disableCertificateVerification=$([ "${verify_ca}" == "False" ] && echo -n "true" || echo -n "false")
```
tuminoid, line 388: same

04_verify.sh
```diff
@@ -49,6 +50,10 @@ check_bm_hosts() {
       RESULT_STR="${NAME} Baremetalhost address correct"
       equals "$(echo "${BARE_METAL_HOST}" | jq -r '.spec.bmc.address')" "${ADDRESS}"
 
+      RESULT_STR="${NAME} Baremetalhost disableCertificateVerification correct"
+      disableCertificateVerification=$([ "${VERIFY_CA}" == "False" ] && echo -n "true" || echo -n "false")
```
tuminoid, line 54: same

lib/common.sh
```diff
@@ -499,11 +499,12 @@ list_nodes() {
...
            user:.driver_info.username,
            password:.driver_info.password,
+           verify_ca:.driver_info.redfish_verify_ca,
            mac: .ports[0].address
-           } |
+           } | if .verify_ca == null then .verify_ca="True" else . end |
            .name + " " +
            .address + " " +
-           .user + " " + .password + " " + .mac' \
+           .user + " " + .password + " " + .mac + " " + .verify_ca' \
```
tuminoid, line 507: Reformat this line to be like the `.name`  and `.address` lines for readability.

tuminoid:
> @tuminoid thanks for your review 🙂 I addressed your feedback.

Looks much cleaner, thank you.

> I'm having trouble to understand why CI fails to apply the BMH CRs... the arguments to `make-bm-worker/main.go` in the console log look good to me, and runing this in openshift dev-scripts is working as well. Is there any chance to see the CRs, or the logs of the baremetal operator? (see https://jenkins.nordix.org/view/Metal3/job/metal3-dev-env-integration-test-ubuntu-main/282/, latest run is running into the same issue 🤔 )

We have some issues with CI logging right now, apologies. Those would probably help debug the issue. We'll get it sorted out soon.

tuminoid:
> finally green CI 🥳
> 
> @tuminoid @dtantsur do you want me to keep the last commit in? It prints the bmh manifest before applying, prints the result of the apply command, and retries once only instead of for 3 hours until the test times out. I understand the while loop was needed before a wait for the BMO rollout was added. Example output when apply fails:
> 
> ```
> [2025-05-13T06:42:14.727Z] + echo 'bmhosts_crs.yaml is applying:'
> [2025-05-13T06:42:14.727Z] bmhosts_crs.yaml is applying:
> [2025-05-13T06:42:14.727Z] + cat /opt/metal3-dev-env/bmhosts_crs.yaml
> [2025-05-13T06:42:14.727Z] + kubectl apply -f /opt/metal3-dev-env/bmhosts_crs.yaml -n metal3
> [2025-05-13T06:42:14.727Z] error: no objects passed to apply
> [2025-05-13T06:42:14.727Z] + echo 'failed, will retry once in 20s'
> [2025-05-13T06:42:14.727Z] failed, will retry once in 20s
> [2025-05-13T06:42:14.727Z] + sleep 20
> [2025-05-13T06:42:38.459Z] + kubectl apply -f /opt/metal3-dev-env/bmhosts_crs.yaml -n metal3
> [2025-05-13T06:42:38.460Z] error: no objects passed to apply
> [2025-05-13T06:42:38.460Z] + echo 'failed to apply bmhosts_crs.yaml'
> [2025-05-13T06:42:38.460Z] failed to apply bmhosts_crs.yaml
> [2025-05-13T06:42:38.460Z] + exit 1
> ```

I agree that interating over the empty file for 3 hours is not productive.

Printing out the bmhosts is not helpful here at least, as it those files are empty. If that seems to be the most common symptom, maybe let's check for that explicitly and just bail out. The empty file is not magically going to have content, if it didn't in the first time around.

Alternatively, I'd maybe move the `list_nodes | make_bm_hosts` inside the retry loop and run in once per 60s for some more reasonable total timeout? This would help in case nodes are slow to appear.

WDYT?

tuminoid:
> > Printing out the bmhosts is not helpful here at least, as it those files are empty.
> 
> sure thing, but it might help in case they are malformed. They were empty here because of a malformed call to `make-bm-worker/main.go`, which exited without output.

OK, now I see what you mean by that. I'd still bail out in case of empty file. It will not fix itself by waiting in the loop, so saving 3 hrs in those cases is good. 

> > If that seems to be the most common symptom, maybe let's check for that explicitly and just bail out. The empty file is not magically going to have content, if it didn't in the first time around. Alternatively, I'd maybe move the list_nodes | make_bm_hosts inside the retry loop and run in once per 60s for some more reasonable total timeout? This would help in case nodes are slow to appear.
> 
> `list_nodes` wasn't an issue. Not sure if it's worth additional effort... 🤔

I'm not sure if we can give up the loop completely. How about making it try 10 times with 1 min interval and then quitting if still failing? 20s is not enough in case failure was about something not being completed in background, but 10 mins surely means it is never going to come up.

03_launch_mgmt_cluster.sh
```diff
@@ -412,15 +413,34 @@ apply_bm_hosts()
...
-        echo "bmhosts_crs.yaml is successfully applied"
-    fi
+    local RETRY=10
+    while [[ "$RETRY" -gt 0 ]]; do
+      echo "bmhosts_crs.yaml is applying"
+      # check if we have nodes
+      if [[ -n "$(list_nodes)" ]]; then
+        list_nodes | make_bm_hosts
+        # check if we have a not empty manifests file
+        if [[ -e "${WORKING_DIR}/bmhosts_crs.yaml" ]] && [[ -n "$(cat "${WORKING_DIR}/bmhosts_crs.yaml")" ]]; then
```
tuminoid, line 423: This is solved by `-s` test. Catting in subshell is no-no.

03_launch_mgmt_cluster.sh
```diff
@@ -412,15 +413,34 @@ apply_bm_hosts()
...
+      fi
+      echo "retrying in 1 minute"
+      sleep 60
+      (( RETRY-=1 ))
+    done
+
     popd
+
+    if [[ "${RETRY}" -eq 0 ]]
+    then
```
tuminoid, line 440:
```suggestion
    if [[ "${RETRY}" -eq 0 ]]; then
```

03_launch_mgmt_cluster_pre_1_10.sh
```diff
@@ -405,15 +406,34 @@ apply_bm_hosts()
...
 
-    list_nodes | make_bm_hosts
-    if [[ -n "$(list_nodes)" ]]; then
-        echo "bmhosts_crs.yaml is applying"
-        while ! kubectl apply -f "${WORKING_DIR}/bmhosts_crs.yaml" -n "${namespace}" &>/dev/null; do
-            sleep 3
-        done
-        echo "bmhosts_crs.yaml is successfully applied"
-    fi
+    local RETRY=10
```
tuminoid, line 409: Same issues as the 03 script .

tuminoid (commented):
Looks much more reasonable, let's just simplify the code . Avoid nesting if's and doing subshell calls for no benefit.

03_launch_mgmt_cluster.sh
```diff
@@ -412,15 +413,28 @@ apply_bm_hosts()
...
+        echo "bmhosts_crs.yaml does not exist or is empty"
+      fi
+      echo "retrying in 1 minute"
+      sleep 60
+      (( RETRY-=1 ))
+    done
+
     popd
+
+    if [[ "${RETRY}" -eq 0 ]];then
```
tuminoid, line 434:
```suggestion
    if [[ "${RETRY}" -eq 0 ]]; then
```

03_launch_mgmt_cluster_pre_1_10.sh
```diff
@@ -405,15 +406,28 @@ apply_bm_hosts()
...
+        echo "bmhosts_crs.yaml does not exist or is empty"
+      fi
+      echo "retrying in 1 minute"
+      sleep 60
+      (( RETRY-=1 ))
+    done
+
     popd
+
+    if [[ "${RETRY}" -eq 0 ]];then
```
tuminoid, line 427:
```suggestion
    if [[ "${RETRY}" -eq 0 ]]; then
```

03_launch_mgmt_cluster.sh
```diff
@@ -412,15 +413,28 @@ apply_bm_hosts()
...
-            sleep 3
-        done
-        echo "bmhosts_crs.yaml is successfully applied"
-    fi
+    local RETRY=10
+    while [[ "${RETRY}" -gt 0 ]]; do
+      echo "bmhosts_crs.yaml is applying"
+      list_nodes | make_bm_hosts
+      # check if we have a not empty manifests file
+      if [[ -s "${WORKING_DIR}/bmhosts_crs.yaml" ]]; then
```
tuminoid, line 421: We're repeating `${WORKING_DIR}/bmhosts_crs.yaml` three times in the function, put it in a variable.

tuminoid (commented):
Couple tiny nits, but this is again much simplified. Easier to read and maintain going forward.

I'm sorry to have sidelined you a bit, but thanks for fixing this!

tuminoid (commented):
/lgtm
/cc @dtantsur  @elfosardo

PR #1531, opened 2025-05-13
Add check for ironic deployments to fix the centOS test failure when SKIP_APPLY_BMH enabled

smoshiur1237:
We are facing a situation where centOS(minikube) test failed when SKIP_APPLY_BMH variable is enabled. But same test was passing in Ubuntu test.  Here we are skipping BMH check, as it will be applied later from CAMP3 side. Test didn't provide enough time from ironic pod to be in running state and  failed during ironic endpoint check in verify script. So this PR will add checks for deployment and pods before the check endpoint check. It will ensure running ironic deployment and pod.

### DIFF
--- a/04_verify.sh
+++ b/04_verify.sh
@@ -74,7 +74,7 @@ check_bm_hosts() {
       equals "$(echo "${CRED_SECRET}" | jq -r '.username' | \
         base64 --decode)" "${USER}"
     fi
-  
+
     # Verify the VM was created
     RESULT_STR="${NAME} Baremetalhost VM exist"
     echo "${BARE_METAL_VMS} "| grep -w "${BARE_METAL_VMNAME}"  > /dev/null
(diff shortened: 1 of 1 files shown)


04_verify.sh
```diff
@@ -305,6 +305,11 @@ for container in ${EXPTD_CONTAINERS}; do
   iterate check_container "$container"
 done
 
+# Add delay before checking the ironic endpoint in Centos(minikube) test
+# This will provide extra time for ironic pod to be in running state
+if [[ "${SKIP_APPLY_BMH:-false}" == "true" ]] && [[ "${EPHEMERAL_CLUSTER}" == "minikube" ]]; then
+  sleep 300
```
adilGhaffarDev, line 311: Can we check the ironic pod status with a timeout instead of `sleep 300`? So we don't have to always wait for 300 and progress further if the pod is ready earlier.

04_verify.sh
adilGhaffarDev, line 1: /lgtm

smoshiur1237:
@kashifest please check, needs approval

kashifest (commented):
/approve

PR #1532, opened 2025-05-13
Improve applying BMHs

slintes:
Splitted off https://github.com/metal3-io/metal3-dev-env/pull/1528

- more sanity checks
- better debugging
- don't retry forever

### DIFF
--- a/03_launch_mgmt_cluster_pre_1_10.sh
+++ b/03_launch_mgmt_cluster_pre_1_10.sh
@@ -405,15 +405,29 @@ apply_bm_hosts()
     local namespace="$1"
     pushd "${BMOPATH}"
 
-    list_nodes | make_bm_hosts
-    if [[ -n "$(list_nodes)" ]]; then
-        echo "bmhosts_crs.yaml is applying"
-        while ! kubectl apply -f "${WORKING_DIR}/bmhosts_crs.yaml" -n "${namespace}" &>/dev/null; do
-            sleep 3
-        done
-        echo "bmhosts_crs.yaml is successfully applied"
-    fi
+    local RETRY=10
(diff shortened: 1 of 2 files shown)


03_launch_mgmt_cluster_pre_1_10.sh
```diff
@@ -405,15 +405,29 @@ apply_bm_hosts()
...
+      # check if we have a not empty manifests file
+      local BMH_FILE="${WORKING_DIR}/bmhosts_crs.yaml"
+      if [[ -s "${BMH_FILE}" ]]; then
+        cat "${BMH_FILE}"
+        kubectl apply -f "${BMH_FILE}" -n "${namespace}" && break
+      else
+        echo "bmhosts_crs.yaml does not exist or is empty"
+      fi
+      echo "retrying in 1 minute"
+      sleep 60
```
Sunnatillo, line 421: LGTM. 
One thing to clarify. Why sleep time is 60 seconds?  How about making sleep time less and RETRY more?

tuminoid: It makes very little sense to do forever looping with 3s interval. Most likely it is going to work first time around, or not at all in case of issues anyways. This now regenerates the crs before retrying etc, so it has real chance of actually getting applied if it fails the first time around, and losing some seconds here is meaningless IMO.

tuminoid (commented):
/approve

03_launch_mgmt_cluster_pre_1_10.sh
Sunnatillo: thank you for explaining

Sunnatillo (commented):
/lgtm

PR #1533, opened 2025-05-15
Remove unused script from verify file

smoshiur1237:
The PR will remove the unused script from verify script where check for pods function was not used. The alternative we use is the Entity check where it checks the deployment with ready replicas with expected replicas.


tuminoid (commented):
/approve

PR #1539, opened 2025-05-27
Use nordix artifactory proxy to download IPA

kashifest:
Lately, we have seen a lot of delay, flakes and failures while downloadind IPA from opendev upstream. Now we have setup a proxy in Nordix rtfactory to proxy the IPA from upstream and use it in our tests in a faster way when cached.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -133,7 +133,7 @@ fi
 
 # shellcheck disable=SC2034
 # USE_LOCAL_IPA and IPA_DOWNLOAD_ENABLED also have effect on BMO repo
-export IPA_BASEURI="${IPA_BASEURI:-https://tarballs.opendev.org/openstack/ironic-python-agent/dib}"
+export IPA_BASEURI="${IPA_BASEURI:-https://artifactory.nordix.org/artifactory/openstack-remote-cache/ironic-python-agent/dib}"
 export IPA_BRANCH="${IPA_BRANCH:-master}"
 export IPA_FLAVOR="${IPA_FLAVOR:-centos9}"
 export USE_LOCAL_IPA="${USE_LOCAL_IPA:-false}"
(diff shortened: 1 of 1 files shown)


lib/common.sh
```diff
@@ -133,7 +133,7 @@ fi
 
 # shellcheck disable=SC2034
 # USE_LOCAL_IPA and IPA_DOWNLOAD_ENABLED also have effect on BMO repo
-export IPA_BASEURI="${IPA_BASEURI:-https://tarballs.opendev.org/openstack/ironic-python-agent/dib}"
+export IPA_BASEURI="${IPA_BASEURI:-https://artifactory.nordix.org/artifactory/openstack-remote-cache/ironic-python-agent/dib}"
```
lentzi90, line 136: I wonder if we should rather do this in project-infra? If we change it here it affects anyone using metal3-dev-env if they don't override. If we set another value in project-infra, it would be only for our CI. This is how we do it for the container registries also:
https://github.com/metal3-io/project-infra/blob/2672f317e5f958867614d1fc073b95e6dc5a77f3/jenkins/scripts/integration_test_env.sh#L64-L66

kashifest: True, this is defaulting here so yeah if someone doesn't want to use it they must override here, which is fine in my opinion. I can move it to project infra as well since that is where the CI vars should be coming from so I see your point. Allthough I dont like the way we do it in project infra either, IMO in project-infra we should just export it not default it anymore.

lentzi90: Hmm that is a good question. Perhaps it should go in [here](https://github.com/metal3-io/cluster-api-provider-metal3/blob/main/scripts/environment.sh) for CAPM3?
My main concern is if we default it to the Nordix proxy, then probably also other people will use it and that is not what it is intended for

tuminoid: I agree with Lennart, its a valid point. e2e setup is just cringy right now, so might require changing both project-infra and CAPM3, we do want to apply it to CAPM3 e2e but also dev-env e2e.

PR #1536, opened 2025-05-21
Add check for Ironic deployment in Centos test

smoshiur1237:
We have seen several tests were failing where Ironic deployment was not in ready state. This PR will add check for ironic deployment in Centos test.

PR #1540, opened 2025-05-28
Update verify script to remove duplicate check

smoshiur1237:
We have added check for ironic deployment in  centos tests in the following [ merged PR](https://github.com/metal3-io/metal3-dev-env/pull/1536). As it is having check, doesn't need the same check during skip bmh action in CAPM3 e2e basic integration test. So this PR will remove the duplicate check in the verify script

### DIFF
--- a/04_verify.sh
+++ b/04_verify.sh
@@ -254,18 +254,7 @@ done
 echo ""
 
 # Verify v1beta1 Operators, Deployments, Replicasets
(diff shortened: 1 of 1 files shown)


tuminoid:
> Just testing this: /cc @github

Yeah this isn't freebie for us.

Sunnatillo (commented):
/lgtm

PR #1544, opened 2025-06-16
IPv6 envvar and ansible fixes

nuhakala:
Small fixes to pave the way to IPv6 only setup.

- Fix few environment variable definitions
- Add new environment variable for Docker network
- Fix variables in Ansible
- Add libvirt network hooks to disable dad with IPv6


lib/network.sh
```diff
@@ -74,8 +74,8 @@ export CLUSTER_BARE_METAL_PROVISIONER_IP
 
 # shellcheck disable=SC2153
 if [[ "$BARE_METAL_PROVISIONER_IP" = *":"* ]]; then
-  export BARE_METAL_RPOVISIONER_URL_HOST="[$BARE_METAL_PROVISIONER_IP]"
-  export CLUSTER_BARE_METAL_PROVISIONER_URL_HOST="[$CLUSTER_BARE_METAL_PROVISIONER_IP]"
+  export BARE_METAL_PROVISIONER_URL_HOST="[$BARE_METAL_PROVISIONER_IP]"
```
tuminoid:
```suggestion
  export BARE_METAL_PROVISIONER_URL_HOST="[${BARE_METAL_PROVISIONER_IP}]"
```
optional: While touching the line, might fix it as well.

lib/network.sh
```diff
@@ -74,8 +74,8 @@ export CLUSTER_BARE_METAL_PROVISIONER_IP
...
+  export CLUSTER_BARE_METAL_PROVISIONER_HOST="[$CLUSTER_BARE_METAL_PROVISIONER_IP]"
```
tuminoid:
```suggestion
  export CLUSTER_BARE_METAL_PROVISIONER_HOST="[${CLUSTER_BARE_METAL_PROVISIONER_IP}]"
```
Same

vars.md
```diff
@@ -11,6 +11,7 @@ assured that they are persisted.
 
 | Name | Option | Allowed values | Default |
 | :------ | :------- | :--------------- | :-------- |
+| DOCKER_IPV6_INTERNALLY | Choose whether Docker will use IPv6 internally. | "true", "false" | false |
```
tuminoid:
```suggestion
| DOCKER_USE_IPV6_INTERNALLY | Choose whether Docker will use IPv6 internally. | "true", "false" | false |
```
Maybe more understandable variable name.

vm-setup/roles/firewall/defaults/main.yml
```diff
@@ -1,4 +1,4 @@
-use_firewalld: true
+use_firewalld: "{{ lookup('env', 'USE_FIREWALLD') | default(true, true) }}"
```
tuminoid, line 1: Q: This is set in lib/common.sh based on the OS, and is not user-configurable.  This is thus fixing a bug in firewall setup, where firewalld was used though it shouldn't have been?

vm-setup/roles/libvirt/templates/libvirt_hook.sh.j2
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
tuminoid:
```suggestion
#!/usr/bin/env bash
```

vm-setup/roles/libvirt/templates/libvirt_hook.sh.j2
```diff
@@ -0,0 +1,6 @@
...
+
+if [[ "$1" = "{{ item.name }}" && "$2" = "started" && "$3" = "begin" ]]; then
```
tuminoid:
```suggestion
if [[ "$1" = "{{ item.name }}" ]] && [[ "$2" = "started" ]] && [[ "$3" = "begin" ]]; then
```

tuminoid (commented):
Some questions and minor nits.

vm-setup/roles/firewall/defaults/main.yml
Rozzii: It is exported in lib/common.sh , I think it makes it much cleaner that we reference the env variable that we have already set in common.sh, maybe if the bug gets resolved it will also make it easier to remove the config from everywhere.

TLDR, I would like to see this env lookup stay.

tuminoid: I have no hard objections to the lookup, I'm just asking why can't we fix this by limiting this firewalld ansible role to only run on centos/rhel as it should be based on lib/common.sh?

Rozzii: I have now checked and it was already limited via the same logic as this "use_firewalld" was toggled in 02_configure_host.sh L294.

From practical perspective the only difference is that we are toggling the feature based on the firewalld specific variable not the more generic "DISTRO". I like this method more because it makes it easier to make it compatible with other distros if needed and provides a dedicated toggle switch for the firewall config.

I don't feel like it worth it to remove this firewalld specific env var .

tuminoid: We can have it this way as well, it is just weird as we have distro specific roles for a reason.

lib/common.sh
```diff
@@ -627,7 +627,7 @@ differs(){
 #
 remove_ironic_containers() {
   #shellcheck disable=SC2015
-  for name in ipa-downloader vbmc sushy-tools httpd-infra ipxe-builder; do
+  for name in ipa-downloader vbmc sushy-tools httpd-infra ipxe-builder registry; do
```
nuhakala, line 630: Not sure why this has not been done earlier. This container is started in 02 script (line 315) and its status is checked there (whether it is exited or exists at all). However, with these modifications it needs to be removed, because it will cause error if the IP stack version is changed during development but the container is not removed. In this situation, the script will fail, because the container is bound to wrong IP-address.

So, should I also simplify the container startup in 02?

tuminoid: I'd say it is not removed by default to preserve images for caching purposes between manual runs. In CI it is useless of course as it is starting from scratch anyways. 

If 02 change is not relevant for this feature, lets do it in another PR.

03_launch_mgmt_cluster.sh
```diff
@@ -624,12 +624,15 @@ create_clouds_yaml()
 #
 launch_kind()
 {
+    # If registry is IPv6 address, '[', ']' are not allowed
+	local registry
```
elfosardo: nit: fix indent

lib/network.sh
```diff
@@ -74,11 +74,11 @@ export CLUSTER_BARE_METAL_PROVISIONER_IP
 
 # shellcheck disable=SC2153
 if [[ "$BARE_METAL_PROVISIONER_IP" = *":"* ]]; then
-  export BARE_METAL_RPOVISIONER_URL_HOST="[$BARE_METAL_PROVISIONER_IP]"
-  export CLUSTER_BARE_METAL_PROVISIONER_URL_HOST="[$CLUSTER_BARE_METAL_PROVISIONER_IP]"
+  export BARE_METAL_PROVISIONER_URL_HOST="[${BARE_METAL_PROVISIONER_IP}]"
+  export CLUSTER_BARE_METAL_PROVISIONER_HOST="[${CLUSTER_BARE_METAL_PROVISIONER_IP}]"
```
elfosardo, line 78: why changing the variable name?

lib/network.sh
```diff
@@ -173,6 +173,7 @@ fi
 if [[ -n "${EXTERNAL_SUBNET_V6}" ]]; then
   prefixlen EXTERNAL_SUBNET_V6_PREFIX "$EXTERNAL_SUBNET_V6"
   export EXTERNAL_SUBNET_V6_PREFIX
+  EXTERNAL_SUBNET_V6_HOST="${EXTERNAL_SUBNET_V6_HOST:-}"
```
elfosardo: this looks like a new variable, should we document it ?

lib/network.sh
nuhakala: At some point I got error that `CLUSTER_BARE_METAL_PROVISIONER_HOST` was not defined. The reason was, that in the IPv6 branch of the condition was the variable `CLUSTER_BARE_METAL_PROVISIONER_URL_HOST` instead of `CLUSTER_BARE_METAL_PROVISIONER_HOST`.

I also grepped through the folder and found no reference to `CLUSTER_BARE_METAL_PROVISIONER_URL_HOST` so figured out it can safely be renamed.

lib/network.sh
nuhakala: It is not exactly a new variable. I just had to add definition like this because the script was complaining that cannot use variables that are not initialized. Furthermore, this variable is assigned a value few lines below with call to `network_address` function, and the value is derived from variable `EXTERNAL_SUBNET_V6` so I would assume that user is not supposed to be setting the value of this and hence no documentation. Maybe @tuminoid or @Rozzii have more information?

lib/network.sh
```diff
@@ -173,6 +173,7 @@ fi
...
   if [[ -z "${EXTERNAL_SUBNET_V6_HOST}" ]]; then
```
tuminoid:
```suggestion
  if [[ -z "${EXTERNAL_SUBNET_V6_HOST:-}" ]]; then
```

lib/network.sh
tuminoid: Instead of setting it empty, we could just amend the L177 as mentioned to achieve the same.

lib/network.sh
Rozzii: Please check BMO repo also, just to make sure CLUSTER_BARE_METAL_PROVISIONER_URL is not used there either as part of the BMO deploy script or its dependencies.

lib/network.sh
Rozzii: Let's go with the syntax suggested by @tuminoid

lib/network.sh
nuhakala: Not used in BMO either. Tbh, it looked like someone had modified ipv4 branch of the if clause and then forgot to apply the same modification to ipv6 branch as well, because these variables look like they are needed regardless of ip version.

Rozzii:
The test has succeeded, but the pipeline has failed because of the known "nohup" issue on centos.
/override metal3-centos-e2e-integration-test-release-1-10

PR #1545, opened 2025-06-19
Get NIC name from bmh for vrrp config

Sunnatillo:
This PR changes hardcoded interface names in vrrp config and makes to get interface names from bmh


smoshiur1237 (commented):
/approve

PR #1546, opened 2025-06-30
Partial revert of getting NIC name from bmh for vrrp config

Sunnatillo:
NIC names are populated differently in centos and ubuntu. 
NICs names in bmh and provisioned os are not same. It is related to "net.ifnames" in kernel. On IPA it is enabled, for centos node image the config is disabled. Thus causing the difference. 

This PR fixes bml failure. Proper fix will be done later


Sunnatillo:
/override metal3-centos-e2e-integration-test-release-1-10

Failed with nohup issue at fetch logs, main part of the test passed.

adilGhaffarDev:
/override metal3-centos-e2e-integration-test-main metal3-ubuntu-e2e-integration-test-main

PR #1548, opened 2025-07-02
Add support for full IPv6 networking

nuhakala:
This PR adds proper support for full IPv6 networking. Works only with kind cluster.

- Fix and complement some environment variables
- Fix kind-cluster IPv6 networking
- Fix kind-cluster registry configuration

### Testing

Here are short steps to test this yourself:

1. Build iPXE image builder with IPv6 support and with correct name (or rename after building). By default the builder is named after the address it is hosted at.
2. Build iPXE with IPv6 support.
3. The Docker engine depends on containerd 1.7. To configure image registries using an IPv6 address, containerd version **2.0 or later** is required. You can download the correct containerd binary and replace the existing one under `/usr/local/bin`.
4. You need to replace the default IPv4 addresses with IPv6 addresses in the environment variables.

After these steps running `make` and all traffic should be over IPv6.

### DIFF
--- a/03_launch_mgmt_cluster.sh
+++ b/03_launch_mgmt_cluster.sh
@@ -624,17 +624,47 @@ create_clouds_yaml()
 #
 launch_kind()
 {
-    # If registry is IPv6 address, '[', ']' are not allowed
-    local registry
-    registry=$(echo "${REGISTRY}" | sed -E 's/\[\|\]//g')
-    cat <<EOF | sudo su -l -c "kind create cluster --name kind --image=${KIND_NODE_IMAGE} --config=- " "${USER}"
+    if [[ "${IP_STACK}" == "v6" ]]; then
+        # If registry is IPv6 address, '[', ']' are not allowed in kind configuration.
+        # We need to create config files for the containerd patches.
+        # Notice that having '[' and ']' in the key requires containerd 2.0 or newer.
+        cat <<EOF | tee /tmp/hosts.toml
+[host."http://${REGISTRY}/v2"]
+  capabilities = ["pull", "resolve", "push"]
+  skip_verify = true
+EOF
+
+        cat <<EOF | sudo su -l -c "kind create cluster --name kind --image=${KIND_NODE_IMAGE} --config=- " "${USER}"
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  extraMounts:
+  - hostPath: /tmp/hosts.toml
+    containerPath: /hosts.toml
+containerdConfigPatches:
+- |-
+  [plugins."io.containerd.grpc.v1.cri".registry]
+    config_path = "/etc/containerd/certs.d"
+networking:
+  ipFamily: ipv6
+EOF
+
+        # Copy config to correct place. Mounting it directly does not work
+        # because the filepath contains colons.
+        docker exec -it kind-control-plane mkdir -p /etc/containerd/certs.d/"${REGISTRY}"
+        docker exec -it kind-control-plane cp /hosts.toml /etc/containerd/certs.d/"${REGISTRY}"/hosts.toml
+
+    else
+        cat <<EOF | sudo su -l -c "kind create cluster --name kind --image=${KIND_NODE_IMAGE} --config=- " "${USER}"
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 containerdConfigPatches:
 - |-
-  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."${registry}"]
+  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."${REGISTRY}"]
     endpoint = ["http://${REGISTRY}"]
 EOF
+    fi
 }
 
 #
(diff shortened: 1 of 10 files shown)


lib/network.sh
```diff
@@ -40,7 +40,6 @@ export POD_CIDR=${POD_CIDR:-"192.168.0.0/18"}
 
 # Enables single-stack IPv6
 BARE_METAL_PROVISIONER_SUBNET_IPV6_ONLY=${BARE_METAL_PROVISIONER_SUBNET_IPV6_ONLY:-false}
-IPV6_ADDR_PREFIX=${IPV6_ADDR_PREFIX:-"fd2e:6f44:5dd8:b856"}
```
nuhakala, line 43: I grepped through the dev env and BMO and found no references to this. I don't really know what this is and it did not seem any effect so I removed it. Seems like this is some old variable that is not used anymore, so please let me know if this is needed.

Rozzii:
> This PR adds proper support for full IPv6 networking. Works only with kind cluster.
> 
>     * Fix and complement some environment variables
> 
>     * Fix kind-cluster IPv6 networking
> 
>     * Fix kind-cluster registry configuration
> 
> 
> ### Testing
> 
> Here are short steps to test this yourself:
> 
>     1. Build iPXE image builder with IPv6 support and with correct name (or rename after building). By default the builder is named after the address it is hosted at.
> 
>     2. Build iPXE with IPv6 support.
> 
>     3. The Docker engine depends on containerd 1.7. To configure image registries using an IPv6 address, containerd version **2.0 or later** is required. You can download the correct containerd binary and replace the existing one under `/usr/local/bin`.
> 
>     4. You need to replace the default IPv4 addresses with IPv6 addresses in the environment variables.
> 
> 
> After these steps running `make` and all traffic should be over IPv6.

Please add this guide to a new IPv6 section in vars.md , an please list there all the relevant environment variables. I think a minority of folks use IPv6 only setups in dev-env if anyone at all so I expect this information to get lost with time.

kashifest:
@fmuyassarov @adilGhaffarDev PTAL

Rozzii (commented):
/lgtm

Rozzii:
> Should we also add a new test for this `metal3-dev-env-integration-ipv6-test-ubuntu-main`? Or maybe just update `metal3-dev-env-integration-test-ubuntu-main` to test both ipv6 and ipv4.

We could develop a test for just this as I think dualstack tests are a bit unreliable, AFAIK we already run in "dualstack" mode but the problem with dualstack is that it is very easy to fall back to ipv4 during workflows that are harder to track e.g. ipxe. Things like Ironic API callback URI also contains a fixed IP that can't be ipv4 and ipv6 at the same time until we start using domain names in the provisioning network instead of ip adresses in the urls. 

Even if we make a proper and properly utilized dualstack deployment we would still need to test ipv6 only use cases so I am on the opinion that we should make an ipv6 test in the short to medium term and then we need more time to develop a framework that can truly validate a dualstack setup.

nuhakala:
ping @tuminoid for approval

03_launch_mgmt_cluster.sh
```diff
@@ -624,17 +624,47 @@ create_clouds_yaml()
...
+  [plugins."io.containerd.grpc.v1.cri".registry]
+    config_path = "/etc/containerd/certs.d"
+networking:
+  ipFamily: ipv6
+EOF
+
+        # Copy config to correct place. Mounting it directly does not work
+        # because the filepath contains colons.
+        docker exec -it kind-control-plane mkdir -p /etc/containerd/certs.d/"${REGISTRY}"
+        docker exec -it kind-control-plane cp /hosts.toml /etc/containerd/certs.d/"$REGISTRY"/hosts.toml
```
tuminoid:
```suggestion
        docker exec -it kind-control-plane cp /hosts.toml /etc/containerd/certs.d/"${REGISTRY}"/hosts.toml
```

hack/replace-containerd2.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/bash
```
tuminoid:
```suggestion
#!/usr/bin/env bash
```

hack/replace-containerd2.sh
```diff
@@ -0,0 +1,39 @@
...
+# will remain.
+
+set -eux
+
+sudo systemctl stop docker
+sudo systemctl stop containerd
+
+SAVE_DIR="/tmp"
+containerd="containerd.tar.gz"
+wget -O "${SAVE_DIR}"/"${containerd}" https://github.com/containerd/containerd/releases/download/v2.1.3/containerd-2.1.3-linux-amd64.tar.gz
```
tuminoid: Can we have the v2.1.3 in variable so it is easier to update, we might add it to renovate config for example.

vars.md
```diff
@@ -193,3 +193,36 @@ For testing purposes, verification of the digests will be skipped if
...
+
+The environment supports IPv6-only networking, but this currently works
+**only with `kind`**. Minikube does not have official IPv6 support at the time
+of writing. IPv6 can also be enabled partially, but for full IPv6
+networking, the following additional steps are required:
+
+1. Build iPXE image builder with IPv6 support and with correct name (or rename
+   after building). By default the builder is named after the address it is
+   hosted at.
+2. Build iPXE with IPv6 support.
```
tuminoid:
```suggestion
1. Build iPXE with IPv6 support.
```
Markdown renders numbered lists correctly from list of 1.s

vars.md
```diff
@@ -193,3 +193,36 @@ For testing purposes, verification of the digests will be skipped if
...
+3. The Docker engine depends on containerd 1.7. To configure image registries
```
tuminoid:
```suggestion
1. The Docker engine depends on containerd 1.7. To configure image registries
```

vars.md
```diff
@@ -193,3 +193,36 @@ For testing purposes, verification of the digests will be skipped if
...
+   using an IPv6 address, containerd version **2.0 or later** is required. You
+   can download the correct containerd binary and replace the existing one under
+   `/usr/local/bin`. There is a convenience script for this
+   `hack/replace-containerd2.sh`.
+4. You need to replace the default IPv4 addresses with IPv6 addresses in the
```
tuminoid:
```suggestion
1. You need to replace the default IPv4 addresses with IPv6 addresses in the
```

Rozzii (commented):
/lgtm

PR #1552, opened 2025-07-28
improve node image download logic

Rozzii:
This commit:
 - Allows the user to provide a qcow2 image locally and convert it to raw image without triggering the download node image  logic thus wasting time.

This commit was needed because in the previous implementation the dev-env pulled the qcow2 node image even if only the raw image was missing. The previous logic had multiple faults:
- It didn't pull if the qcow2 was missing even though users might not use raw image streaming
- Pulled the qcow image if the raw image was missing instead of just converting the qcow2 to raw


lentzi90 (commented):
/lgtm

PR #1557, opened 2025-08-04
Add support for openSUSE Leap 15.6

nuhakala:
This PR adds support for openSUSE Leap 15.6. It supports both kind and minikube.

- Install required packages in 01 script and add suse specific installation/configuration to ansible.
- Modify the network configurations to support also suse.
- Force ansible python interpreter, for some reason it does not use the correct python interpreter from the venv without explicitly specifying it.
- Add templates for provisioning suse images.
- Add cleanup for suse


nuhakala:
Minikube runs ironic inside a kubernetes pod, so it requires the ironic endpoint to be present only inside the VM. Ironic runs on the host with kind, so it requires the ironicendpoint interface to be present on the host. Furthermore, keepalived requires that the ironicendpoint interface has at least one IP-address assigned.

Hence we have effectively two different network setups and I separated those into different functions to make it more readable. I also believe that with this configuration we could make kind work also on centos, because the interface creation stuff should not be distro specific.

Another thing is that the kind network configuration is doing more or less the same thing as first half of `ubuntu_bridge_network_configuration.sh`. Hence it would not require much to also get rid of that script.

One thing is that how do we want to write the configuration? Do we want to explicitly write the configuration for different distros with if clauses or do we want to make the configuration depend on the used kubernetes provider?

tldr: should I refactor the networking setup even further or should I revert it and make some specific case only for suse?

Rozzii:
> /hold @nuhakala hi! thanks for the effort but I believe adding support for yet another distribution needs to be discussed first, also considering that we don't have CI jobs running on Suse, and we don't have people in the team (AFAIK) that can maintain that

Just to reflect on the manpower issue, we are willing to maintain it and integrate it to the CI, work is already ongoing.
We have a need to test with the the community variants of all the 3 large enterprise distro Ubuntu, RHEL, SLES, and we have been thinking that we would share it with the community.

I 100% agree that we have to discuss this with the community but AFAIK there won't be any negative side effect for the already supported distros and we are not aiming to design any SLES specific use cases.

kashifest:
> /hold @nuhakala hi! thanks for the effort but I believe adding support for yet another distribution needs to be discussed first, also considering that we don't have CI jobs running on Suse, and we don't have people in the team (AFAIK) that can maintain that

I think manpower and maintaining CI jobs wont be an issue as we are willing to maintain it along with other CI jobs as well.

elfosardo:
my comment comes mainly from observation of the latest year and a half advancement in CI
the work to upgrade CI to current stable versions of ubuntu (ubuntu noble) and CentOS (CentOS Stream 10) is stagnating since a while, despite having full support in metal3-dev-env, so I don't see how we will be able to maintain also a third distribution
but again, we need to discuss this first, better if in a meeting

kashifest:
> my comment comes mainly from observation of the latest year and a half advancement in CI the work to upgrade CI to current stable versions of ubuntu (ubuntu noble) and CentOS (CentOS Stream 10) is stagnating since a while, despite having full support in metal3-dev-env, so I don't see how we will be able to maintain also a third distribution but again, we need to discuss this first, better if in a meeting

I think this is pretty orthogonal to the Ubuntu or CentOS work, but sure lets discuss in the upcoming community meeting so that we are on the same page.

kashifest:
@elfosardo since we agreed over the community meeting few weeks back, unholding this to take it forward.
/hold cancel

Meanwhile here is the Centos stream 10 tracking issue https://github.com/metal3-io/project-infra/issues/1080 which we will take up in parallel.

tests/inventory.ini
```diff
@@ -1,2 +1,2 @@
 [virthost]
-localhost ansible_python_interpreter="auto"
+localhost ansible_python_interpreter="{{ lookup('env', 'ANSIBLE_VENV') | default('/opt/metal3-dev-env/venv/', true) }}/bin/python3"
```
nuhakala: I added this to pass the tests. It seems that this PR somehow affects ansible so that it cannot find the python interpreter from the venv in which ansible is installed. I have zero clue why this happens, as this PR seemingly does not touch anything ubuntu related, and even less anything ansible related in 01 and 02 scripts. I had to force the python interpreter for suse also, but that code should not have any effect on what happens 01/02 scripts, as the setup-playbook in 02 is the first one showing this problem.

If anyone has any idea why this happens, I would be glad to learn and fix it.

host_cleanup.sh
```diff
@@ -55,27 +55,27 @@ ANSIBLE_FORCE_COLOR=true "${ANSIBLE}-playbook" \
...
+    if [[ "${MANAGE_EXT_BRIDGE}" = "y" ]]; then
+        sudo ip link delete external || true
+    fi
+else
     sudo rm -rf /etc/NetworkManager/conf.d/dnsmasq.conf
     if [[  "${MANAGE_PRO_BRIDGE}" == "y" ]]; then
-        sudo nmcli con delete ironic-peer
-        sudo nmcli con delete "${BARE_METAL_PROVISIONER_INTERFACE}"
+        sudo ip link del ironic-peer
+        sudo ip link del ironicendpoint
```
tuminoid: Why are we removing use of the variable`BARE_METAL_PROVISIONER_INTERFACE`?

lib/common.sh
```diff
@@ -61,7 +66,7 @@ source /etc/os-release
 export DISTRO="${ID}${VERSION_ID%.*}"
 export OS="${ID}"
 export OS_VERSION_ID="${VERSION_ID}"
-export SUPPORTED_DISTROS=(centos9 rhel9 centos10 rhel10 ubuntu20 ubuntu22 ubuntu24)
+export SUPPORTED_DISTROS=(centos9 rhel9 centos10 rhel10 ubuntu20 ubuntu22 ubuntu24 opensuse-leap15)
 
 if [[ ! "${SUPPORTED_DISTROS[*]}" =~ ${DISTRO} ]]; then
   echo "Supported OS distros for the host are: CentOS Stream 9 or RHEL9 or Ubuntu20.04 or Ubuntu 22.04"
```
tuminoid: Fix the error message as well.

lib/common.sh
```diff
@@ -468,10 +473,19 @@ case "${FSTYPE}" in
...
   ;;
   'xfs')
-    # shellcheck disable=SC2143
-    if [[ $(xfs_info "${FILESYSTEM}" | grep -q "ftype=1") ]]; then
-      echo "XFS filesystem must have ftype set to 1"
-      exit 1
+    # Suse has xfs_info under /usr/sbin and it needs to run with sudo
+    if [[ "${OS}" = "opensuse-leap"* ]]; then
+      # shellcheck disable=SC2143
+      if [[ $(sudo xfs_info "${FILESYSTEM}" | grep -q "ftype=1") ]]; then
```
tuminoid:
```suggestion
      if [[ $(sudo xfs_info "${FILESYSTEM}" | grep "ftype=1" &>/dev/null) ]]; then
```
`grep -q` has some weird quirks that sometimes lead to failures in seemingle basic grep operations, better not use `-q` and just redirect output to dev/null.

lib/common.sh
```diff
@@ -468,10 +473,19 @@ case "${FSTYPE}" in
...
+          echo "XFS filesystem must have ftype set to 1"
+          exit 1
+      fi
+    else
+      # shellcheck disable=SC2143
+      if [[ $(xfs_info "${FILESYSTEM}" | grep -q "ftype=1") ]]; then
```
tuminoid: Same

lib/images.sh
```diff
@@ -18,6 +18,9 @@ elif [[ "${IMAGE_OS}" == "centos" ]]; then
 elif [[ "${IMAGE_OS}" == "flatcar" ]]; then
   export IMAGE_NAME=${IMAGE_NAME:-flatcar_production_qemu_image.img.bz2}
   export IMAGE_LOCATION=${IMAGE_LOCATION:-https://stable.release.flatcar-linux.net/amd64-usr/current/}
+elif [[ "${IMAGE_OS}" == "leap" ]]; then
+  export IMAGE_NAME=${IMAGE_NAME:-LEAP_15_6_NODE_IMAGE_K8S_v1.33.0.qcow2}
```
tuminoid: We need to follow-up with k8s 1.34 bump for leap?

tests/roles/run_tests/templates/main/cluster-template-controlplane-kubeadm-config-leap.yaml
tuminoid, line 1: YAML needs to be k8s yaml formatted, ie. no extra indentation for lists.

tests/roles/run_tests/templates/main/cluster-template-controlplane-kubeadm-config-leap.yaml
```diff
@@ -0,0 +1,147 @@
+# CentOS specific controlplane kubeadm config
+preKubeadmCommands:
+  - systemctl stop wicked
```
tuminoid:
```suggestion
- systemctl stop wicked
```
This is correct indentation level for yamls.

tests/roles/run_tests/templates/main/cluster-template-workers-kubeadm-config-leap.yaml
tuminoid, line 1: Same

tests/run.sh
```diff
@@ -32,8 +32,15 @@ export ANSIBLE_HOST_KEY_CHECKING=False
...
+    ansible_python_interpreter="-e ansible_python_interpreter=${ANSIBLE_VENV}/bin/python"
+fi
+
+# shellcheck disable=SC2086
 ANSIBLE_FORCE_COLOR=true "${ANSIBLE}-playbook" \
-   -e "metal3_dir=$SCRIPTDIR" \
-   -e "v1aX_integration_test_action=${ACTION}" \
-   -i "${METAL3_DIR}/tests/inventory.ini" \
-   -b -v "${METAL3_DIR}/tests/main.yml"
+    -e "metal3_dir=$SCRIPTDIR" \
```
tuminoid:
```suggestion
    -e "metal3_dir=${SCRIPTDIR}" \
```
Let's fix this as you touch the line anyways.

vm-setup/roles/packages_installation/defaults/main.yml
```diff
@@ -99,8 +99,34 @@ packages:
       - jmespath
       - kubernetes==25.3.0
       - passlib
+  suse:
+    common:
+      packages:
+        - jq
```
tuminoid, line 105: Ok, there is plenty of pre-existing wrong indents here, so this is fine. Let's fix all the ansible ymls in another cleanup PR.

tests/inventory.ini
tuminoid: Hardcoding /opt/metal3-dev-env is wrong, the fallback should be system `/usr`. `$WORKING_DIR` (ie. /opt/metal3-dev-env) is applied via ANSIBLE_VENV already.

Btw, do you test with "CI images" with Leap as well (ie. pre-built images where 01 has previously been run) or do you start manually from empty vanilla, then run 01 and 02? This little difference of basically rebooting between 01 and 02 sometimes make huge difference how things work. 

Also verify from project-infra that in CI runs we do run 01 as well. At some point we skipped 01 for saving time in CI (since it was run in image creation), this could lead to some funky issues when touching 01 scripts.

host_cleanup.sh
nuhakala: Tbh, I am not sure what we should do with this. The variable is used somewhere, but at in most places not. For example `ubuntu_bridge_network_configuration.sh` does not use it at all. The interface `ironicendpoint` is hardcoded in most templates and in some places a variable `INT_IF` is used instead (like in 02 script). So at this point it felt most clear to just hardcode it, because it was just confusing. I can revert it though if you want.

lib/images.sh
nuhakala: Umm, no I guess? The bump was not done when I created the PR and now when I rebased these PRs I used K8s v1.34.0 for testing and it passed. I will update this too.

tests/roles/run_tests/templates/main/cluster-template-controlplane-kubeadm-config-leap.yaml
nuhakala: Are they now correct?

tests/run.sh
nuhakala: Actually, after touching the python interpreter in the inventory I realized that adding this logic here is unnecessary as the inventory change will do the same thing. Hence I reverted the change back to original, but now that I was at it I fixed this variable anway.

nuhakala:
I have now tested these changes with CI images and manual tests passed. I also tried to test the CI image on jenkins but it failed because node images were not available online to download to test environment. The test results are [here](https://github.com/metal3-io/project-infra/pull/1063#issuecomment-3270348700)

host_cleanup.sh
tuminoid: I think its better the use variable as it defaults to `ironicendpoint` as there is no harm in that.

lib/common.sh
```diff
@@ -468,10 +473,19 @@ case "${FSTYPE}" in
...
+    # Suse has xfs_info under /usr/sbin and it needs to run with sudo
+    if [[ "${OS}" = "opensuse-leap"* ]]; then
+      # shellcheck disable=SC2143
+      if [[ $(sudo xfs_info "${FILESYSTEM}" | grep "ftype=1" &>/dev/null) ]]; then
+          echo "XFS filesystem must have ftype set to 1"
+          exit 1
+      fi
+    else
+      # shellcheck disable=SC2143
+      if [[ $(xfs_info "${FILESYSTEM}" | grep "ftype=1" &>/dev/null) ]]; then
```
tuminoid:
```suggestion
      if ! xfs_info "${FILESYSTEM}" | grep "ftype=1" &>/dev/null; then
```
Were not testing anything (which [[ does), so its just bare command check.

lib/common.sh
```diff
@@ -468,10 +473,19 @@ case "${FSTYPE}" in
   'ext4'|'btrfs')
   ;;
   'xfs')
-    # shellcheck disable=SC2143
-    if [[ $(xfs_info "${FILESYSTEM}" | grep -q "ftype=1") ]]; then
-      echo "XFS filesystem must have ftype set to 1"
-      exit 1
+    # Suse has xfs_info under /usr/sbin and it needs to run with sudo
```
tuminoid: Also, cannot we run xfS_info with sudo in all distros? It would simplify the code here. We're configuring passwordless sudo anyways.

lib/common.sh
```diff
@@ -468,10 +473,19 @@ case "${FSTYPE}" in
...
+    if [[ "${OS}" = "opensuse-leap"* ]]; then
+      # shellcheck disable=SC2143
+      if [[ $(sudo xfs_info "${FILESYSTEM}" | grep "ftype=1" &>/dev/null) ]]; then
```
tuminoid: Same as before, but prefer duplication removed.

lib/common.sh
nuhakala: Yes it can, little silly of me not to consider that

tests/roles/run_tests/tasks/main.yml
```diff
@@ -1,4 +1,8 @@
 ---
+- name: testi
```
tuminoid: Some test code here.
/hold

lib/common.sh
```diff
@@ -465,11 +470,9 @@ fi
 FSTYPE="$(df "${FILESYSTEM}" --output=fstype | tail -n 1)"
 
 case "${FSTYPE}" in
-  'ext4'|'btrfs')
-  ;;
+  'ext4'|'btrfs') ;;
   'xfs')
-    # shellcheck disable=SC2143
-    if [[ $(xfs_info "${FILESYSTEM}" | grep -q "ftype=1") ]]; then
+    if sudo xfs_info "${FILESYSTEM}" | grep "ftype=1" &>/dev/null; then
```
tuminoid, line 475: Now this exits if ftype=1, you missed !

```suggestion
    if ! sudo xfs_info "${FILESYSTEM}" | grep "ftype=1" &>/dev/null; then
```

lib/common.sh
```diff
@@ -465,11 +470,9 @@ fi
 FSTYPE="$(df "${FILESYSTEM}" --output=fstype | tail -n 1)"
 
 case "${FSTYPE}" in
-  'ext4'|'btrfs')
-  ;;
+  'ext4'|'btrfs') ;;
```
tuminoid:
```suggestion
  ext4|btrfs) ;;
```
Not needed.

lib/common.sh
```diff
@@ -465,11 +470,9 @@ fi
...
   'xfs')
```
tuminoid:
```suggestion
  xfs)
```
Same

tests/roles/run_tests/tasks/main.yml
nuhakala: Removed

nuhakala: /unhold

PR #1562, opened 2025-08-20
Fix docker exec to run in CI

nuhakala:
This pr removes `-it` flag from docker exec commands, because they will crash if ran in CI.

Furthermore, this also removes a comment about containerd patch in 01 script. After testing more, it apparently is not mandatory to patch containerd on the host. Kind docker container contains containerd 2.x and the configuration requiring containerd 2.x goes inside kind container.

Finally also IPv6 docs is updated in `vars.md`


tuminoid (commented):
/approve

Issue #1556, opened 2025-08-04
Remove hard coded CAPICAPIRELEASE when uplifting to CAPI v1.11.0-beta/rc/stable


tuminoid:
Reminder task to remove hard coded CAPICAPIRELEASE when uplifting to CAPI v1.11.0-beta/rc/stable.

Ref: https://github.com/metal3-io/metal3-dev-env/pull/1551

/assign @adilGhaffarDev 
/triage accepted

PR #1566, opened 2025-09-01
🌱 Remove hard coded version of CAPIRELEASE

adilGhaffarDev:
Capi minor is released, we can remove the hard-coded version of CAPIRELEASE.

Fixes: https://github.com/metal3-io/metal3-dev-env/issues/1556


tuminoid (commented):
/approve

PR #1567, opened 2025-09-04
Add NAT64 to enable IPv6 provision in v4 only host

nuhakala:
This adds
- Tayga as NAT64 service
- Bind9 and configure it as local DNS server with DNS64
- IPv6 specific addresses and logic to cluster, controlplane, and worker templates
- Updates vars.md with new variables
- Updates CNI configuration for IPv6

Why we need this? Previously IPv6 support was added so that the dev env could create bare metal hosts over IPv6. This PR extends that so that we can also provision the bare metal hosts over IPv6 and the provisioned images will be IPv6 only. The tricky part is that our CI environment does not support IPv6 natively, so the VMs cannot access internet over IPv6 and hence also cannot download Kubernetes images and other needed images to set up cluster.

Introducing NAT64/DNS64 solves that issue. It essentially allows the dev env to be deployed with IPv6 only scenario on IPv4 host. Furthermore, this PR introduces required changes to the templates so that they are configured to use IPv6.

**NOTE!** Running `make test` does not pass yet. However, these changes should allow IPv6 only BMH to be provisioned with operating system and creating a K8s cluster in those BMHs. Only working with centos node images.

Other not directly related changes that could actually be in their own PR:
- Moving the docker daemon template in `vm-setup/roles/packages_installation/files/daemon.json` to `vm-setup/roles/packages_installation/templates/daemon.json` to better reflect the purpose of the file.
- Fixing kubernetes yum repository address in template.

Other considerations
- Currently tests failed after pivoting the control plane components to target cluster. In this step the target cluster needs to create pods for capm3 and other metal3 related components, but this failed because crio inside the target VM was not able to resolve IPv6 address in the registry address. So pulling images failed with `cannot parse input: [fd55::1]:5000/localimages/cluster-api-provider-metal3:main`. I manually tested this and crio was able to pull images after creating a hostname in `/etc/hosts` and specifying that hostname instead of bare IPv6 address.
- I had to change `CLUSTER_APIENDPOINT_HOST` into `CLUSTER_APIENDPOINT_IP` in some templates, because having brackets around the IPv6 address caused errors.
- Introducing domain names would solve both these problems, but do we want to spend time on that. Though as this PR introduces proper DNS server, and configures the environment to use that, using local domain names could be feasible.


nuhakala:
/cc @Rozzii @tuminoid 

This PR is quite big so I can also split it into smaller PRs.

The IPv6 is not ready in dev env after this, but I just want to have some feedback and discussion on how to do the DNS stuff.

The bot already pinged @elfosardo but you might also want to test this on your tests, because this is quite big and introduces new moving parts.

tests/roles/run_tests/templates/main/metal3datatemplate-template.yaml
```diff
@@ -66,5 +66,5 @@ spec:
         - 8.8.8.8
 {% endif %}
 {% if IP_STACK == 'v6' or IP_STACK == 'v4v6' %}
-        - 2001:4860:4860::8888
+        - {{ LOCAL_DNS_V6 }}
```
nuhakala, line 73: This is little controversial. I don't think that it really matters inside the dev env, as it does not really use IPv6 itself, but this might break some users use case.

tests/roles/run_tests/templates/main/cluster-template-workers-kubeadm-config-centos.yaml
```diff
@@ -1,5 +1,7 @@
 # CentOS specific worker kubeadm config
 preKubeadmCommands:
+  - sysctl -w net.ipv6.conf.all.forwarding=1
```
terror96, line 4: Would there be need to guard all changes (in this file) to inside `{% if IP_STACK == "v6" %}`?

tests/roles/run_tests/templates/main/cluster-template-controlplane-kubeadm-config-centos.yaml
```diff
@@ -1,5 +1,7 @@
 # CentOS specific controlplane kubeadm config
 preKubeadmCommands:
+  - sysctl -w net.ipv6.conf.all.forwarding=1
```
terror96, line 4: Would there be need to guard all changes (in this file) to inside `{% if IP_STACK == "v6" %}`?

tests/roles/run_tests/templates/main/cluster-template-controlplane-kubeadm-config-centos.yaml
```diff
@@ -66,7 +68,11 @@ files:
           priority 101
           advert_int 1
           virtual_ipaddress {
-              {{ CLUSTER_APIENDPOINT_HOST }}
+{% if IP_STACK == "v6" %}
+{# According to VRRP, link local address should come first with v6 #}
+              {{ CLUSTER_APIENDPOINT_LINK_LOCAL }}
+{% endif %}
+              {{ CLUSTER_APIENDPOINT_IP }}
```
terror96, line 77: Same as earlier: is there an ...

tests/roles/run_tests/templates/main/cluster-template-cluster.yaml
```diff
@@ -25,7 +25,7 @@ metadata:
   namespace: ${ NAMESPACE }
 spec:
   controlPlaneEndpoint:
-    host: ${ CLUSTER_APIENDPOINT_HOST }
+    host: ${ CLUSTER_APIENDPOINT_IP }
```
terror96, line 28: Is there an original setup where we would prefer using CLUSTER_APIENDPOINT_HOST?

tests/roles/run_tests/vars/main.yml
```diff
@@ -61,6 +62,7 @@ CALICO_MINOR_RELEASE: "{{ lookup('env', 'CALICO_MINOR_RELEASE') | default('v3.25
 CALICO_PATCH_RELEASE: "{{ lookup('env', 'CALICO_PATCH_RELEASE') | default('v3.25.1', true) }}"
 DOCKER_HUB_PROXY: "{{ lookup('env', 'DOCKER_HUB_PROXY') }}"
 WORKING_DIR: "{{ lookup('env', 'WORKING_DIR') | default('/opt/metal3-dev-env', true) }}"
+LOCAL_DNS_V6: "{{ lookup('env', 'LOCAL_DNS_V6') | default('fd00:abcd::1', true) }}"
```
terror96, line 68: How about keeping the original default value: '2001:4860:4860::8888' or is that some obsolete value?

vars.md
```diff
@@ -217,6 +222,10 @@ networking, the following additional steps are required:
 1. You need to replace the default IPv4 addresses with IPv6 addresses in the
    environment variables.
 
+Notice that if your host does not have support for native IPv6, you need to
```
terror96, line 227: "It would most likely be beneficial to support cases where NAT64 and DNS64 are deployed separately, because there could be scenarios where one uses a shared DNS64."

tests/roles/run_tests/vars/main.yml
```diff
@@ -61,6 +62,7 @@ CALICO_MINOR_RELEASE: "{{ lookup('env', 'CALICO_MINOR_RELEASE') | default('v3.25
...
```
terror96: Traditionally fd ULA-addresses have a 40-bit of Global ID: https://en.wikipedia.org/wiki/Unique_local_address ... yes, it can be zeros, but would still pick something else .... not that it really matters in this case.

vm-setup/roles/nat64/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
+nat64_action: setup
+ENABLE_NAT64: "{{ lookup('env', 'ENABLE_NAT64') | default('false', true) }}"
+TEMPLATES_PATH: "{{ metal3_dir }}/vm-setup/roles/nat64/files"
+DNS_UPSTREAM: "{{ lookup('env', 'DNS_UPSTREAM') | default('10.1.0.2', true) }}"
+DNS64_PREFIX: "{{ lookup('env', 'DNS64_PREFIX') | default('fd00:ffff:ffff::/96', true) }}"
+# Local DNS is only used in IPv6 only env
+LOCAL_DNS_V4: "{{ lookup('env', 'LOCAL_DNS_V4') | default('127.0.0.2', true) }}"
+LOCAL_DNS_V6: "{{ lookup('env', 'LOCAL_DNS_V6') | default('fd00:abcd::1', true) }}"
```
terror96: Same ULA nit.

vm-setup/roles/nat64/tasks/teardown_nat.yml
```diff
@@ -0,0 +1,23 @@
...
+  shell:
+    cmd: "ip addr del {{ LOCAL_DNS_V4 }}/32 dev lo"
+  become: yes
+  ignore_errors: true
+
+- name: Delete DNS-server IPv6-address from loopback interface
+  shell:
+    cmd: "ip addr del {{ LOCAL_DNS_V6 }}/128 dev lo"
+  become: yes
+  ignore_errors: true
```
terror96, line 23: We could also add IPv4 and IPv6 addresses to the tunnel device in order to enable sending of ICMP(v6) messages. It is a stupid that the need to be added manually, because they also need to be specified in the tayga configuration file. Well, life is.

vm-setup/roles/nat64/templates/tayga.conf
```diff
@@ -0,0 +1,5 @@
+tun-device nat64
+ipv4-addr 192.168.255.1
+prefix {{ DNS64_PREFIX }}
+dynamic-pool 192.168.255.0/24
+data-dir /var/spool/tayga
```
terror96, line 5: Maybe also IPv6 address for ICMPv6 traffic.

vm-setup/roles/packages_installation/templates/daemon.json
```diff
@@ -1,4 +1,7 @@
 {
+{% if ENABLE_NAT64|bool %}
+    "dns": [ "{{ DOCKER_LOCAL_DNS_V6 }}" ],
+{% endif %}
     "ipv6": {{ DOCKER_IPV6_SUPPORT }},
     "fixed-cidr-v6": "fd00:d0c4::/32",
```
terror96: Not your doing, but I would stick to a minimum of a 48-bit prefix. See: https://en.wikipedia.org/wiki/Unique_local_address

terror96 (commented):
Minor comments & suggestions

tests/roles/run_tests/templates/main/cluster-template-workers-kubeadm-config-centos.yaml
nuhakala: Added.

tests/roles/run_tests/templates/main/cluster-template-cluster.yaml
nuhakala: This is rather annoying. Using host would be usually the way to go, but in this case if we add the brackets around IPv6 address to make it look like host, kubeadm errors and the cluster does not get connection.

Also, host names are not used anywhere, so it should not be a problem to use ATM, but of course moving to use host names later would be annoying.

tests/roles/run_tests/vars/main.yml
nuhakala: Should not be, changed that to be the default.

vars.md
nuhakala: I agree. But ATM nobody seems to use IPv6 anyway, so not sure if it makes sense to make this more complicated than needed.

tests/roles/run_tests/vars/main.yml
nuhakala: Changed this to the original address

vm-setup/roles/nat64/tasks/teardown_nat.yml
nuhakala: Yeah, that would be good, but again, I feel that introducing any extra complexity at this point is not really needed.

vm-setup/roles/packages_installation/templates/daemon.json
nuhakala: Actually, was my doing in a previous PR :D Anyway, fixed this also.

tuminoid:
Ping @nuhakala, will you pick this up again ?

PR #1581, opened 2025-10-16
:seedling: Fix PodCIDR replace in Calico manifest

peppi-lotta:
We are getting the calico manifest from https://raw.githubusercontent.com/projectcalico/calico/v3.30.3/manifests/calico.yaml and replacing the PodCIDR. You well see by reading the manifest that the line that is being replaced is actually commented out so the replace is not doing anything. This changes the code so that it replaces the value and un-comments the line. 
Also:
- Update calico version
- Simplify Calico version


adilGhaffarDev (approved):
/lgtm

PR #1594, opened 2025-11-17
dev: add tiltfile for local development

fmuyassarov:
This is initial version which is expected to iterate over to have more customization points such as, disabling auto update or continuing on top of metal3-dev-env prepared environment. Right now, it doesn't have and dependency nor connection to the way that metal3-dev-env prepares the dev environment but I plan to make it easier to integrate it in case developers want to continue modifying and testing their metal3 controllers on the environment prepared by metal3-dev-env (optionally).

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -1,0 +1,298 @@
+# -*- mode: Python -*-
+# External Tilt extensions we depend on
+load('ext://git_resource', 'git_resource', 'git_checkout')
+load('ext://color', 'color')
+load('ext://cert_manager', 'deploy_cert_manager')
+
+# Custom flag for tilt down
+config.define_bool("prune")
+cfg = config.parse()
+
+allow_k8s_contexts('kubernetes-admin@kubernetes')
+# Anonymous, ephemeral image registry
+default_registry('ttl.sh')
+
+# ==============================================================================
+# Vars
+# ==============================================================================
+# versions
+CAPI_VERSION = os.getenv('CAPI_VERSION', 'v1.11.3')
+CERT_MANAGER_VERSION = os.getenv('CERT_MANAGER_VERSION', 'v1.19.1')
+
+# capi core providers
+CAPI_BOOTSTRAP_PROVIDER = os.getenv('CAPI_BOOTSTRAP_PROVIDER', 'kubeadm')
+CAPI_CONTROLPLANE_PROVIDER = os.getenv('CAPI_CONTROLPLANE_PROVIDER', 'kubeadm')
+CAPI_CORE_PROVIDER = os.getenv('CAPI_CORE_PROVIDER', 'cluster-api')
+
+capi_components_url = 'https://github.com/kubernetes-sigs/cluster-api/releases/download/{}/cluster-api-components.yaml'.format(CAPI_VERSION)
+local_dev_dir = os.getenv('LOCAL_DEV_DIR', './local-dev')
+
+bmo = 'baremetal-operator'
+capm3 = 'cluster-api-provider-metal3'
+ipam = 'ip-address-manager'
+irso = 'ironic-standalone-operator'
+bmo_kustomize_dir = '{}/{}/config/'.format(local_dev_dir, bmo)
+capm3_kustomize_dir = '{}/{}/config/default'.format(local_dev_dir, capm3)
+ipam_kustomize_dir = '{}/{}/config/default'.format(local_dev_dir, ipam)
+irso_kustomize_dir = '{}/{}/config/default'.format(local_dev_dir, irso)
+
+# ==============================================================================
+# Helper Functions
+# ==============================================================================
+def check_tools(tools):
+    for name in tools:
+        if str(local("%s version 2>/dev/null || true" % name, quiet=True, echo_off=True)) == "":
+            fail("%s not found in PATH" % name)
+        print_success("%s OK" % name)
+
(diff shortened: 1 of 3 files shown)


tuminoid (commented):
Default images run as non-root and with other security hardenings, which should prevent Tilt from updating the manager binaries when they change. I don't see that patched here, is it now handled differently? Ref: https://github.com/metal3-io/cluster-api-provider-metal3/blob/11189599fd7b84280e8c1ce0f56d44154cb86d80/Tiltfile#L4

tuminoid:
> > > > > > Default images run as non-root and with other security hardenings, which should prevent Tilt from updating the manager binaries when they change. I don't see that patched here, is it now handled differently? Ref: https://github.com/metal3-io/cluster-api-provider-metal3/blob/11189599fd7b84280e8c1ce0f56d44154cb86d80/Tiltfile#L4
> > > > > 
> > > > > 
> > > > > I'm not sure I understand what you mean, can you elaborate ? Have you tested this? I didn't have any issues with live update across all repos.
> > > > 
> > > > 
> > > > No, I did not test this, just noticed its different on this part. If you check the line referenced, it calls remove_sec_ctx.py which removes the securityContext from the yaml, to allow Tilt to update the manager binary on the running containers. It could be Tilt has been updated to workaround this then? Or is it deployed some other more permissive way?
> > > 
> > > 
> > > Not sure to be honest. I'm trying to understand why would Tilt not be able to deploy k8s objects if they have some `securityContext` settings.
> > 
> > 
> > SecurityContext might have `readOnlyRoot: true` or `user: nobody` and the manager binary is root owned etc.
> 
> As discussed offline, let's leave this kind of build process optimization for later as follow up.

And just for completeness sake, this is not an issue in the current form of PR as it does full container rebuild instead of binary replacement.

This support needs to be brought back later as it is making development cycle faster by avoiding full redeployment. We also have some other custom parts (like BMH creation button) that need to back added later.

Tiltfile
```diff
@@ -0,0 +1,298 @@
...
+allow_k8s_contexts('kubernetes-admin@kubernetes')
+# Anonymous, ephemeral image registry
+default_registry('ttl.sh')
+
+# ==============================================================================
+# Vars
+# ==============================================================================
+# versions
+CAPI_VERSION = os.getenv('CAPI_VERSION', 'v1.11.3')
+CERT_MANAGER_VERSION = os.getenv('CERT_MANAGER_VERSION', 'v1.3.1')
```
lentzi90: Is this a typo? It is very old!

```suggestion
CERT_MANAGER_VERSION = os.getenv('CERT_MANAGER_VERSION', 'v1.19.1')
```

Tiltfile.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+This will remove all deployed resources AND delete the `./local-dev`
+directory with cloned repositories.
+
+## Customization
+
+Customize deployment parameters using environment variables:
+
+### Repository Sources
```
lentzi90, line 73: Why do we want this? I would expect to have all relevant repositories already locally, and perhaps dirty repos with changes I want to test. Having to push first so that I can run this seems counter intuitive to me. Could we have paths to the repos instead?

Tiltfile.md
```diff
@@ -0,0 +1,123 @@
...
+
+```bash
+# Choose ONE of the following formats:
+
+# Format 1: Repository name only (falls back to default org - Nordix)
```
lentzi90: We should use the upstream I think.

```suggestion
# Format 1: Repository name only (falls back to default org - metal3-io)
```

fmuyassarov: I used Nordix since we always submit patches from Nordix. But considering the fact that it is not only us who may use the Tilt (though I expect it will be only us) it makes sense to default to metal3-io. I was just trying to minimize required exports before running tilt up. I will convert it to metal3.-io.

Tiltfile.md
fmuyassarov: I've already added a variable `LOCAL_DEV_DIR` that stores local copies of Metal3 projects. If you specify certain directory for `LOCAL_DEV_DIR` then Tilt logic will check if the Metal3 projects are there and clone them only if they don't exits on the given path.

lentzi90: Ok that sounds good :+1:

Tiltfile.md
```diff
@@ -0,0 +1,123 @@
...
+
+```bash
+tilt up
+```
+
+This will:
+
+- Check that all required tools are installed
+- Clone repositories to `./local-dev` (only if they don't exist)
+- Deploy cert-manager (by default v1.3.1)
```
lentzi90:
```suggestion
- Deploy cert-manager (by default v1.19.1)
```

Tiltfile.md
lentzi90: Code is now using metal3-io but this document still has Nordix

lentzi90:
Ah the Nordix comment still shows as resolved. Commenting here for visibility.

The Tiltfile now uses metal3-io but the doc still has Nordix in one place.

Tiltfile.md
fmuyassarov: thanks for the catch. Fixed, it

adilGhaffarDev:
/override metal3-centos-e2e-integration-test-release-1-11 metal3-dev-env-integration-test-ubuntu-main

PR #1595, opened 2025-11-19
virtbmc: valid SubjectAltName for TLS on sushy-tools

dtantsur:
This change allows potentially enabling proper TLS validation, now that
Ironic supports it.

### DIFF
--- a/vm-setup/roles/virtbmc/tasks/setup_tasks.yml
+++ b/vm-setup/roles/virtbmc/tasks/setup_tasks.yml
@@ -103,9 +103,18 @@
   community.crypto.openssl_privatekey:
     path: "{{ working_dir }}/virtualbmc/sushy-tools/key.pem"
 
+- name: Create CSR for redfish TLS
+  community.crypto.openssl_csr_pipe:
+    privatekey_path: "{{ working_dir }}/virtualbmc/sushy-tools/key.pem"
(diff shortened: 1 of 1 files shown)


vm-setup/roles/virtbmc/tasks/setup_tasks.yml
```diff
@@ -103,9 +103,18 @@
   community.crypto.openssl_privatekey:
     path: "{{ working_dir }}/virtualbmc/sushy-tools/key.pem"
 
+- name: Create CSR for redfish TLS
+  community.crypto.openssl_csr_pipe:
+    privatekey_path: "{{ working_dir }}/virtualbmc/sushy-tools/key.pem"
+    common_name: metal3.io
```
lentzi90, line 109: Do we need the common name? I have been recommended to avoid it if possible and just use SAN.

dtantsur: I know that at least some tools freak out if they don't find a common name.

tuminoid: CommonName is deprecated and used only as fallback if SAN is missing, and currently for example browsers must reject it if it only has CN and no SANs. Having them both is fine for now.

PR #1598, opened 2025-11-28
pin ovmf for ubuntu24.04 to fix regression

Rozzii:
After the most recent ovmf ubuntu package upgrade, the UEFI jobs started failing as the vms never attempt network boot. As a workaround we pin the ovmf pkg to the precedent working version.

### DIFF
--- a/vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
+++ b/vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
@@ -1,17 +1,23 @@
 ---
 - name: Install required Ubuntu packages
   block:
-    - name: Update all packages to their latest version
+    - name: Enable Ubuntu noble-proposed repository
+      apt_repository:
+        repo: "deb http://archive.ubuntu.com/ubuntu/ noble-proposed main"
+        state: present
+        update_cache: yes
+
+    - name: Make sure required distro packages are present
       apt:
         name: "*"
-        state: latest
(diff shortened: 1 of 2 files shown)


vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
```diff
@@ -6,12 +6,20 @@
         name: "*"
         state: latest
 
+    - name: "Pin OVMF to specific version on Ubuntu Noble"
+       apt:
+         name: "ovmf=2024.02-2ubuntu0.5"
```
tuminoid:
```suggestion
         name: "ovmf=2024.02-2ubuntu0.6"
```

Rozzii: The regression comes from 0.6

tuminoid: Oh yes, we don't want the new one :)

tuminoid:
> @Rozzii thinking about it, if ovmf is already present it gets updated to latest, then we ask for a downgrade but I don't think it will work can you confirm that this is actually WAI ? I can't see the jenkins logs anywhere

Log collector explodes with current Ubuntu image. The image somehow gets corrupted once in a while and it starts happening.

That said, I think better strategy would be to install right version at image build and then hold it to prevent it to get updated.

PR #1599, opened 2025-12-02
change golang download url to official url + bump 1.24.10

tuminoid:
Storage bucket is not allowing anonymous access, and official dl urls are expected to be used. Bump to 1.24.10 while at it.


tuminoid:
/override metal3-dev-env-integration-test-ubuntu-main
unrelated dead worker, go 1.24.10 was installed fine.

PR #1669, opened 2026-04-21
fix: add libvirt-python to ansible venv pip install

### LINKED ISSUE #1667: 01_prepare_host.sh: libvirt-python not installed in ansible venv causing VM creation to fail
## Description
When running metal3-dev-env on Ubuntu 22.04 on GCP with nested 
virtualization, the vm-setup/setup-playbook.yml fails with:

"The `libvirt` module is not importable. Check the requirements."

## Root cause
01_prepare_host.sh creates the Ansible venv and installs ansible 
but does not install libvirt-python:

sudo python -m venv --system-site-packages "${ANSIBLE_VENV}"
sudo "${ANSIBLE_VENV}/bin/pip" install ansible=="${ANSIBLE_VERSION}"

The community.libvirt Ansible module requires libvirt-python to be 
installed in the same Python environment. Since the venv is used to 
run ansible-playbook, libvirt-python must be installed in the venv.

## Fix
Add libvirt-python to the pip install step in 01_prepare_host.sh:

sudo "${ANSIBLE_VENV}/bin/pip" install \
  --ignore-installed \
  ansible=="${ANSIBLE_VERSION}" \
  libvirt-python

## Environment
- Host OS: Ubuntu 22.04 LTS
- Platform: GCP n2-standard-8 with nested virtualization
- metal3-dev-env: main branch

## Workaround
Manually install libvirt-python in the venv:
/opt/metal3-dev-env/venv/bin/pip install libvirt-python

Ahmedaltu:
Fixes #1667

## What this PR does
Adds `libvirt-python` to the pip install step in `01_prepare_host.sh`
so it is installed in the Ansible venv alongside ansible.

## Why
The `vm-setup/setup-playbook.yml` uses the `community.libvirt` Ansible
module to create and manage KVM virtual machines. This module requires
the `libvirt` Python package to be importable in the same Python
environment that Ansible is running in.

Without this fix, the venv contains ansible but not libvirt-python,
causing the VM creation step to fail with:

"The `libvirt` module is not importable. Check the requirements."

This causes node_0 and node_1 to never be created, leaving Ironic
stuck waiting for bare metal nodes that don't exist.

## Change

Before:
sudo "${ANSIBLE_VENV}/bin/pip" install --ignore-installed ansible=="${ANSIBLE_VERSION}"

After:
sudo "${ANSIBLE_VENV}/bin/pip" install --ignore-installed ansible=="${ANSIBLE_VERSION}" libvirt-python

## Environment tested
- Host OS: Ubuntu 22.04 LTS
- Platform: GCP n2-standard-8 with nested virtualization
- metal3-dev-env: main branch

## Workaround (no longer needed with this fix)
/opt/metal3-dev-env/venv/bin/pip install libvirt-python

### DIFF
--- a/01_prepare_host.sh
+++ b/01_prepare_host.sh
@@ -80,7 +80,7 @@ rm -rf "${ANSIBLE_VENV}"
 sudo python -m venv --system-site-packages "${ANSIBLE_VENV}"
 # TODO: since ansible 8.0.0, pinning by digest is PITA, due additional ansible
 # dependencies, which would need to be pinned as well, so it is skipped for now
-sudo "${ANSIBLE_VENV}/bin/pip" install --ignore-installed ansible=="${ANSIBLE_VERSION}"
+sudo "${ANSIBLE_VENV}/bin/pip" install --ignore-installed ansible=="${ANSIBLE_VERSION}" libvirt-python
 
 # Install requirements
 "${ANSIBLE}-galaxy" install -r vm-setup/requirements.yml


tuminoid (commented):
/ok-to-test

tuminoid:
/hold
There would be no change left after bot suggestion, so let's discuss in issue what the reproduction steps are.

Ahmedaltu:
Hi @tuminoid, happy to discuss the reproduction steps here.

**Environment:**
- GCP n2-standard-8, Ubuntu 22.04 LTS, nested virtualization enabled
- Running `make` as a non-root user with passwordless sudo
- metal3-dev-env main branch

**What fails:**
After `01_prepare_host.sh` completes, `02_configure_host.sh` calls `vm-setup/setup-playbook.yml`. The `community.libvirt` tasks fail with:

"The `libvirt` module is not importable. Check the requirements."

This causes node_0 and node_1 to never be created in libvirt, leaving Ironic stuck waiting for bare metal nodes that don't exist.

**Investigation:**
After looking at the code more carefully, I think the real root cause may be different from my initial fix. The `packages_installation` role does target `${ANSIBLE_VENV}/bin/pip` via:

executable: "{{ ANSIBLE_VENV | default('/usr') }}/bin/pip"

However, the playbook in `01_prepare_host.sh` is run with `-b` (become/sudo), which drops environment variables. This means `ANSIBLE_VENV` is not available inside the playbook, so the role defaults to `/usr/bin/pip` and installs `libvirt-python` into system Python instead of the venv.

**Possible correct fix:**
Pass `ANSIBLE_VENV` explicitly as an extra var to the playbook:

ANSIBLE_FORCE_COLOR=true "${ANSIBLE}-playbook" \
    -e "working_dir=${WORKING_DIR}" \
    -e "metal3_dir=${SCRIPTDIR}" \
    -e "virthost=${HOSTNAME}" \
    -e "ANSIBLE_VENV=${ANSIBLE_VENV}" \
    -i vm-setup/inventory.ini \
    -b vm-setup/install-package-playbook.yml

Would this be the right direction? Happy to update the PR with this fix instead.

tuminoid:
Did you reboot between 01 and 02? That reboot is crucial to get docker working, get you the groups, paths etc?

Ahmedaltu:
Hi @tuminoid, I did not reboot between 01 and 02 — I ran them sequentially as part of the `make` command which calls all four scripts in order without rebooting in between.

Could the missing reboot be the reason `ANSIBLE_VENV` group/env changes don't take effect, causing libvirt-python to install into system Python instead of the venv?

If a reboot is required between `01_prepare_host.sh` and `02_configure_host.sh`, should that be documented in the README or enforced in the Makefile?

Ahmedaltu:
After further testing on both Ubuntu 22.04 and 24.04, I found that libvirt-python is now installed automatically as a dependency of virtualbmc. The original issue may have been caused by an older virtualbmc version that did not declare this dependency.
Leaving this PR open for maintainer review — happy to close it if deemed unnecessary, or keep it as an explicit defensive dependency.

Ahmedaltu:
Closing as the issue appears to be resolved upstream via virtualbmc now declaring libvirt-python as an explicit dependency. Thanks for the guidance @tuminoid!

PR #1683, opened 2026-05-28
Do not output on stdout

terror96:
There is no need to output kubectl completion content on stdout.

Less noise in logs.


tuminoid (commented):
/approve

Sunnatillo (commented):
/lgtm

PR #1690, opened 2026-06-08
fix: remove conflicting Docker apt source files before adding repo

### LINKED ISSUE #1367: Error installing Ubuntu packages in case Docker repository already setup
In my scenario, I had previously configured the Docker apt repository following the guidelines outlined in the official documentation: [Docker Installation Guide for Ubuntu](https://docs.docker.com/engine/install/ubuntu/).

However, a problem arises when comparing this setup with the setup in Metal3-dev-env. The Docker documentation 
uses the plain key format (.asc) for the apt repository signature, whereas the Metal3-dev-env setup utilizes a binary key format (.gpt). This discrepancy results in a clash within the signature, leading to the following error during installation:

```
TASK [packages_installation : Add Docker Repository] ***************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: apt_pkg.Error: E:Conflicting values set for option Signed-By regarding source https://download.docker.com/linux/ubuntu/ focal: /etc/apt/keyrings/docker.asc != /etc/apt/keyrings/docker.gpg,
```

In the code, it can be seen that there is already two tasks addressing this particular issue, both aimed at removing outdated entries with varying apt repository structures. These tasks are flagged with TODO comments to be removed further down the line

A potential solution could be to find all apt repositories associated with the URL https://download.docker.com/linux/ubuntu/, removing all entries and enforcing the utilization of signed-by=/etc/apt/keyrings/docker.gpg.

Ahmedaltu:
When Docker is pre-installed via the official Docker docs, it creates
/etc/apt/sources.list.d/docker.list with Signed-By pointing to docker.asc.
Metal3-dev-env previously used apt_repository with docker.gpg, causing:

  E:Conflicting values set for option Signed-By regarding source
  https://download.docker.com/linux/ubuntu/:
  /etc/apt/keyrings/docker.asc != /etc/apt/keyrings/docker.gpg

Fix by aligning with the official Docker docs:
- Switch to DEB822 format, writing directly to docker.sources
- Use docker.asc directly in Signed-By — no dearmor step needed
- Remove legacy docker.list if present to clean up old-format installs

Fixes #1367

### DIFF
--- a/vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
+++ b/vm-setup/roles/packages_installation/tasks/ubuntu_required_packages.yml
@@ -93,9 +93,10 @@
         mode: '0644'
         force: true
 
-    - name: Dearmor GPG key
-      shell: |
-        cat /etc/apt/keyrings/docker.asc | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg --yes
+    - name: Remove legacy Docker apt source file
+      file:
+        path: /etc/apt/sources.list.d/docker.list
+        state: absent
 
     # TODO: (Sunnatillo) Remove this task after fully removing apt-key
     - name: Remove Docker old repository (without gpg key file location)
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
Wouldn't it make more sense to check if they are there and in that case not add the conflicting ones?
We could also consider changing the way we add them to match the official docs. I guess that would also avoid the issue?

/ok-to-test

Ahmedaltu:
@lentzi90 Thanks for the feedback! You're right on both points.
The current approach is more of a workaround than a proper fix. A cleaner solution would be writing directly to /etc/apt/sources.list.d/docker.sources in DEB822 format using the armored .asc key directly in Signed-By. This way we own the file entirely and there's nothing to conflict with. It would also let us drop the dearmor step.
Would you like me to go ahead with that approach?

lentzi90:
Yes, please go ahead!

lentzi90 (commented):
Thanks!
/lgtm
Please update the PR description also to match the latest.

Ahmedaltu:
Done

kashifest (commented):
/approve

Issue #1691, opened 2026-06-10
Simplify dev-env tests by standardizing on Kind, IRSO, and vbmctl


adilGhaffarDev:
Current state:

- Ubuntu tests: `kind cluster` + `Ironic` running locally (docker containers)
- `CentOS` tests: `minikube cluster` + `Ironic` deployed in-cluster
- VMs created via `Ansible` roles in `vm-setup`
- BMC emulator (sushy-tools/vbmc) managed by [vbmc.sh]
- Network setup done manually via shell scripts

Goal:

- Use kind for both `CentOS` and `Ubuntu`
- Default to `IRSO` for Ironic deployment (management cluster + target cluster after pivot)
- Use `vbmctl` from BMO for `VM/network/BMC` management
- Simplify dev-env by removing duplicated infrastructure code
- Move `deploy.sh` and `run_local_ironic.sh` to metal3-dev-env
  - We will not use them in our CI
  - They will be used for development purposes.

---

It can be divided into following tasks:

- [ ] Standardize on Kind cluster
  - [ ] Remove minikube-specific network setup in `03_launch_mgmt_cluster.sh`
  - [ ] Remove minikube-specific Ironic deployment logic in launch_ironic() function
  - [ ] Update `common.sh` to default EPHEMERAL_CLUSTER to kind regardless of OS
  - [ ] Remove minikube VM setup from `vm-setup` Ansible roles
  - [ ] Update CI configuration in `project-infra` to use `kind` for `CentOS` jobs
  - [ ] Update documentation `README.md`, `vars.md` to reflect Kind-only support

---

- [ ] Default to IRSO for Ironic in management cluster
  - [ ] Set `USE_IRSO=true` as default in common.sh
  - [ ] Remove `run_local_ironic.sh` execution from default Kind path (keep behind flag)
  - [ ] Move `run_local_ironic.sh` and `deploy.sh` script from `bmo` to `metal3-dev-env`
  - [ ] User should still be able to use `deploy.sh` by setting `USE_IRSO=false` 
  - [ ] User should still be able to use `run_local_ironic.sh` by setting `USE_IRSO=false` and `IRONIC_RUN_LOCAL=true`
  - [ ] By default CI will will always deploy in-cluster ironic with IRSO. `run_local_ironic.sh` and `deploy.sh`  are only for development purposes. 

---

- [ ] Use IRSO in target cluster after pivot (tests)
  - [ ] Update move.yml to install IRSO and create Ironic CR
[...]

PR #1692, opened 2026-06-11
Default to Kind and in-cluster Ironic via IRSO

adilGhaffarDev:
This PR simplifies metal3-dev-env by standardizing on Kind as the only
bootstrap cluster and using ironic-standalone-operator (IRSO) for
in-cluster Ironic deployment by default.

This PR is doing following:
- Remove minikube support and standardize on Kind as the only bootstrap
  cluster option (tilt still supported for development)
- Change USE_IRSO default from "false" to "true" to run Ironic inside
  the Kubernetes cluster using ironic-standalone-operator
- Add setup_kind_provisioning_network() to connect Kind node to the
  provisioning bridge via veth pair when using IRSO
- Update pivot/repivot tests to use IRSO (Ironic CR) instead of
  container-based Ironic deployment

This simplifies the deployment architecture by:
- Using a single bootstrap cluster type (Kind) for all environments
- Using in-cluster Ironic for both management and target clusters
- Aligning with the recommended production deployment model

To run Ironic locally: export USE_IRSO="false"

Part of https://github.com/metal3-io/metal3-dev-env/issues/1691


lib/common.sh
```diff
@@ -165,7 +165,7 @@ export CAPM3_BASE_URL="${CAPM3_BASE_URL:-metal3-io/cluster-api-provider-metal3}"
 export CAPM3REPO="${CAPM3REPO:-https://github.com/${CAPM3_BASE_URL}}"
 export CAPM3RELEASEBRANCH="${CAPM3RELEASEBRANCH:-main}"
 
-export USE_IRSO="${USE_IRSO:-false}"
+export USE_IRSO="${USE_IRSO:-true}"
```
dtantsur: Are we going to support deployments that don't use IrSO? Or is it for stable branches?

tests/roles/run_tests/tasks/move.yml
```diff
@@ -21,49 +16,35 @@
...
+    api_version: ironic.metal3.io/v1alpha1
+    kind: Ironic
+    name: ironic
+    namespace: "{{ IRONIC_NAMESPACE }}"
+    kubeconfig: "{{ KUBECONFIG_PATH }}"
+    state: absent
+  when: BOOTSTRAP_CLUSTER == "kind"
 
-- name: Remove Ironic from source cluster (minikube cluster)
+- name: Remove IRSO operator from source cluster (kind cluster)
```
dtantsur: nit: not strictly required

tests/roles/run_tests/tasks/move_back.yml
```diff
@@ -1,12 +1,21 @@
...
+- name: Remove Ironic CR from target cluster
+  kubernetes.core.k8s:
+    api_version: ironic.metal3.io/v1alpha1
+    kind: Ironic
+    name: ironic
+    namespace: "{{ IRONIC_NAMESPACE }}"
+    state: absent
+    kubeconfig: "/tmp/kubeconfig-{{ CLUSTER_NAME }}.yaml"
+
+- name: Remove IRSO operator from target cluster
```
dtantsur: nit: not strictly required

tests/roles/run_tests/tasks/move_back.yml
```diff
@@ -18,20 +27,21 @@
   args:
     chdir: "{{ BMOPATH }}"
 
-- name: Install Ironic in Source cluster (Bootstrap Cluster is kind)
-  shell: "{{ BMOPATH }}/tools/run_local_ironic.sh"
-  environment:
-    CONTAINER_RUNTIME: "{{ CONTAINER_RUNTIME }}"
+- name: Install IRSO operator in Source cluster
+  shell: "make -C {{ IRSOPATH }} install deploy IMG={{ IRSO_IMAGE }}"
```
dtantsur: Are we planning to ever use IrSO from source (e.g. for the CI)? If not, you may be better off using released manifests.

lib/common.sh
adilGhaffarDev: for CI no but user can still user run-local and deploy scripts for development purposes. 
We will move those scripts from bmo to dev-env.

tests/roles/run_tests/tasks/move.yml
adilGhaffarDev: Ok, i will test it and see how it goes.

tests/roles/run_tests/tasks/move_back.yml
adilGhaffarDev: In CI no, in CI we will always use release branches. I will fix this.

PR #1693, opened 2026-06-12
Increase Minikube defaults to 4 CPUs and 6Gi memory

kashifest:
Recently few runs showed symptomps that etcd got choked with resource constraints, bumping the resources for minikube a bit further.


kashifest:
@Sunnatillo can you please trigger the other tests you mentioned offline that we should run here ?

Sunnatillo (commented):
We do not this PR anymore

PR #1694, opened 2026-06-12
Fix Go tarball and CPU model for aarch64 hosts

fonta-rh:
## Summary

Two fixes for running metal3-dev-env on native aarch64 hosts (e.g. AWS
Graviton bare metal):

1. **Go tarball download** — `go_tarball` in the `packages_installation`
   role defaults hardcodes `linux-amd64`, fetching the wrong binary on
   aarch64. Adds a `go_arch` variable derived from `ansible_architecture`.

2. **CPU model template** — `baremetalvm.xml.j2` unconditionally uses
   `cortex-a57` for all aarch64 VMs. That model only works under QEMU
   emulation (cross-arch from x86_64). Narrows the conditional to
   `{% if is_aarch64 and libvirt_domain_type == 'qemu' %}` so native
   aarch64 KVM falls through to `host-passthrough`.

## Changes

| File | Fix |
|------|-----|
| `vm-setup/roles/packages_installation/defaults/main.yml` | Add `go_arch` mapping, use in `go_tarball` |
| `vm-setup/roles/libvirt/templates/baremetalvm.xml.j2` | Narrow CPU model conditional to QEMU emulation only |

## Context

Discovered while enabling `openshift-metal3/dev-scripts` on AWS Graviton
bare metal (native aarch64 KVM). Related:
[openshift-metal3/dev-scripts#1908](https://github.com/openshift-metal3/dev-scripts/pull/1908).

## Test plan

- [x] Full OCP 4.22 fencing-IPI deployment on AWS c7g.metal (Graviton3,
  native aarch64 KVM) — 2m+1a arbiter cluster, 50m41s, 35/35 COs Available
- [x] Go 1.24.10 arm64 installed via patched defaults, checksum verified
- [x] No change on x86_64 hosts — `go_arch` evaluates to `amd64`,
  `libvirt_domain_type` is `qemu` for cross-arch (cortex-a57 still used)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1695, opened 2026-06-12
Fix CPU model for native aarch64 KVM hosts

fonta-rh:
## Summary

The `baremetalvm.xml.j2` template unconditionally uses a custom CPU model
(cortex-a57) for all aarch64 VMs. This only works under QEMU emulation
(cross-arch from x86_64 host). Native aarch64 KVM hosts — such as AWS
Graviton bare metal — require `host-passthrough` to expose the real CPU
features to guests.

## Changes

- `vm-setup/roles/libvirt/templates/baremetalvm.xml.j2`: Narrow the CPU
  model conditional from `{% if is_aarch64 %}` to
  `{% if is_aarch64 and libvirt_domain_type == 'qemu' %}` so native KVM
  falls through to `host-passthrough`

The `<os>`, `<features>`, and VNC sections remain gated on `is_aarch64`
alone (unchanged) — those are architecture-specific regardless of
emulation vs native KVM.

## Context

Discovered while enabling `openshift-metal3/dev-scripts` on AWS Graviton
bare metal (native aarch64 KVM). With cortex-a57, VMs fail to boot under
KVM. Related: [openshift-metal3/dev-scripts#1908](https://github.com/openshift-metal3/dev-scripts/pull/1908).

## Test plan

- [x] Full OCP 4.22 fencing-IPI deployment on AWS c7g.metal (Graviton3,
  native aarch64 KVM) — 2m+1a arbiter cluster, 50m41s, 35/35 COs Available
- [x] No change for cross-arch use case — `libvirt_domain_type` is `qemu`
  when architectures differ, so cortex-a57 is still used for x86→aarch64

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/vm-setup/roles/libvirt/templates/baremetalvm.xml.j2
+++ b/vm-setup/roles/libvirt/templates/baremetalvm.xml.j2
@@ -45,14 +45,15 @@
     <pae/>
 {% endif %}
   </features>
-{% if is_aarch64 %}
+{% if is_aarch64 and libvirt_domain_type == 'qemu' %}
   <!--
-       On aarch64 we use a fixed/custom CPU model (typically "cortex-a57",
-       configured via libvirt_aarch64_cpu_model) instead of host-model or
-       host-passthrough. This avoids inconsistencies in how libvirt exposes
-       host CPUs on ARM platforms and provides a stable, migration-friendly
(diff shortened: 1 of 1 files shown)


fonta-rh:
Consolidated into #1694.

PR #1696, opened 2026-06-16
Fix leap keepalived interface in template

nuhakala:
The leap node image provisioning fails because the keepalived interface is wrong.


lentzi90 (commented):
Good find!
/approve

smoshiur1237 (commented):
/lgtm

PR #1701, opened 2026-06-25
vbmc: use local libvirt socket instead of SSH

tuminoid:
Replace the qemu+ssh:// connection URI with a direct qemu:///system socket connection for VirtualBMC. Mount the libvirt socket into the vbmc container to enable this.

By removing SSH from the critical path, vbmc talks to libvirtd through a Unix socket, eliminating the network/auth layer that caused the race.

### DIFF
--- a/02_configure_host.sh
+++ b/02_configure_host.sh
@@ -532,14 +532,20 @@ else
 fi
 
 # Start vbmc and sushy containers
+# Mount the libvirt socket directly to avoid SSH-based qemu+ssh:// connections
+# which are prone to transient failures causing power-on commands to be acked
+# by vbmc but not executed by libvirt (observed as VMs stuck powered off after
+# cleaning in CI, e.g. pivoting test flakiness).
 # shellcheck disable=SC2086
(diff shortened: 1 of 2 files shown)


tuminoid:
/hold
I'm testing this to get rid of flakyness in periodics.

tuminoid:
/unhold
/cc @Rozzii @nuhakala @lentzi90 

Any reason we aren't doing this, which is maybe less flaky than ssh?

lentzi90 (commented):
Thanks! This has been bugging me for a long time, since it requires the dev-env host to accept ssh for root.
And no, I don't see any reason why we would need to do it over ssh. That would be needed only if vbmc and libvirt were on separate hosts. A valid scenario but not one we support AFAIK.
/approve

kashifest (commented):
/lgtm

dtantsur:
Huh, interesting! @lentzi90 do you think this was also the cause of soft power off failures in the servicing tests?

lentzi90:
> Huh, interesting! @lentzi90 do you think this was also the cause of soft power off failures in the servicing tests?

Hmm not sure about that. What servicing tests do you mean? In BMO E2E? I think there we always just mounted the socket so it would not be applicable.

PR #1717, opened 2026-08-20
Add check for the creation of VMs

### LINKED ISSUE #1420: Creation of VMs is never checked or waited for
Currently in Centos e2e, the VMs are not created so BMH cannot be inspected nor serial logs fetched etc. It seems we never check if those VMs are created, running, or healthy; we just assume they are. This makes debugging issues harder as we go all the way until end of `04_verify.sh` to find out BMH inspection fails, which could of course be happening for any reason.

At the start of 04, verify the VMs are running and healthy.

/kind bug

UncivilizedGhost:
Check VMs are defined in the node list.

Verify BMH VMs exist and, when provisioned, are powered on. The power
check only runs for provisioned hosts (since Ironic leaves ready hosts
powered off) via .status.poweredOn.

During verify script
To fix https://github.com/metal3-io/metal3-dev-env/issues/1420

### DIFF
--- a/04_verify.sh
+++ b/04_verify.sh
@@ -101,9 +101,17 @@ check_bm_hosts() {
     # Skip introspection verification in no BMH applied
     if [[ "${SKIP_APPLY_BMH:-false}" != "true" ]]; then
       #Verify the introspection completed successfully
+      PROV_STATE="$(echo "${BARE_METAL_HOST}" | jq -r '.status.provisioning.state')"
       RESULT_STR="${NAME} Baremetalhost introspecting completed"
-      is_in "$(echo "${BARE_METAL_HOST}" | jq -r '.status.provisioning.state')" \
-        "ready available"
+      is_in "${PROV_STATE}" "ready available provisioned"
+
+      # Only assert the VM is powered on when the host is actually provisioned.
+      # In ready/available state Ironic leaves the node powered off, so a
+      # "running" assertion would never converge and would time out.
+      if [[ "${PROV_STATE}" == "provisioned" ]]; then
+        RESULT_STR="${NAME} Baremetalhost VM is running"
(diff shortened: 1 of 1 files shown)


04_verify.sh
```diff
@@ -101,9 +101,17 @@ check_bm_hosts() {
     # Skip introspection verification in no BMH applied
     if [[ "${SKIP_APPLY_BMH:-false}" != "true" ]]; then
       #Verify the introspection completed successfully
+      PROV_STATE="$(echo "${BARE_METAL_HOST}" | jq -r '.status.provisioning.state')"
       RESULT_STR="${NAME} Baremetalhost introspecting completed"
-      is_in "$(echo "${BARE_METAL_HOST}" | jq -r '.status.provisioning.state')" \
-        "ready available"
+      is_in "${PROV_STATE}" "ready available"
```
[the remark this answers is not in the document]
UncivilizedGhost: doone

04_verify.sh
```diff
@@ -223,6 +231,29 @@ BMO_RUN_LOCAL="${BMO_RUN_LOCAL:-false}"
 CAPM3_RUN_LOCAL="${CAPM3_RUN_LOCAL:-false}"
 
 
+# Verify VMs are defined (skip on fake platform)
+if [[ "${NODES_PLATFORM}" != "fake" ]]; then
+  VM_OK=true
```
kashifest: Its bit strange for me to articulate this. We start with assuming that the vms are defined even before the verification starts, what if the next if block is never entered, then it will assume VM is defined. 

Irrespective of if the logic here is correct or not, I would rather start assuming VM is not defined and only set it to true when I find it is.

UncivilizedGhost: I flipped the logic so it assumes no VMs are defined at the start and starts checking the missing VMs.
It only continues if they are no missing VMs.

PR #1728, opened 2026-09-09
Remove release-1.11 templates and add 1.14 templates

nuhakala:
Renames old templates to 1.14 templates. Or actually deleted the old templates and copied the templates from main folder, but apparently there is no changes and git detects this as rename.


lentzi90 (commented):
Sigh. These templates are ridiculous. We template the API version but still have a hard coded API we follow. So in theory you can set `CAPM3_VERSION=v1beta1`, but in practice that will error at the webhook when you try to apply a Metal3MachineTemplate with `spec.template.spec.image.diskFormat` instead of `format`. 🤦 

So to whoever reads this. Please avoid doing more pointless templates like these.
If a value cannot in practice be changed, there is no point having a variable for it to begin with.

/approve

Sunnatillo (commented):
/lgtm

PR #1731, opened 2026-09-18
detach from nordix artifactory

Rozzii:
This PR:
- Pins ubuntu, centos, and leap node images in lib/images.sh to fixed
  OCI Object Storage URLs instead of Nordix Artifactory
- Removes the KUBERNETES_VERSION templating from IMAGE_NAME/IMAGE_LOCATION
  defaults for these OS images; this templating was legacy logic that only
  ever worked reliably against the Metal3 upstream CI node image
  naming/layout, not a generally portable pattern
- Updates config_example.sh, vars.md, and AGENTS.md examples/docs to match,
  and adds a caveat note to vars.md documenting this behavior change
- Aligns tests/roles/run_tests/vars/main.yml and
  tests/roles/run_tests/tasks/upgrade_k8s.yml to use the same fixed image

Caveat:
  upgrade_k8s.yml no longer downloads an image matching
  UPGRADED_K8S_VERSION; it reuses the same fixed image, so only the
  kubeadm-level version upgrade is exercised, not a matching image swap

Reason:
 Nordix Artifactory is experiencing technical issues, so image
 downloads are moved off it to avoid CI/dev-env failures

A follow-up PR will revisit this once Metal3 CI infrastructure has
fully recovered


Rozzii:
/override metal3-centos-e2e-integration-test-release-1-13
this is still trying to pull ipa from artifactory, so IMO this test is not representative of this dev-env change because with this PR dev-env does not contain any reference to Nordix Artifactory

kashifest (commented):
/approve

lentzi90 (commented):
/lgtm
I wonder do we even run the k8s upgrade in dev-env in any jobs? I think we do al upgrade tests in CAPM3 nowadays

